PR #27, opened 2025-05-01
A rough sketch of a fast caching algorithm using shared buffers that …

kriszyp:
…might work


kriszyp:
I am not really working on this, so will close this, at least for now.

PR #149, opened 2025-09-18
[CORE-2698] Transaction log store

cb1kenobi:
Adding support for the transaction log store.

Fixes https://github.com/HarperFast/rocksdb-js/issues/217.

### DIFF
--- a/docs/transaction-log.md
+++ b/docs/transaction-log.md
@@ -45,11 +45,11 @@ Log files follow the pattern: `{name}.{sequenceNumber}.txnlog`
 ### File Header (13 bytes)
 
 ```
-| Offset | Size | Type    | Field          | Description            |
-|--------|------|---------|----------------|------------------------|
-| 0      | 4    | uint32  | token          | Transaction log token  |
-| 4      | 1    | uint8   | version        | Format version         |
-| 5      | 8    | double  | fileTimestamp  | The latest timestamp   |
+| Offset | Size | Type    | Field      | Description            |
+|--------|------|---------|------------|------------------------|
+| 0      | 4    | uint32  | token      | Transaction log token  |
+| 4      | 1    | uint8   | version    | Format version         |
+| 5      | 8    | double  | timestamp  | The latest timestamp   |
 ```
 
 #### `token`
@@ -60,19 +60,19 @@ The token is used to validate that the file is indeed a transaction log.
 
 The transaction log file format version. Currently, version `1` is the latest.
 
-#### `fileTimestamp`
+#### `timestamp`
 
-?????
+The timestamp of the most recent transaction log batch that has been written.
 
 ### Transaction Header (13 bytes)
 
-| Offset | Size | Type    | Field         | Description                    |
-|--------|------|---------|---------------|--------------------------------|
-| 0      | 8    | double  | txnTimestamp  | Timestamp transaction created  |
-| 8      | 4    | uint32  | entrySize     | Size of the entry data         |
-| 12     | 1    | uint8   | flags         | Transaction flags              |
+| Offset | Size | Type    | Field      | Description                    |
+|--------|------|---------|------------|--------------------------------|
+| 0      | 8    | double  | timestamp  | Timestamp transaction created  |
+| 8      | 4    | uint32  | entrySize  | Size of the entry data         |
+| 12     | 1    | uint8   | flags      | Transaction flags              |
 
-#### `txnTimestamp`
+#### `timestamp`
 
 The timestamp the associated transaction was created.
 
--- a/src/binding/transaction_log_store.cpp
+++ b/src/binding/transaction_log_store.cpp
@@ -45,9 +45,9 @@ void TransactionLogStore::close() {
 	this->purge();
 }
 
-TransactionLogFile* TransactionLogStore::getLogFile(const uint32_t sequenceNumber) {
+std::shared_ptr<TransactionLogFile> TransactionLogStore::getLogFile(const uint32_t sequenceNumber) {
 	auto it = this->sequenceFiles.find(sequenceNumber);
-	auto logFile = it != this->sequenceFiles.end() ? it->second.get() : nullptr;
+	std::shared_ptr<TransactionLogFile> logFile = it != this->sequenceFiles.end() ? it->second : nullptr;
 
 	if (!logFile) {
 		DEBUG_LOG("%p TransactionLogStore::getLogFile Store path \"%s\" (seq=%u) no log file found, creating\n",
@@ -58,8 +58,8 @@ TransactionLogFile* TransactionLogStore::getLogFile(const uint32_t sequenceNumbe
 
 		std::string filename = std::to_string(sequenceNumber) + ".txnlog";
 		auto logFilePath = this->path / filename;
-		logFile = new TransactionLogFile(logFilePath, sequenceNumber);
-		this->sequenceFiles[sequenceNumber] = std::unique_ptr<TransactionLogFile>(logFile);
+		logFile = std::make_shared<TransactionLogFile>(logFilePath, sequenceNumber);
+		this->sequenceFiles[sequenceNumber] = logFile;
 	}
 
 	return logFile;
@@ -143,8 +143,8 @@ void TransactionLogStore::purge(std::function<void(const std::filesystem::path&)
 void TransactionLogStore::registerLogFile(const std::filesystem::path& path, const uint32_t sequenceNumber) {
 	std::lock_guard<std::mutex> lock(this->storeMutex);
 
-	auto logFile = std::make_unique<TransactionLogFile>(path, sequenceNumber);
-	this->sequenceFiles[sequenceNumber] = std::move(logFile);
+	auto logFile = std::make_shared<TransactionLogFile>(path, sequenceNumber);
+	this->sequenceFiles[sequenceNumber] = logFile;
 
 	if (sequenceNumber > this->currentSequenceNumber) {
 		this->currentSequenceNumber = sequenceNumber;
@@ -165,9 +165,14 @@ void TransactionLogStore::writeBatch(TransactionLogEntryBatch& batch) {
 
 	std::lock_guard<std::mutex> lock(this->storeMutex);
 
+	if (batch.timestamp > this->latestTimestamp) {
+		DEBUG_LOG("%p TransactionLogStore::commit Setting latest timestamp to batch timestamp: %f > %f\n", this, batch.timestamp, this->latestTimestamp)
+		this->latestTimestamp = batch.timestamp;
+	}
+
 	// write entries across multiple log files until all are written
 	while (!batch.isComplete()) {
-		TransactionLogFile* logFile = nullptr;
+		std::shared_ptr<TransactionLogFile> logFile = nullptr;
 
 		// get the current log file and rotate if needed
 		while (logFile == nullptr && this->currentSequenceNumber) {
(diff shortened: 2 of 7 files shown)


src/binding/transaction_log.cpp
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+	if (transactionId == 0) {
+		::napi_throw_error(env, nullptr, "Invalid argument, expected a transaction id");
+		return nullptr;
+	}
+
+	// // create a reference to pin the buffer in memory (prevents GC)
+	napi_ref bufferRef = nullptr;
+	NAPI_STATUS_THROWS(::napi_create_reference(env, argv[0], 1, &bufferRef));
```
kriszyp: I think we should copy the contents of the buffer into `data`, keep this as managed memory in C (delete when done) rather than maintaining a JS ref:
* Remove the complexity of managing, tracking, and deleting the JS reference
* Allow the JS buffer to be immediately GC'ed. For very large txns, helps protect against JS OOM, better scalability.
* Or allow for the safe reuse of shared buffer, that can be written over and over. Right now, I think this expects that the buffer won't be modified until the commit takes place, but this shouldn't be assumed, copying the data keeps it safe.

src/binding/transaction_log_file_posix.cpp
```diff
@@ -0,0 +1,97 @@
...
+		this->fd = -1;
+	}
+}
+
+void TransactionLogFile::flushFile() {
+	if (this->fd < 0) {
+		return;
+	}
+
+	if (::fsync(this->fd) < 0) {
```
kriszyp: I think we probably want to use `fdatasync`, a little faster and we don't need to update file metadata with every flush.

cb1kenobi: Ah, yes `fdatasync()`. I'm fine with this, but we need to be comfortable with the tradeoffs. Since `fdatasync()` does not update the log file's mtime, the mtime will be stale until the kernel writeback thread runs and updates the file's mtime. Apparently this can take 5-30 seconds. Should the actualy OS panic or get powered off during this time, then we would have a stale mtime and the rocksdb-js code could potentially delete a transaction log outside the user-defined retention window. We default to 3 days, so odds of this being a problem are really slim. Just something to be aware of.

cb1kenobi: And apparently macOS also doesn't support `fdatasync()`.

docs/transaction-log-v1.md
```diff
@@ -0,0 +1,313 @@
...
+- `sequenceNumber`: Sequential integer starting from 1
+- Example: `mylog.1.txnlog`, `mylog.2.txnlog`
+
+## Binary Format Specification
+
+### File Header (10 bytes)
+
+```
++------------------+
+| WOOF Token       | 4 bytes
```
kriszyp: Confirmed! https://github.com/HarperFast/rocksdb-js/pull/238/files#diff-22a3c4e9b693561b0d3d6534ae81880ea0dec05f14fc12b7bbc837a5748e59e5R118

src/binding/transaction_log_store.h
```diff
@@ -0,0 +1,143 @@
...
+	/**
+	 * Closes the transaction log store and all associated log files.
+	 * This method waits for all active operations to complete before closing.
+	 */
+	void close();
+
+	/**
+	 * Writes a batch of transaction log entries to the store.
+	 */
+	void commit(TransactionLogEntryBatch& batch);
```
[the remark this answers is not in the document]
cb1kenobi: I love it! I'll make it so.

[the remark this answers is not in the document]
kriszyp: Excellent, sounds good to me. A transaction can be "written" and then "commited" :+1: .

docs/transaction-log-v1.md
```diff
@@ -0,0 +1,313 @@
...
+Total: 4096 bytes
+```
+
+### Block Header (14 bytes)
+
+All multi-byte integers are encoded in **big-endian** format.
+
+| Offset | Size | Type    | Field           | Description                                  |
+|--------|------|---------|-----------------|----------------------------------------------|
+| 0      | 8    | uint64  | startTimestamp  | Block timestamp (milliseconds since epoch)   |
```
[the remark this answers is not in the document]
cb1kenobi: Done!

src/binding/util.cpp
```diff
@@ -400,4 +415,67 @@ const char* getNapiBufferFromArg(
...
+	uint64_t last = lastTimestamp.load(std::memory_order_acquire);
+	uint64_t now = std::chrono::duration_cast<std::chrono::nanoseconds>(
+		std::chrono::system_clock::now().time_since_epoch()
+	).count();
+
+	do {
+		if (now <= last) {
+			now = last + 1;
+		}
+	} while (!lastTimestamp.compare_exchange_strong(last, now, std::memory_order_acq_rel));
```
kriszyp: TIL that `compare_exchange_strong` updates the `last` (expected) value if it doesn't match, completing the update for the loop. Brilliant!

src/binding/util.cpp
```diff
@@ -400,4 +415,67 @@ const char* getNapiBufferFromArg(
...
+
+	return static_cast<double>(now) / 1000000000.0;
```
kriszyp: I think you have 3 extra zeros in here, nanoseconds to milliseconds should be dividing by 1000000 (million), right?

src/binding/transaction_log_store.cpp
```diff
@@ -0,0 +1,309 @@
...
+		this, batch.entries.size(), this->name.c_str(), batch.timestamp)
+
+	std::lock_guard<std::mutex> lock(this->storeMutex);
+
+	// write entries across multiple log files until all are written
+	while (!batch.isComplete()) {
+		TransactionLogFile* logFile = nullptr;
+
+		// get the current log file and rotate if needed
+		while (logFile == nullptr && this->currentSequenceNumber) {
```
kriszyp: As far as I can tell, this is not implementing https://github.com/HarperFast/rocksdb-js/pull/238/commits/6dd20de7af5f83c82ec465c5d4e375ded6691cac#diff-2872eef24b542a107a9c6f0b14427d656cf6cd3bef907f9a991b1482d71855acR18:

> A new transaction log file may be started even if the previous file has not reached its maximum size. A new transaction log file should be started if the timestamp is older than 1/8 of the transaction retention period.

(1/4 might be better though)

src/binding/util.cpp
```diff
@@ -400,4 +415,73 @@ const char* getNapiBufferFromArg(
...
+double getTimestamp() {
+	uint64_t now = std::chrono::duration_cast<std::chrono::nanoseconds>(
+		std::chrono::system_clock::now().time_since_epoch()
+	).count();
+
+	double result = static_cast<double>(now / 1000000) + static_cast<double>(now % 1000000) / 1000000.0;
+
+	double last = lastTimestamp.load(std::memory_order_acquire);
+	if (result <= last) {
+		result = std::nextafter(last, std::numeric_limits<double>::infinity());
```
kriszyp: Seems to fail to compile. Maybe just
`result = (double) ((uint64_t)result + 1);` ?

src/binding/util.cpp
cb1kenobi: Fixed!

src/binding/transaction_log_file_posix.cpp
```diff
@@ -0,0 +1,114 @@
...
+			std::filesystem::create_directories(parentPath);
+		} catch (const std::filesystem::filesystem_error& e) {
+			DEBUG_LOG("%p TransactionLogFile::openFile Failed to create parent directory: %s (error=%s)\n",
+				this, parentPath.string().c_str(), e.what())
+			throw std::runtime_error("Failed to create parent directory: " + parentPath.string());
+		}
+	}
+
+	// open file for both reading and writing
+	this->fd = ::open(this->path.c_str(), O_RDWR | O_CREAT, 0644);
```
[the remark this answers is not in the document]
cb1kenobi: Fixed!

src/binding/util.h
```diff
@@ -373,6 +488,82 @@ struct AsyncWorkHandle {
...
+	uint64_t bits;
+	std::memcpy(&bits, &value, sizeof(double));
+	writeUint64BE(buffer, bits);
+}
+
+inline double readDoubleBE(const char* buffer) {
+	// Read uint64_t in big endian and interpret as double
+	uint64_t bits = readUint64BE(buffer);
+	double value;
+	std::memcpy(&value, &bits, sizeof(double));
```
kriszyp: Can we can cast without a memcpy? `*((double*) &bits)` works, not sure if there is prettier way.

src/binding/util.h
```diff
@@ -373,6 +488,82 @@ struct AsyncWorkHandle {
...
+	buffer[0] = static_cast<char>((value >> 8) & 0xFF);
+	buffer[1] = static_cast<char>(value & 0xFF);
+}
+
+inline void writeUint8(char* buffer, uint8_t value) {
+	buffer[0] = static_cast<char>(value);
+}
+
+inline uint64_t readUint64BE(const char* buffer) {
+	return (static_cast<uint64_t>(static_cast<uint8_t>(buffer[0])) << 56) |
```
kriszyp: At some point, maybe we could use `htonl` or `bswap_64` for a single machine operation conversion.

cb1kenobi:
@kriszyp I think this is looking good and stable. I believe I have addressed all feedback. I have code reviewed it an tests are passing, so I'm eager to fix any issues you find. Thanks!

src/binding/binding.cpp
```diff
@@ -30,23 +32,23 @@ NAPI_MODULE_INIT() {
 	napi_create_string_utf8(env, rocksdb::GetRocksVersionAsString().c_str(), NAPI_AUTO_LENGTH, &version);
 	napi_set_named_property(env, exports, "version", version);
 
-	[[maybe_unused]] int refCount = ++moduleRefCount;
```
kriszyp: I am curious about this change, as I have been using the same shorthand incrementation/decrementation pattern you are deleting. I thought https://en.cppreference.com/w/cpp/atomic/atomic/operator_arith.html explicitly guarantees `++moduleRefCount` and `moduleRefCount.fetch_add(1) + 1` are identical expressions (and same is true that `--moduleRefCount` and `moduleRefCount.fetch_sub(1) - 1` are guaranteed to be identical) because  `std::atomic` does operator overloading of `++` and `--` to ensure they are atomic. Do you know how they differ or sure they differ?

src/binding/transaction_log.cpp
```diff
@@ -0,0 +1,198 @@
...
+			::napi_throw_type_error(env, nullptr, "Invalid argument, transaction id must be a non-negative integer");
+			return nullptr;
+		}
+	} else if (transactionId == 0) {
+		::napi_throw_type_error(env, nullptr, "Missing argument, transaction id is required");
+		return nullptr;
+	}
+
+	try {
+		std::unique_ptr<char[]> copyData(new char[size]);
```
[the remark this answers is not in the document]
cb1kenobi: Done!

docs/transaction-log.md
```diff
@@ -0,0 +1,256 @@
...
++---------------------+
+```
+
+### File Header (5 bytes)
+
+```
+| Offset | Size | Type    | Field      | Description            |
+|--------|------|---------|------------|------------------------|
+| 0      | 4    | uint32  | token      | Transaction log token  |
+| 4      | 1    | uint8   | version    | Format version         |
```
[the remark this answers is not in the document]
cb1kenobi: Done!

PR #186, opened 2025-10-20
Fix getKeys() iterable to return the keys and not objects of keys

cb1kenobi:
`getKeys()` iterable is supposed to produce strings, not objects. Whoops.


kriszyp (approved):
Looks good!

Issue #219, opened 2025-11-06
Transaction Log: Bind commit to RocksDB flush


cb1kenobi:
The current Transaction Log design writes and flushes entries to disk prior to calling RocksDB's transaction commit function. Instead, we may want to commit the transaction log during the RocksDB Transaction `OnFlushBegin` callback. This may allow us to avoid committing the transaction log entries until we know the transaction is conflict-free and committable.

Additionally, we may want to add a `OnFlushCompleted` listener so that the transaction log system can write a marker to a file in the database directory indicating the SST has been successfully flushed to disk and prior transactions do not need to be replayed. It is possible that RocksDB already writes this information in one of the manifest files (CURRENT -> MANIFEST?).

RocksDB EventListener docs: https://github.com/facebook/rocksdb/wiki/EventListener

PR #238, opened 2025-11-13
Transaction log querying

### LINKED ISSUE #218: Query Transaction Log
We need an API for querying a transaction log and return all entries for a given start and end timestamp range. Entries contain a timestamp and a buffer with the entry data.

```typescript
const log = db.useLog('foo');

const iter = log.query({
  start: Date.now() - (60 * 60 * 1000), // last hour
  end: Date.now()
});

for (const entry of iter) {
  console.log(entry);
}
```

Under the hood, the Transaction Log Store maintains a directory of active Transaction Logs and metadata including the earliest transaction timestamp and a handle to a memory map for each sequence log file. Once the query logic identifies the first transaction log sequence file, it will access the file via the memory map and use a binary search algorithm to quickly locate the first entry that is greater than or equal to the specified start timestamp. From there, it simply maintains a cursor that emits each matching entry until the end timestamp or end of file.

kriszyp:
Implements transaction log querying and iterating, fixes #218.

This is implemented using memory maps to read the data from transaction logs. This allows the iterating through a transaction log to be done entirely in JavaScript using data views and subarrays, for most iterations, with typical iterations taking around 50ns for returning the next entry (with a timestamp and buffer). The primary public API for querying and reading the transaction log is the `log.query(options)` method, which provides options for defining a `start` time, `end` time, and some additional options for the range of entries to match (including `exclusiveStart`). The `query` method returns an `Iterator<TransactionEntry>` (`& Iterable`). Notably, the returned `Iterator`/`Iterable` can be used multiple times to resume iteration, allowing one to resume and continue iterating through a transaction log after iterator has previously finished, but additional commits have been made, yielding more possible iterations. (This will be used heavily by replication and real-time subscriptions).

Querying to find a starting position is implemented in C++, by creating an on-demand index of each log file that will be searched. The index is constructed as an (ordered) `std::map` that can be queried in `O(log n)` time. The result of a query for a start time finds the log position for which there are no entries with an earlier timestamp before the returned position. Note that the transaction log is _not_ guaranteed to be ordered. It is roughly ordered, we write transactions chronologically, but it is possible for one transaction to take longer than another, in which case it will receive an earlier timestamp (at the start of the transaction), but will be written after a shorter transaction with a later starting timestamp. With the returned position, the iterator is responsible for filtering out any transactions that don't match the conditions (possibly due to out-of-order writes).

The concept of a transaction log "position" is used heavily throughout the querying and is represented by a 64-bit number where the high 32-bits are the log sequence number, and the lower 32-bits are the offset within a log file. Together, this 64-bit position provides a single absolute definition of a position withing a named log sequence. This is returned by the `findPosition` method to determine a starting point (returned as 64-bit float since that is the most efficient 64-bit primitive).

The "end" of the transaction log is also represented with a (64-bit) position. By default, the end of the transaction log is defined as the last position in the transaction log for which all previous transactions have been committed (to the RocksDB database). There may be additional transaction entries in transaction log that have been written but not yet committed. Typically we don't want to access these entries because doing so makes visible race conditions and inconsistencies between log entries and database values within transactions that may or may not be fully committed (breaking ACID compliance). This end position or "lastPositionCommitted" is determined by tracking the written and uncommitted transactions, removing them when they are committed and then finding the new last position after each commit completes. This position is made available to JavaScript through a shared external buffer.

A query can also be executed with a `readUncommitted` that can be used to access to transactions that have not been committed yet. This is particularly useful for replaying transaction logs on restart where many transactions that have been "committed" to RocksDB may have effectively become "uncommitted" after a process crashes/terminates because the data was not written before termination.

This PR also includes the basic mechanism for associating the transaction log position with database sequence numbers so that when a database SST file is "flushed", we can track how much of the transaction log has data that has been fully flushed to disk. This is critical for determining how much of the transaction log needs to be replayed on startup (any data that might not be flushed to disk). This mechanism's primary API is a `TransactionLogStore::databaseFlushed` method that can be called with the sequence number of a database flush so that https://github.com/HarperFast/rocksdb-js/issues/219 can be completed.

To no one's surprise, Windows throws in an extra element of fun. In Linux (and macos), we can create an arbitrarily large memory map, and as a file grows, the additional data in the memory map will be accessible (accessing data beyond the file size is a bus error). However, Windows requires that a memory map be fully backed by a full of the same (or larger) size. Well, you can actually get around this by directly hacking into the NT.DLL using driver-level APIs, but I assumed we probably didn't want to carry these techniques into this library:
https://github.com/kriszyp/lmdb-js/blob/master/dependencies/lmdb/libraries/liblmdb/mdb.c#L5998-L6012
So instead the file size is adjusted to match the memory map when a memory map is needed. This probably is good for performance, ensuring more linear disk usage.

### DIFF
--- a/docs/transaction-log.md
+++ b/docs/transaction-log.md
@@ -45,11 +45,11 @@ Log files follow the pattern: `{name}.{sequenceNumber}.txnlog`
 ### File Header (13 bytes)
 
 ```
-| Offset | Size | Type    | Field          | Description            |
-|--------|------|---------|----------------|------------------------|
-| 0      | 4    | uint32  | token          | Transaction log token  |
-| 4      | 1    | uint8   | version        | Format version         |
-| 5      | 8    | double  | fileTimestamp  | The latest timestamp   |
+| Offset | Size | Type    | Field      | Description            |
+|--------|------|---------|------------|------------------------|
+| 0      | 4    | uint32  | token      | Transaction log token  |
+| 4      | 1    | uint8   | version    | Format version         |
+| 5      | 8    | double  | timestamp  | The latest timestamp   |
 ```
 
 #### `token`
@@ -60,19 +60,19 @@ The token is used to validate that the file is indeed a transaction log.
 
 The transaction log file format version. Currently, version `1` is the latest.
 
-#### `fileTimestamp`
+#### `timestamp`
 
-?????
+The timestamp of the most recent transaction log batch that has been written.
 
 ### Transaction Header (13 bytes)
 
-| Offset | Size | Type    | Field         | Description                    |
-|--------|------|---------|---------------|--------------------------------|
-| 0      | 8    | double  | txnTimestamp  | Timestamp transaction created  |
-| 8      | 4    | uint32  | entrySize     | Size of the entry data         |
-| 12     | 1    | uint8   | flags         | Transaction flags              |
+| Offset | Size | Type    | Field      | Description                    |
+|--------|------|---------|------------|--------------------------------|
+| 0      | 8    | double  | timestamp  | Timestamp transaction created  |
+| 8      | 4    | uint32  | entrySize  | Size of the entry data         |
+| 12     | 1    | uint8   | flags      | Transaction flags              |
 
-#### `txnTimestamp`
+#### `timestamp`
 
 The timestamp the associated transaction was created.
 
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "@vitest/coverage-v8": "3.2.4",
     "cross-env": "10.1.0",
     "dotenv": "17.2.3",
-    "lefthook": "2.0.4",
+    "lefthook": "2.0.5",
     "lmdb": "3.4.4",
     "node-gyp": "12.1.0",
     "oxlint": "1.31.0",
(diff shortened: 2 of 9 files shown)


docs/transaction-log.md
```diff
@@ -0,0 +1,18 @@
+The transaction log is a file that contains a sequence of entries that describe the updates that have been made to the database. These updates are intended to be largely opaque to RocksDB, and to be can be used by the rocksdb-js consumer to implement their own replay and subscription semantics.
```
Ethan-Arrowood:
```suggestion
The transaction log is a file that contains a sequence of entries that describe the updates that have been made to the database. These updates are intended to be largely opaque to RocksDB, and can be used by the rocksdb-js consumer to implement their own replay and subscription semantics.
```

src/binding/transaction_log.cpp
Ethan-Arrowood: I noticed a handful of TODO comments in here; is this code ready for review or are you still working on it?

kriszyp: Sorry, this is in progress, I forgot that this repo starts with things in ready state, I'm used to harperdb starting things in draft.

kriszyp: Now its ready for review.

PR #261, opened 2025-12-03
Transaction log querying

### LINKED ISSUE #218: Query Transaction Log

kriszyp:
Implements transaction log querying and iterating, fixes #218.

This is implemented using memory maps to read the data from transaction logs. This allows the iterating through a transaction log to be done entirely in JavaScript using data views and subarrays, for most iterations, with typical iterations taking around 50ns for returning the next entry (with a timestamp and buffer). The primary public API for querying and reading the transaction log is the `log.query(options)` method, which provides options for defining a `start` time, `end` time, and some additional options for the range of entries to match (including `exclusiveStart`). The `query` method returns an `Iterator<TransactionEntry>` (`& Iterable`). Notably, the returned `Iterator`/`Iterable` can be used multiple times to resume iteration, allowing one to resume and continue iterating through a transaction log after iterator has previously finished, but additional commits have been made, yielding more possible iterations. (This will be used heavily by replication and real-time subscriptions).

Querying to find a starting position is implemented in C++, by creating an on-demand index of each log file that will be searched. The index is constructed as an (ordered) `std::map` that can be queried in `O(log n)` time. The result of a query for a start time finds the log position for which there are no entries with an earlier timestamp before the returned position. Note that the transaction log is _not_ guaranteed to be ordered. It is roughly ordered, we write transactions chronologically, but it is possible for one transaction to take longer than another, in which case it will receive an earlier timestamp (at the start of the transaction), but will be written after a shorter transaction with a later starting timestamp. With the returned position, the iterator is responsible for filtering out any transactions that don't match the conditions (possibly due to out-of-order writes).

The concept of a transaction log "position" is used heavily throughout the querying and is represented by a 64-bit number where the high 32-bits are the log sequence number, and the lower 32-bits are the offset within a log file. Together, this 64-bit position provides a single absolute definition of a position withing a named log sequence. This is returned by the `findPosition` method to determine a starting point (returned as 64-bit float since that is the most efficient 64-bit primitive).

The "end" of the transaction log is also represented with a (64-bit) position. By default, the end of the transaction log is defined as the last position in the transaction log for which all previous transactions have been committed (to the RocksDB database). There may be additional transaction entries in transaction log that have been written but not yet committed. Typically we don't want to access these entries because doing so makes visible race conditions and inconsistencies between log entries and database values within transactions that may or may not be fully committed (breaking ACID compliance). This end position or "lastPositionCommitted" is determined by tracking the written and uncommitted transactions, removing them when they are committed and then finding the new last position after each commit completes. This position is made available to JavaScript through a shared external buffer.

A query can also be executed with a `readUncommitted` that can be used to access to transactions that have not been committed yet. This is particularly useful for replaying transaction logs on restart where many transactions that have been "committed" to RocksDB may have effectively become "uncommitted" after a process crashes/terminates because the data was not written before termination.

This PR also includes the basic mechanism for associating the transaction log position with database sequence numbers so that when a database SST file is "flushed", we can track how much of the transaction log has data that has been fully flushed to disk. This is critical for determining how much of the transaction log needs to be replayed on startup (any data that might not be flushed to disk). This mechanism's primary API is a `TransactionLogStore::databaseFlushed` method that can be called with the sequence number of a database flush so that https://github.com/HarperFast/rocksdb-js/issues/219 can be completed.

To no one's surprise, Windows throws in an extra element of fun. In Linux (and macos), we can create an arbitrarily large memory map, and as a file grows, the additional data in the memory map will be accessible (accessing data beyond the file size is a bus error). However, Windows requires that a memory map be fully backed by a full of the same (or larger) size. Well, you can actually get around this by directly hacking into the NT.DLL using driver-level APIs, but I assumed we probably didn't want to carry these techniques into this library:
https://github.com/kriszyp/lmdb-js/blob/master/dependencies/lmdb/libraries/liblmdb/mdb.c#L5998-L6012
So instead the file size is adjusted to match the memory map when a memory map is needed. This probably is good for performance, ensuring more linear disk usage.

### DIFF
--- a/benchmark/setup.ts
+++ b/benchmark/setup.ts
@@ -7,6 +7,7 @@ import { parentPort, Worker, workerData } from 'node:worker_threads';
 import { setImmediate as rest } from 'node:timers/promises';
 import { rm } from 'node:fs/promises';
 import { pathToFileURL } from 'node:url';
+const isWindows = process.platform === 'win32';
 
 const vitestBench = workerData?.benchmarkWorkerId ? () => {
 	throw new Error('Workers should not be directly calling vitest\'s bench()');
@@ -522,6 +523,9 @@ export function concurrent<T, U, S extends BenchmarkOptions<T, U>>(suite: S & Ha
 		async bench(ctx: BenchmarkContext<T>) {
 			await currentlyExecuting[index]; // await the previous execution for this slot
 			currentlyExecuting[index] = suite.bench(ctx) as Promise<void>; // let it execute concurrently and resolve after concurrencyMaximum number of executions
+			if (isWindows) { // don't do any concurrency on windows
+				return currentlyExecuting[index];
+			}
 			index = (index + 1) % concurrencyMaximum; // cycle in a loop
 			if (restEachTurn) { // let asynchronous actions have turns in the event queue, more realistic for most scenarios
 				await rest();


src/parse-transaction-log.ts
```diff
@@ -70,6 +70,11 @@ export function parseTransactionLog(path: string): TransactionLog {
 
 		while (fileOffset < size) {
 			const timestamp = read(8).readDoubleBE(0);
+			if (timestamp === 0) {
+				// if we encounter zero padding, we can stop reading the entries since the next entry will start at the next 8-byte boundary, which is the same as the current file offset.
+				size = fileOffset - 8;
+				break;
+			}
```
cb1kenobi: What's the point of this? `timestamp` should never be zero. Commenting out this change has no effect on the tests.

src/binding/transaction_log_file.h
```diff
@@ -33,13 +35,15 @@
 #include <sys/stat.h>
 
 #define TRANSACTION_LOG_TOKEN 0x574f4f46
+#define POSITION_OF_LOG_FILE_TIMESTAMP 5
```
cb1kenobi: I've been trying to name all the preprocessor macros with a prefix of the context, so something like this:

```suggestion
#define TRANSACTION_LOG_FILE_TIMESTAMP_POSITION 5
```

src/binding/transaction_log_file.cpp
```diff
@@ -187,4 +191,46 @@ void TransactionLogFile::writeEntriesV1(TransactionLogEntryBatch& batch, const u
...
 
+/**
+ * Find the position of a timestamp, specifically the position where there are no transactions with an earlier timestamp
+ * @param timestamp - the timestamp to find the position of
+ * @param mapSize - the size of the memory map to search in
+ * @return the position of the timestamp, or zero if comes before this logfile, or 0xFFFFFFFF if it comes after this logfile
+ */
+uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t mapSize) {
+	std::lock_guard<std::mutex> indexLock(this->indexMutex);
+	MemoryMap* memoryMap = getMemoryMap(mapSize);
```
cb1kenobi: nit: I know adding the `this->` is optional, but I found it super helpful to use this convention to quickly identify where to find the method/function.

```suggestion
	MemoryMap* memoryMap = this->getMemoryMap(mapSize);
```

src/binding/transaction_log_file_posix.cpp
```diff
@@ -62,6 +63,22 @@ void TransactionLogFile::openFile() {
 		this, this->path.string().c_str(), this->size)
 }
 
+MemoryMap* TransactionLogFile::getMemoryMap(uint32_t fileSize) {
+	if (!memoryMap) {
+		void* map = mmap(NULL, fileSize, PROT_READ, MAP_SHARED, fd, 0);
+		DEBUG_LOG("%p TransactionLogFile::getMemoryMap new memory map: %p\n", this, map);
+		if (!map) return nullptr;
```
cb1kenobi: This should be:

```suggestion
		if (map == MAP_FAILED) return nullptr;
```

src/binding/transaction_log_file_posix.cpp
```diff
@@ -62,6 +63,22 @@ void TransactionLogFile::openFile() {
 		this, this->path.string().c_str(), this->size)
 }
 
+MemoryMap* TransactionLogFile::getMemoryMap(uint32_t fileSize) {
+	if (!memoryMap) {
```
cb1kenobi:
```suggestion
	if (!this->memoryMap) {
```

src/binding/transaction_log_file_posix.cpp
```diff
@@ -128,6 +145,13 @@ int64_t TransactionLogFile::writeToFile(const void* buffer, uint32_t size, int64
 	return static_cast<int64_t>(::write(this->fd, buffer, size));
 }
 
+MemoryMap::MemoryMap(void* map, uint32_t mapSize) : map(map), mapSize(mapSize) {}
```
cb1kenobi: This constructor should go in `transaction_log_file.cpp` or inlined in `transaction_log_file.h`.

src/binding/transaction_log_file_windows.cpp
```diff
@@ -204,6 +297,13 @@ std::string getWindowsErrorMessage(DWORD errorCode) {
 	return message;
 }
 
+MemoryMap::MemoryMap(void* map, uint32_t mapSize) : map(map), mapSize(mapSize) {}
+
+MemoryMap::~MemoryMap() {
+	UnmapViewOfFile(map);
+	CloseHandle(mapHandle);
```
cb1kenobi:
```suggestion
	if (this->map != nullptr) {
		::UnmapViewOfFile(this->map);
	}
	if (this->mapHandle != INVALID_HANDLE_VALUE) {
		::CloseHandle(this->mapHandle);
	}
```

src/binding/transaction_log_file.h
```diff
@@ -125,6 +138,21 @@ struct TransactionLogFile final {
 	 */
 	void writeEntries(TransactionLogEntryBatch& batch, const uint32_t maxFileSize = 0);
 
+	/**
+	 * Return a memory map of the file and mark it as in use
+	 */
+	MemoryMap* getMemoryMap(uint32_t fileSize);
```
cb1kenobi: I really think this should return a `std::shared_ptr<MemoryMap>`. This pointer is leaving the safety of the `TransactionLogFile` instance, passed through the store to `TransactionLog::GetMemoryMapOfFile()`.

src/binding/transaction_log_file_windows.cpp
```diff
@@ -73,8 +73,101 @@ void TransactionLogFile::openFile() {
...
 	}
 	this->size = static_cast<size_t>(fileSize.QuadPart);
+	DEBUG_LOG(stderr, "File size: %zu file path: %s\n", this->size, this->path.string().c_str());
+	// On Windows, we have to create the full file size for memory maps, and it is zero-padded, so the act of indexing allows us to find
+	// the end, and adjust the real size accordingly.
+	// TODO: Future optimization is to only do this if the file is a multiple of the page size, and ensure
+	// files that are expanded to a memory page are memory page aligned, with (this->size & 0xFFF) == 0
+	if (this->size > 0) {
+		this->findPositionByTimestamp(0, this->size);
+		DEBUG_LOG("New file size: %zu file path: %s\n", this->size, this->path.string().c_str());
```
cb1kenobi:
```suggestion
		DEBUG_LOG("%p TransactionLogFile::openFile New file size: %zu file path: %s\n",
			this, this->size, this->path.string().c_str());
```

src/binding/transaction_log_file_windows.cpp
```diff
@@ -73,8 +73,101 @@ void TransactionLogFile::openFile() {
...
+	DEBUG_LOG("%p TransactionLogFile::getMemoryMap creating new memory map: %u\n", this, fileSize);
+	// In windows, we can not map beyond the size of the file (without using driver-level APIs that directly call procedures
+	// in NT.DLL). So we must expand the file to the full size before we can map it.
+	if (fileSize > this->size)
+	{
+		LARGE_INTEGER currentPos;
+		LARGE_INTEGER distanceToMove;
+		// First, we have to get the current position, so we can restore it (if we get to a point where no other code relies on position, could remove this)
+		distanceToMove.QuadPart = 0; // We want to move 0 bytes to query current position
+		if (!SetFilePointerEx(this->fileHandle, distanceToMove, &currentPos, FILE_CURRENT)) {
```
cb1kenobi: While not required, it's a good habit to specify the scope resolution operator to all external APIs.

```suggestion
		if (!::SetFilePointerEx(this->fileHandle, distanceToMove, &currentPos, FILE_CURRENT)) {
```

benchmark/transaction-log.bench.ts
```diff
@@ -0,0 +1,126 @@
...
+		});
+	});
+});
+function concurrent<C>(execute: (ctx: C) => Promise<void>, concurrencySlowdown = 100): (ctx: C) => Promise<void> {
+	let outstanding = 0;
+	let rejection;
+	return (ctx: C) => {
+		outstanding++;
+		execute(ctx).then(() => outstanding--, (error) => { rejection = error });
+		if (rejection) {
```
cb1kenobi: This will never be `true`.

benchmark/worker-transaction-log.bench.ts
```diff
@@ -0,0 +1,53 @@
...
+				const { db } = ctx;
+				return db.put(ctx.index++, data);
+			}),
+		});
+	});
+});
+function concurrent<C>(execute: (ctx: C) => Promise<void>, concurrencySlowdown = 100): (ctx: C) => Promise<void> {
+	let outstanding = 0;
+	let rejection;
+	return (ctx: C) => {
```
cb1kenobi: Same problem here as the other benchmark file.

package.json
```diff
@@ -21,7 +21,7 @@
     "build": "pnpm build:bundle && pnpm rebuild",
     "build:binding": "node-gyp build",
     "build:binding:debug": "node-gyp build --coverage --debug --verbose",
-    "build:bundle": "rimraf dist && tsdown -c tsdown.config.ts",
+    "build:bundle": "rimraf dist && tsdown -c tsdown.config.ts --sourcemap",
```
cb1kenobi: Instead of setting `--sourcemap` here, we should set `sourcemap: true` in the `tsdown.config.ts` file.

cb1kenobi:
The PR description says:
> [The position] is represented by a 64-bit number where the high 32-bits are the log sequence number, and the lower 32-bits are the offset within a log file

Is this true? I see `SequencePosition` which seems to do what you describe and frankly I like the struct.

cb1kenobi (changes requested):
Wow, this is awesome stuff! I need to spend a little more time going through the position related logic. I found a few things for you to chew on. :)

src/binding/transaction_log_file_posix.cpp
```diff
@@ -128,6 +145,13 @@ int64_t TransactionLogFile::writeToFile(const void* buffer, uint32_t size, int64
...
+
+MemoryMap::~MemoryMap()
+{
+	munmap(map, mapSize);
```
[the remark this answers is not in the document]
kriszyp: Is this really possible (when the constructor is guarded with a null check)? Or you are suggesting it is just good practice? (which is fine, happy to accept this suggestion, just making sure I am understanding).

src/binding/transaction_log_file_posix.cpp
```diff
@@ -62,6 +63,22 @@ void TransactionLogFile::openFile() {
...
+		DEBUG_LOG("%p TransactionLogFile::getMemoryMap new memory map: %p\n", this, map);
+		if (!map) return nullptr;
+		// If successful, return a MemoryMap object for tracking references.
+		// Note, that we do not need to do any cleanup from this class's
+		// destructor. Removing files that are memory mapped is perfectly fine,
+		// and the memory map can be safely used indefinitely (the file descriptor
+		// doesn't need to be kept open either).
+		memoryMap = new MemoryMap(map, fileSize);
+	}
+	memoryMap->fileSize = fileSize;
```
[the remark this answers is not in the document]
kriszyp: That's fine, can do, but the `fileSize` is still assigned on every access in case the (effective) file size changes (from a max memory for a log to a specific file size once the log is no longer active).

src/parse-transaction-log.ts
cb1kenobi: I did comment it out and tested it on macOS and the tests worked. I'll test removing it on Windows tomorrow after this branch is stable again.

src/binding/transaction_log_file.h
```diff
@@ -81,11 +85,20 @@ struct TransactionLogFile final {
...
+	 */
+	MemoryMap* memoryMap = nullptr;
+
 	/**
 	 * The mutex used to protect the file (open/close, read/write, etc).
 	 */
 	std::mutex fileMutex;
 
+	std::map<double, uint32_t> positionByTimestampIndex;
+	uint32_t lastIndexedPosition = POSITION_OF_LOG_FILE_TIMESTAMP;
```
cb1kenobi:
```suggestion
	uint32_t lastIndexedPosition = TRANSACTION_LOG_FILE_TIMESTAMP_POSITION;
```

src/binding/transaction_log_store.cpp
```diff
@@ -264,9 +354,61 @@ void TransactionLogStore::writeBatch(TransactionLogEntryBatch& batch) {
...
+	sequencePosition.position = fullyCommittedPosition;
+	sequencePosition.rocksSequenceNumber = rocksSequenceNumber;
+	// Now we record this in our array of sequence number + position combinations. However, we don't want to keep a huge
+	// array so we keep an array where each n position represents an n^2 frequencies of correlations. We are not keeping
+	// an exact map of every pairing, and we don't need to. We don't need to know the exact rocks sequence number, we just
+	// need a sequence number that is not greater than the point of the flush. But we want to record enough that we
+	// won't lose more than half of what has to be replayed since the last flush.
+	unsigned int count = nextSequencePositionsCount++;
+	int index = 0;
+	for (; index < 19; index++) {
```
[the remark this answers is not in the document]
kriszyp: Using a constant now, with some commentary, let me know if it is unclear.

benchmark/transaction-log.bench.ts
kriszyp: It is set in the rejection handler, so it can be truthy. But adjusted the code to hopefully be clearer about that.

benchmark/transaction-log.bench.ts
```diff
@@ -0,0 +1,134 @@
...
+});
+function concurrent<C>(execute: (ctx: C) => Promise<void>, concurrencySlowdown = 100): (ctx: C) => Promise<void> {
+	let outstanding = 0;
+	let previousRejection: Error;
+	return async (ctx: C) => {
+		outstanding++;
+		execute(ctx).then(() => outstanding--, (error) => {
+			previousRejection = error
+		});
+		if (previousRejection) {
```
cb1kenobi: This will never be true.

benchmark/transaction-log.bench.ts
```diff
@@ -0,0 +1,114 @@
+import { describe } from 'vitest';
+import { benchmark, concurrent } from './setup.js';
+
+describe('Transaction log', () => {
+	const ENTRY_COUNT = 1000;
+	const data = Buffer.alloc(100, 'a');
+
+	describe('write log with 100 byte records', () => {
+		benchmark('rocksdb', concurrent({
+			async setup(ctx) {
```
cb1kenobi: Need a type to make TS happy.

```suggestion
			async setup(ctx:  BenchmarkContext<RocksDatabase>) {
```

benchmark/setup.ts
```diff
@@ -507,3 +508,28 @@ function withResolvers<T>() {
 		promise
 	};
 }
+
+export function concurrent<T, U, S extends BenchmarkOptions<T, U>>(suite: S & HasConcurrencyOptions): S {
```
cb1kenobi: This is looking good, but now you have a ton of TypeScript issues. Just like `benchmark()` above, you need explicit method signatures for the two database types (rocks and lmdb).

```suggestion
export function concurrent(suite: BenchmarkOptions<RocksDatabase, RocksDatabaseOptions> & HasConcurrencyOptions): BenchmarkOptions<RocksDatabase, RocksDatabaseOptions>;
export function concurrent(suite: BenchmarkOptions<LMDBDatabase, lmdb.RootDatabaseOptions> & HasConcurrencyOptions): BenchmarkOptions<LMDBDatabase, lmdb.RootDatabaseOptions>;
export function concurrent<T, U, S extends BenchmarkOptions<T, U>>(suite: S & HasConcurrencyOptions): S {
```

You also need to export the `LMDBDatabase` and `BenchmarkContext<T>` types at the top of the file because the bench files need them.

benchmark/transaction-log.bench.ts
```diff
@@ -0,0 +1,114 @@
...
+				ctx.last = 0;
+			},
+			bench(ctx) {
+				const { db, last } = ctx;
+				for (let i = 0; i < 100; i++) {
+					for (let { key, value } of db.getRange({ start: ctx.last })) {
+						ctx.last = key + 1;
+					}
+				}
+				return db.put(ctx.index++, data) as Promise<void>;
```
cb1kenobi: Apparently lmdb-js' `put()` wants a string. Also need to force it to return a `Promise<void>`.

```suggestion
				return db.put(String(ctx.index++), data) as unknown as Promise<void>;
```

benchmark/worker-transaction-log.bench.ts
```diff
@@ -0,0 +1,38 @@
...
+
+	describe('write log with 100 byte records', () => {
+		benchmark('rocksdb', concurrent({
+			numWorkers: 4,
+			async setup(ctx) {
+				const db = ctx.db;
+				const log = db.useLog('0');
+				ctx.log = log;
+			},
+			bench({ db, value, log, start, duration }) {
```
cb1kenobi: Unused vars

```suggestion
			bench({ db, log }) {
```

benchmark/setup.ts
```diff
@@ -507,3 +508,28 @@ function withResolvers<T>() {
...
+		},
+		async teardown(ctx: BenchmarkContext<T>) {
+			// wait for all outstanding executions to finish
+			await Promise.all(currentlyExecuting);
+			// any more tearing down
+			return suite.teardown?.(ctx);
+		}
+	}
+}
+type HasConcurrencyOptions = { concurrencyMaximum?: number, restEachTurn?: boolean };
```
cb1kenobi: Missing `numWorkers`:

```suggestion
type HasConcurrencyOptions = {
	concurrencyMaximum?: number;
	numWorkers?: number;
	restEachTurn?: boolean;
};
```

benchmark/worker-transaction-log.bench.ts
```diff
@@ -0,0 +1,38 @@
+import {
+	concurrent,
+	workerDescribe as describe,
+	workerBenchmark as benchmark
+} from './setup.js';
+
+describe('Transaction log with workers', () => {
```
cb1kenobi: I was running this test only and we have an uncaught exception:
```
 ❯ benchmark/transaction-log.bench.ts [queued]

 Test Files 0 passed (2)
      Tests 0 passed (0)
   Start at 16:15:05
   Duration 0ms
libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in last_write_time: No such file or directory ["/var/folders/cv/dbyf82nx55x3gy5l5s9yny1r0000gp/T/rocksdb-benchmark-1180b3a3841f193b/transaction_logs/0/1.txnlog"]
 ELIFECYCLE  Command failed.
[1]    118 abort      pnpm bench transaction-log
```

It's happening in `TransactionLogStore::load()` when getting the log file's mtime. The retention check code needs to be moved inside the try/catch like this:
```
			try {
				sequenceNumber = std::stoul(sequenceNumberStr);

				// check if the file is too old
				if (retentionMs.count() > 0) {
					auto mtime = std::filesystem::last_write_time(filePath);
					auto mtime_sys = convertFileTimeToSystemTime(mtime);
					auto now = std::chrono::system_clock::now();
					auto fileAgeMs = std::chrono::duration_cast<std::chrono::milliseconds>(now - mtime_sys);
					auto delta = fileAgeMs - retentionMs;

					if (delta.count() > 0) {
						// file is too old, remove it
						DEBUG_LOG("%p TransactionLogStore::load File \"%s\" age=%lldms, expired %lldms ago, purging\n",
							store.get(), filePath.filename().string().c_str(), fileAgeMs.count(), delta.count())
						try {
							std::filesystem::remove(filePath);
						} catch (const std::filesystem::filesystem_error& e) {
							DEBUG_LOG("%p TransactionLogStore::load Failed to remove expired file %s: %s\n",
								store.get(), filePath.string().c_str(), e.what())
						}
						continue;
					} else {
						DEBUG_LOG("%p TransactionLogStore::load File \"%s\" age=%lldms, not expired, %lldms left\n",
							store.get(), filePath.filename().string().c_str(), fileAgeMs.count(), delta.count() * -1)
					}
				}

				store->registerLogFile(filePath, sequenceNumber);
			} catch (const std::filesystem::filesystem_error& e) {
				DEBUG_LOG("%p TransactionLogStore::load Failed to get last write time for file %s: %s\n",
					store.get(), filePath.string().c_str(), e.what())
			} catch (const std::exception& e) {
				DEBUG_LOG("%p TransactionLogStore::load Failed to load file %s: %s\n",
					store.get(), filePath.string().c_str(), e.what())
			} catch (...) {
				DEBUG_LOG(
					"DBDescriptor::discoverTransactionLogStores Invalid sequence number in file: %s\n",
					filename.c_str()
				)
			}
```

src/binding/transaction_log_file_posix.cpp
```diff
@@ -100,6 +100,12 @@ int64_t TransactionLogFile::readFromFile(void* buffer, uint32_t size, int64_t of
 bool TransactionLogFile::removeFile() {
 	std::unique_lock<std::mutex> lock(this->fileMutex);
 
+	if (this->memoryMap) {
+		DEBUG_LOG("%p TransactionLogFile::removeFile Releasing memory map before removing file: %s\n",
+			this, this->path.string().c_str())
+		this->memoryMap.reset();
```
kriszyp: @cb1kenobi I thought this would happen automatically in the destructor (and doesn't that happen right after removing the file)? Curious of why it would need to be done here.

test/lib/util.ts
```diff
@@ -82,30 +82,26 @@ export async function dbRunner(
 			}
 			databases.push({ db, dbPath: path });
 		}
-		await testFn(...databases);
```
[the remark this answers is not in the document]
cb1kenobi: Brings whole new meaning to test cleanup. Whoops!

Issue #246, opened 2025-11-21
Automate cleanup, compaction, and flush on exit


kriszyp:
Currently exiting a process (via SIGINT, SIGTERM, process.exit()) will not automatically trigger the cleanup of the databases and the associated final compaction/flush. Currently if applications are not diligent about calling `close()`, there is unwritten in-memory KV data on exit. We could rely on applications to hopefully maybe call `close()`, but that puts extra effort and complication on the application. I think this library should be able to automate this, exposing an `onExit` function that can be registered (in JS) with `process.onExit(onExit)`, and internally can ensure that all databases are properly destroyed, going through the compaction to ensure all data is written prior to exiting. We should also ensure that the final write in DBDeleter also calls `db.Flush()` to ensure data is actually on disk and that the eventual `OnFlushCompleted` callback (#219 ) has finished (allowing transaction log to be fully synchronized).

kriszyp:
> I like the idea of onExit, but I'm not a fan of the name. Maybe closeAll() or shutdown()?

Yes, I like your names better.

Issue #235, opened 2025-11-13
Add method for writing data in memory to disk


kriszyp:
Basically, I think we want to expose the `db->CompactRange(options, nullptr, nullptr)` method, or something similar, so we can proactively tell RocksDB to write all in-memory data to disk and flush it.

PR #268, opened 2025-12-10
Flush events

kriszyp:
This implements listeners for flush events to ensure that transaction logs are flushed prior to database writes and we track the last flush to know which transaction logs need to be replayed on restart (using `startFromLastFlush`), resolving #219, #246, and #235.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Opens the database at the given path. This must be called before performing
 any data operations.
 
 ```typescript
-import { RocksDatabase } from '@harperdb/rocksdb-js';
+import { RocksDatabase } from '@harperfast/rocksdb-js';
 
 const db = new RocksDatabase('path/to/db');
 db.open();
@@ -329,7 +329,7 @@ are automatically committed. However, if an error is thrown during the
 callback, all database operations will be rolled back.
 
 ```typescript
-import type { Transaction } from '@harperdb/rocksdb-js';
+import type { Transaction } from '@harperfast/rocksdb-js';
 await db.transaction(async (txn: Transaction) => {
 	await txn.put('foo', 'baz');
 });
@@ -364,7 +364,7 @@ Inside a synchronous transaction, use `getSync()`, `putSync()`, and
 `removeSync()`.
 
 ```typescript
-import type { Transaction } from '@harperdb/rocksdb-js';
+import type { Transaction } from '@harperfast/rocksdb-js';
 db.transactionSync((txn: Transaction) => {
 	txn.putSync('foo', 'baz');
 });
@@ -817,7 +817,7 @@ Returns an object containing all of the information in the log file.
 
 The `shutdown()` will flush all in-memory data to disk and wait for any outstanding compactions to finish, for all open databases. It is highly recommended to call this in a `process` `exit` event listener (on the main thread), to ensure that all data is flushed to disk before the process exits:
 ```typescript
-import { shutdown } from '@harperdb/rocksdb-js';
+import { shutdown } from '@harperfast/rocksdb-js';
 process.on('exit', shutdown);
 ```
 
@@ -856,7 +856,7 @@ To use it, extend the default `Store` and pass in an instance of your store
 into the `RocksDatabase` constructor.
 
 ```typescript
-import { RocksDatabase, Store } from '@harperdb/rocksdb-js';
+import { RocksDatabase, Store } from '@harperfast/rocksdb-js';
 
 class MyStore extends Store {
   get(context, key, resolve, reject, txnId) {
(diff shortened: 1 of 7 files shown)


src/binding/db_registry.cpp
```diff
@@ -208,6 +208,45 @@ void DBRegistry::PurgeAll() {
...
+				if (entry.descriptor) {
+					descriptorsToClose.push_back(entry.descriptor);
+				}
+			}
+		}
+
+		// Close all descriptors without holding the lock
+		for (auto& descriptor : descriptorsToClose) {
+			DEBUG_LOG("%p DBRegistry::Shutdown Closing database: %s\n", instance.get(), descriptor->path.c_str())
+			descriptor->closing.store(true);
```
cb1kenobi: I'm inclined to add a `close()` method so that we don't leak these internal things such as `closing` and `db`. I'm surprised we don't already have a `close()` method. I probably did have one at some point.

What about closing all of the `DBDescriptor`'s active `DBHandles` and closing the transaction log stores?

test/transaction-optimistic.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import { RocksDatabase, Transaction } from '../src/index.js';
+import { dbRunner } from './lib/util.js';
+describe('Optimistic Transaction Specific Behavior', () => {
+	it(`async should be able to resume a conflicted transaction`, () => dbRunner({
+	}, async ({ db }) => {
+		await db.put('foo', 'bar1');
+
+		setTimeout(() => db?.put('foo', 'bar2'));
+
+		let transaction = new Transaction(db.store);
```
cb1kenobi: nit: consider `const`?

```suggestion
		const transaction = new Transaction(db.store);
```

README.md
```diff
@@ -744,6 +760,7 @@ Returns an iterable/iterator that streams all log entries for the given filter.
...
   - `readUncommitted?: boolean` When `true`, this will include uncommitted transaction entries. Normally transaction entries that haven't finished committed are not included. This is particularly useful for replaying transaction logs on startup where many entries may have been written to the log but are no longer considered committed if they were not flushed to disk.
+  - `startFromLastFlush?: boolean` When `true`, this will only match transactions that have been flushed from RocksDB's memtables to disk (and are within any provided `start` and `end` filters, if included). This is useful for replaying transaction logs on startup where many entries may have been written to the log but are no longer considered committed if they were not flushed to disk.
```
cb1kenobi: The code uses the name "startFromLastFlushed":

```suggestion
  - `startFromLastFlushed?: boolean` When `true`, this will only match transactions that have been flushed from RocksDB's memtables to disk (and are within any provided `start` and `end` filters, if included). This is useful for replaying transaction logs on startup where many entries may have been written to the log but are no longer considered committed if they were not flushed to disk.
```

src/load-binding.ts
```diff
@@ -193,3 +197,5 @@ export const NativeIterator: typeof NativeIteratorCls = binding.Iterator;
 export const NativeTransaction: NativeTransaction = binding.Transaction;
 export const TransactionLog: TransactionLog = binding.TransactionLog;
 export const version: string = binding.version;
+
+process.on('exit', binding.shutdown);
```
cb1kenobi: Hmm, I'm not a huge fan of libraries listening to `on('exit')`. I would prefer we document in the readme that users of this library should add this line to their app.

src/binding/transaction_log_store.cpp
```diff
@@ -68,6 +69,11 @@ void TransactionLogStore::close() {
 		logFile->close();
 	}
 
+	// Close the state file if it's open
+	if (flushedStateFile.is_open()) {
+		flushedStateFile.close();
```
cb1kenobi: Silly question, is the state good after we close the store? Should we delete the state file?

```cpp
// untested
std::filesystem::remove(this->path / "txn.state");
```

kriszyp: The main purpose of the `txn.state` file is to hold the last known flushed point in the transaction log, so that on a restart we can replay any transactions that haven't been flushed. So it is important that this file continue to exist and be accessible the next time the database and transaction log is opened.

src/binding/transaction_log_file_posix.cpp
```diff
@@ -31,6 +31,47 @@ void TransactionLogFile::close() {
 	}
 }
 
+void TransactionLogFile::flush() {
+	int fdToFlush = -1;
+	uint32_t currentSize = 0;
+
+	{
+		std::unique_lock<std::mutex> lock(this->fileMutex);
```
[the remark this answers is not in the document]
kriszyp: I think I did what you suggested.

src/binding/db_registry.cpp
kriszyp: Do you want to do that and handle the `CloseDB()` path?

src/binding/transaction_log_store.cpp
cb1kenobi: That's what I figured. Sounds good!

PR #277, opened 2025-12-16
Cleanup and re-enable ongoing concurrent benchmarking

kriszyp:
The cleans up some unnecessary shared pointer resets and re-enables the ongoing concurrency of the benchmark tests which is more realistic (than starting all the actions at the same time and waiting for all to complete before the next iteration). Which seems to all work fine.


cb1kenobi:
I don't think skipping Windows is a good idea because this PR actually segfaults on Windows. I think I know why. Stay tuned.

PR #282, opened 2025-12-18
Publish workflow

### LINKED ISSUE #200: Publish to npm workflow
Add a workflow that on release will builds the native binding for each target, publishes those prebuilds as separate packages, then publishes rocksdb-js to npm.

When publishing the prebuilds, ensure the prebuild’s `package.json` has the proper `"os"` and `"cpu"` settings:

```json
{
  "name": "@harperdb/rocksdb-js-linux-x64",
  "os": ["linux"],
  "cpu": ["x64"]
}
```

rocksdb-js will list all prebuilt packages as an optional dependency just like TurboRepo: https://github.com/vercel/turborepo/blob/v2.5.0/packages/turbo/package.json#L20-L27. npm will install the appropriate optional dependency.

We will need to extend `loadBinding()` to find the prebuild in the `node_modules`. The path is known. It’s just a matter of looking if it exists.

Also, research a way to semver bump the release prior to publishing.

cb1kenobi:
This new publish workflow will:
1. Perform preflight version checks
2. Build and publish platform-specific bindings
3. Build and publish the main `rocksdb-js` package

The binding is built for the following:

| OS | Arch | Package |
| --- | --- | --- |
| Linux | arm64 | `@harperfast/rocksdb-js-linux-arm64` |
| Linux | x64 | `@harperfast/rocksdb-js-linux-x64` |
| macOS | arm64 | `@harperfast/rocksdb-js-darwin-arm64` |
| macOS | x64 | `@harperfast/rocksdb-js-darwin-x64` |
| Windows | arm64 | `@harperfast/rocksdb-js-win32-arm64` |
| Windows | x64 | `@harperfast/rocksdb-js-win32-x64` |

After each binding is built, the workflow does a quick smoke test to make sure the binding actually works.

To create a new release, do:
1. Make sure the `version` and `optionalDependencies` versions are correct
4. Go to https://github.com/HarperFast/rocksdb-js/releases/new
5. Create a tag `v<major>.<minor>.<patch>`
6. Set release title to `v<major>.<minor>.<patch>`
7. Click "Generate release notes" button
8. Click "Publish release" button

If a binding fails to build or publish, then the `@harperfast/rocksdb-js` package will NOT be published. If NO packages were successfully built, go to https://github.com/HarperFast/rocksdb-js/releases and delete the release, then go to https://github.com/HarperFast/rocksdb-js/tags and delete the tag. If any packages do successfully publish and one or more packages fail (regardless of reason), then the version/tag must be abandoned, bump versions in `package.json`, and a new release must be created.

Fixes #200.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   preflight:
     name: Preflight
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
@@ -58,7 +58,7 @@ jobs:
           - runner: ubuntu-24.04-arm
             name: linux-arm64
             target: arm64
-          - runner: ubuntu-22.04
+          - runner: ubuntu-latest
             name: linux-x64
             target: x64
           - runner: windows-11-arm
@@ -108,7 +108,7 @@ jobs:
 
   publish:
     name: Publish packages
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     needs: build
     steps:
       - name: Checkout repository
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         node: [20, 22, 24]
-        os: [ubuntu-22.04, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
       - name: Checkout repository
@@ -81,7 +81,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
       - name: Checkout repository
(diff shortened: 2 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,188 @@
...
+          - runner: macos-latest
+            name: darwin-arm64
+            target: arm64
+          - runner: macos-latest
+            name: darwin-x64
+            target: x64
+          - runner: ubuntu-24.04-arm
+            name: linux-arm64
+            target: arm64
+          - runner: ubuntu-latest
```
kriszyp: The latest? I believe the platform we compile on usually binds to a glibc version and dictates the minimum glibc version. And compiling on ubuntu v24 would mean these binaries require at least glibc 2.39, which would exclude any earlier ubuntu versions (which are very commonly used). I thought we would want to compile on the absolute oldest OS versions we can possibly use; ubuntu v20 would give 2.31. In lmdb-js I have been compiling with the `node:18` container, which uses debian v11, which also uses glibc 2.31.
I believe most of our current containers use Ubuntu v22. However, IIRC AWS's linux is often using a much older glibc version.
Anyway, we could certainly try this and see if it works. Maybe the bindings are different here.

cb1kenobi: I'm basing this entirely off what we use to build RocksDB: https://github.com/HarperFast/rocksdb-prebuilds/blob/main/.github/workflows/build.yml#L51-L62.

Ubuntu 20.04 LTS is no longer supported as of May 31, 2025.

PR #291, opened 2025-12-26
Reenable unit tests

kriszyp:
https://github.com/HarperFast/rocksdb-js/pull/261/changes/b336ce240d814dfe0af593f8bbf586fb742d5272 had skipped all the test functions in our test suite. This PR re-enables the tests.
After the tests had been disabled, https://github.com/HarperFast/rocksdb-js/pull/261/changes/7b23d0eed1bb800e944c218f8d093c6582016351 had broken the transaction log tests (if they were running, but they weren't, so we didn't see this). This PR reverts that change (I didn't investigate closely as to the exact cause).
(also, removed an inadvertent test file addition).
We can certainly try to reapply the changes in https://github.com/HarperFast/rocksdb-js/commit/7b23d0eed1bb800e944c218f8d093c6582016351, but obviously with unit tests running and really passing.


src/transaction-log-reader.ts
```diff
@@ -4,7 +4,7 @@ import {
 	constants,
 	TransactionLogQueryOptions,
 	TransactionEntry,
-} from './load-binding.js';
+} from './load-binding';
```
cb1kenobi: Why this change?

cb1kenobi (approved):

Issue #285, opened 2025-12-19
Verify blob usage


kriszyp:
Our benchmarks of 10MB writes strongly suggest that blobs aren't being properly utilized for storage of large records, as they should. In addition the benchmark seems to only produce .sst files, and I believe there is a distinct file extension for the database files that house blobs. We should create a test that verifies that after writing a 10KB record to the database, that the `rocksdb.num-blob-files` database property increases. This is dependent on #256 to access that database property.

Issue #256, opened 2025-12-01
Add `database.getDBProperty(propertyName)` method for accessing DB properties


kriszyp:
RocksDB's DB interface has a `GetProperty` and `GetIntProperty` methods for accessing database properties and statistics. We should generically make this available for accessing this information. Users should be able to access properties like:
```
levelStats = database.getDBProperty('rocksdb.levelstats');
```
Some of the properties are available as integers through the `GetIntProperty` method. If we can't automatically determine the correct method to call, we could have a separate `getDBIntProperty` to access these properties:
```
blobFiles = database.getDBIntProperty('rocksdb.num-blob-files');
```
This can basically be implemented like `GetOldestSnapshotTimestamp`, but have the property name be passed in as an argument:
https://github.com/HarperFast/rocksdb-js/blob/main/src/binding/database.cpp#L385-L404

kriszyp:
Fixed in https://github.com/HarperFast/rocksdb-js/pull/292

PR #292, opened 2025-12-26
Get db property

kriszyp:
Add support for getting database properties #256.
With this in place, test for blob file creation, and fix blob file creation by setting the blob configuration on the column families #285.


cb1kenobi (approved):
Found a couple of nits, but overall this is awesome!

Issue #255, opened 2025-11-30
Support reusable buffers


kriszyp:
`napi_*_buffer_*`, and `napi_*_named_property` methods are very expensive, we would like to provide a way to avoid calling these methods for the majority of CRUD operations. Specifically, we would like to be able to reuse buffers for writing keys and values, and when a previously used buffer contains a key or value, the `put`, `delete`, `get`, and `addEntry` functions can utilize the already determined address of the buffer rather than having to calculate it again with these methods. (We have discussed a number of different ways of communicating this through the native layer, but the general goal of this ticket is ensuring some mechanism can be used to avoid these calls when possible).

Issue #197, opened 2025-10-28
Use `PinnableSlice`


cb1kenobi:
The `PinnableSlice`s are intended to provide zero-copy access to values from `Get` operations, with performance improvements. We should try to leverage this functionality, which could be beneficial for larger values.

PR #294, opened 2025-12-27
Use shared buffers for transferring keys and values for fast get/getSync

kriszyp:
This addresses #255 for `get`s.
According to the benchmarks, this improves the performance of `getSync` calls by about 2x and `get` calls by about 2.5x (at least in the common case of cached values).
With `getSync`, the breakdown of time spent is now about 420ns spent on the RocksDB side, and 240ns spent in this library (bindings and JS). Since it was about twice as slow before, I think it was previously spending about 800ns in this library, so really the performance gain on our side is 3-4x (and even more with `get`). At this point, RocksDB is dominating the time spent, so I think we can consider `get` to be pretty well optimized. This still isn't as fast as LMDB. However, once we get a better cache layer implemented, we might be competitive.
This also fixes an issue with tracking the key in the `AsyncGetState`. Previously this was held by a RocksDB `Slice`, which does not copy or own the target reference, and so it could change before the async task was called. The key is now safely copied into a `std::string`. The concurrent get test I added fails on main due to this problem, but is corrected in this branch.
This also minimizes the amount of value copying by using `PinnableSlice` for `get`s (#197).
There are still a significant number of places that would benefit from using these patterns, including `put`, `remove`, `attemptLock`/`unlock`, and `getUserSharedBuffer`.
The native layer provides a means for larger keys by allowing buffers to be passed directly in.

### DIFF
--- a/src/binding/database.cpp
+++ b/src/binding/database.cpp
@@ -21,13 +21,13 @@ namespace rocksdb_js {
  * ```
  */
 napi_value Database::Constructor(napi_env env, napi_callback_info info) {
-	NAPI_CONSTRUCTOR_WITH_DATA("Database")
+	NAPI_CONSTRUCTOR_WITH_DATA("Database");
 
 	// create shared_ptr on heap so it persists after function returns
 	napi_ref exportsRef = reinterpret_cast<napi_ref>(data);
 	auto* dbHandle = new std::shared_ptr<DBHandle>(std::make_shared<DBHandle>(env, exportsRef));
 
-	DEBUG_LOG("Database::Constructor Creating NativeDatabase DBHandle=%p\n", dbHandle->get())
+	DEBUG_LOG("Database::Constructor Creating NativeDatabase DBHandle=%p\n", dbHandle->get());
 
 	try {
 		NAPI_STATUS_THROWS(::napi_wrap(
@@ -35,7 +35,7 @@ napi_value Database::Constructor(napi_env env, napi_callback_info info) {
 			jsThis,
 			reinterpret_cast<void*>(dbHandle),
 			[](napi_env env, void* data, void* hint) {
-				DEBUG_LOG("Database::Constructor NativeDatabase GC'd dbHandle=%p\n", data)
+				DEBUG_LOG("Database::Constructor NativeDatabase GC'd dbHandle=%p\n", data);
 				auto* dbHandle = static_cast<std::shared_ptr<DBHandle>*>(data);
 				if (*dbHandle) {
 					DBRegistry::CloseDB(*dbHandle);
@@ -64,8 +64,8 @@ napi_value Database::Constructor(napi_env env, napi_callback_info info) {
  * ```
  */
 napi_value Database::Clear(napi_env env, napi_callback_info info) {
-	NAPI_METHOD_ARGV(3)
-	UNWRAP_DB_HANDLE_AND_OPEN()
+	NAPI_METHOD_ARGV(3);
+	UNWRAP_DB_HANDLE_AND_OPEN();
 
 	napi_value resolve = argv[0];
 	napi_value reject = argv[1];
@@ -76,11 +76,11 @@ napi_value Database::Clear(napi_env env, napi_callback_info info) {
 		"database.clear",
 		NAPI_AUTO_LENGTH,
 		&name
-	))
+	));
 
 	auto state = new AsyncClearState(env, *dbHandle);
-	NAPI_STATUS_THROWS(::napi_create_reference(env, resolve, 1, &state->resolveRef))
-	NAPI_STATUS_THROWS(::napi_create_reference(env, reject, 1, &state->rejectRef))
+	NAPI_STATUS_THROWS(::napi_create_reference(env, resolve, 1, &state->resolveRef));
+	NAPI_STATUS_THROWS(::napi_create_reference(env, reject, 1, &state->rejectRef));
 
 	NAPI_STATUS_THROWS(::napi_create_async_work(
 		env,       // node_env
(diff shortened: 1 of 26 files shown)


src/binding/database.cpp
```diff
@@ -305,26 +306,10 @@ napi_value Database::Get(napi_env env, napi_callback_info info) {
...
-	// try to get the value from the block cache
-	std::string value;
-	rocksdb::Status status = (*dbHandle)->descriptor->db->Get(
-		readOptions,
-		(*dbHandle)->column.get(),
-		keySlice,
-		&value
-	);
-
-	if (!status.IsIncomplete()) {
```
cb1kenobi: Don't we still want this if the key is not using the shared buffer?

src/store.ts
```diff
@@ -660,6 +695,27 @@ export class Store {
 	}
 }
 
+/**
+ * Ensure that they key has been copied into our shared buffer, and return the ending position
+ * @param keyBuffer
+ */
+function getKeyParam(keyBuffer: BufferWithDataView): number | Buffer {
+	if (keyBuffer.buffer === KEY_BUFFER_ARRAY_BUFFER) {
+		if (keyBuffer.end >= 0) {
```
cb1kenobi: Is `end` ever negative? I thought it was either a positive integer or `undefined`.

src/binding/util.cpp
```diff
@@ -436,6 +436,27 @@ const char* getNapiBufferFromArg(
 	return data;
 }
 
+const rocksdb::Slice getSliceFromArg(napi_env env, napi_value arg, char* defaultBuffer, const char* errorMsg) {
+	int32_t length;
+	char* data;
+	napi_status argStatus = ::napi_get_value_int32(env, arg, &length);
```
cb1kenobi: It's not clear here, in `load-binding.ts` for the `get()`/`getSync()` methods (both db and txn), or in the native implementation of those methods that when `key` is a number, it's actually the length of the shared buffer to use.

I think this is pretty important to note. Can we mentioned this in comment docs before `getSliceFromArg()` declaration in `utils.h`. Furthermore, I'd argue we should also document this in the `Database::Get()` and `Database::GetSync()` docs in `database.cpp` as well as `Transaction::Get()` and `Transaction::GetSync()` in `transaction.cpp`.

src/store.ts
```diff
@@ -22,8 +23,12 @@ import { DBIterator, type DBIteratorValue } from './dbi-iterator.js';
 import { Transaction } from './transaction.js';
 import { ExtendedIterable } from '@harperfast/extended-iterable';
 import { parseDuration } from './util.js';
-
+const { ONLY_IF_IN_MEMORY_CACHE_FLAG, NOT_IN_MEMORY_CACHE_FLAG, ALWAYS_CREATE_BUFFER_FLAG } = constants;
 const KEY_BUFFER_SIZE = 4096;
+export const KEY_BUFFER: BufferWithDataView = createFixedBuffer(KEY_BUFFER_SIZE);
```
cb1kenobi: Does the `KEY_BUFFER` have to be global? It seems to me this should live in the `Store` class instance? By keep it global, all database instances (regardless of column family) will share the same buffer. I get that an instance is bound to a single process and we shouldn't need to worry about race conditions, but I believe that coordinating async/await operations across database instances introduce unnecessary complexity.

src/binding/util.cpp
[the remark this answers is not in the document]
kriszyp: I do like this pattern, `napi_get_value_int32` does a type check for us anyways, so it seems like it is convenient and seemingly faster to make a single call to try to get the number and have it tell us if we should try a different type, especially here where the majority of the time the input value will be a number.

src/binding/database.cpp
```diff
@@ -618,6 +624,59 @@ napi_value Database::GetSync(napi_env env, napi_callback_info info) {
...
+ * *extremely* expensive, and by using a single shared buffer, we can avoid the overhead of buffer creation, and instead
+ * copy directly from the database to the shared buffer. So this sets the value buffer that will be used for transferring
+ * smaller values to and from JavaScript to C++. Note that we generally still use new buffers for larger values, as the
+ * overhead of buffer creation is smaller compared to the cost of the copying of data.
+ */
+napi_value Database::SetDefaultValueBuffer(napi_env env, napi_callback_info info) {
+	NAPI_METHOD_ARGV(1)
+	UNWRAP_DB_HANDLE()
+
+	if (argv[0] == nullptr) {
```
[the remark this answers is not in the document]
kriszyp: > napi_get_buffer_info() would return a not-ok status and NAPI_STATUS_THROWS would throw

Yeah, I like leveraging this native functionality, seems like a clean single call.

(I think addressed the other points).

src/store.ts
cb1kenobi: I don't think allocating 80KB for each db instance one time is that big of a deal when you're running on machines with many gigabytes of memory. We probably waste more memory than that else where on logging or something.

Not sure I buy into the L1/L2 cache logic. The L1/L2 cache size is so small that I'm sure there's other V8/Node stuff that will boot the shared buffer out to L3.

I am going to shelf my potential global shared buffer race condition concerns and resolving this for now.

src/binding/database.cpp
cb1kenobi: Ah, I missed that in the review. I didn't see async `get()` first tries `context.getSync()`. Not sure if the overhead of creating the promise is any worse than querying RocksDB twice, but it's probably tiny.

tsdown.config.ts
```diff
@@ -11,7 +11,7 @@ const config: UserConfig = defineConfig({
 	],
 	entry: './src/index.ts',
 	format: ['es', 'cjs'],
-	minify: !process.env.SKIP_MINIFY,
+	minify: Boolean(process.env.MINIFY),
```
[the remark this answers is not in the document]
kriszyp: Well, I did it yesterday, but doing it again today :shrug:

kriszyp:
> It would be a good idea to had a couple tests for `setDefaultKeyBuffer()` and `setDefaultValueBuffer()`.

Well, the tests definitely test this, it is executed as part of creating a database, and the buffers are heavily used for the the gets. I am not sure what other visible behavior of we would want to test. What do you have in mind?

PR #299, opened 2026-01-02
Don't minify (by default)

kriszyp:
I think we originally thought minification would improve load speed. It turns out that it does not. I measured load time with and without minification, and it is consistently 12-13ms regardless of minification. Any impact of minification is less than statistical noise, less than a millisecond, and negligible. Avoiding minification makes it much easier to debug, and doesn't seem to incur any performance regresssion.
I did still include a script for doing minification, but it isn't used by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "build:binding": "node-gyp build",
     "build:binding:debug": "node-gyp build --coverage --debug --verbose",
     "build:bundle": "node scripts/clean-dist.mjs && tsdown -c tsdown.config.ts",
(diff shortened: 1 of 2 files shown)


kriszyp:
> How did you benchmark this?

`let start = performance.now(); require('.'); console.log(performance.now() - start);`

cb1kenobi (approved):
No worries, let's merge it!

kriszyp:
oh, ok, I was just going to close it, I thought you didn't want this, but sure, cool!

PR #304, opened 2026-01-07
Minifed vs unminified bundle benchmarks

cb1kenobi:
Run `./compare-bundles.sh` on Mac/Linux, `./compare-bundles.cmd` on Windows.

It will build a minified bundle, then benchmark it 50 times. Afterwards, it does a unminified bundle and runs the benchmark 50 times.

Based on my Mac M4 Pro, the minified bundle is consistently 0.248 to 0.29 ms faster than the unminified bundle, which makes logical sense.

On Windows, I'm getting mixed results. Sometimes the unminified bundle is 0.168 to 3.167 ms faster than the minified bundle. Other times the minified bundle is 0.115 to 0.903 ms faster than the unminified bundle.

On WSL, I'm also getting mixed results. Sometimes the unminified is .249 to 1.935 ms faster. Other times the minified bundle is 0.006 ms faster.

Note: This PR is not meant to be merged.

### DIFF
--- a/compare-bundles.cmd
+++ b/compare-bundles.cmd
@@ -1,0 +1,17 @@
+@echo off
+setlocal
+
+set "ITERATIONS=%~1"
+if "%ITERATIONS%"=="" set "ITERATIONS=50"
+
+rem Prefer PowerShell 7+ (pwsh) if available, otherwise fall back to Windows PowerShell.
+where pwsh >nul 2>nul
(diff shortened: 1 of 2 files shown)


kriszyp:
> consistently 0.29 ms to 0.248 ms than

I realized I actually don't what difference you saw, I think this is missing the "greater" or "lesser" :)

Issue #205, opened 2025-10-28
Investigate more efficient range count


cb1kenobi:
The correct range record counter loops over the entire iterator. This is slow and not ideal. There may be a better way to estimate this:

```cpp
db->GetIntProperty(DB::Properties::kTotalSstFilesSize, &sstSize);
db->GetIntProperty("rocksdb.estimate-num-keys", &count);
double meanSize = sstSize / count; // calculate the mean size of an entry
std::array<Range, 1> ranges;
std::array<uint64_t, 1> sizes;
SizeApproximationOptions options;
options.include_memtabtles = true;
options.files_size_error_margin = 0.1;
ranges[0].start = start_key;
ranges[0].limit = end_key;
// get the approximate size of the range
Status s = GetApproximateSizes(options, column_family, ranges.data(), 1, sizes.data());
count = sizes[0] / meanSize; // calculate estimated number of entries
return count;
```

RocksDB Docs: https://github.com/facebook/rocksdb/wiki/Approximate-Size

PR #311, opened 2026-01-13
Initial attempt at getApproximateCount for #205

kriszyp:


kriszyp:
Superseded by #778, which takes a different approach to the bytes→count conversion: range-local SST entry density from GetPropertiesOfTablesInRange (no cached whole-CF mean or write-count invalidation), GetApproximateMemTableStats for a direct memtable count, guarded open-ended/inverted/zero-length bounds, and a progressive CountEstimator that refines the estimate while riding an iterator. Closing this one per Kris.

— Claude (Fable 5)

PR #380, opened 2026-02-05
Logs should be visible only after commit succeeds

kriszyp:
Currently logs were being made visible after a commit finishes, regardless of whether or not it actually committed the data. Specifically if the commit is "busy" (conflicted), the log position was being updated even though the corresponding data was not committed and visible/readable yet.
With this PR, the logs won't be made immediately visible if the commit is conflicted and does not actually complete. In order for the logs to be made visible and the position to be properly updated on the retry, I believe we need to preserve the existing transaction instance/handle to continue tracking the log position and a reset the transaction so it can be properly re-executed and the preserved log position can be used when the commit does eventually succeed for accurate/consistent reporting of log visibility. (I know we have previously discussed a preference for creating entirely new transaction wrappers to reset, but I am not sure how we would do that and continue to track the internal log position without exposing more complexity).

### DIFF
--- a/src/binding/transaction.cpp
+++ b/src/binding/transaction.cpp
@@ -237,10 +237,7 @@ napi_value Transaction::Commit(napi_env env, napi_callback_info info) {
 					descriptor->notify("committed", nullptr);
 				} else if (state->status.IsBusy()) {
 					// clear/delete the previous transaction and create a new transaction so that it can be retried
-					txnHandle->txn->ClearSnapshot();
-					delete txnHandle->txn;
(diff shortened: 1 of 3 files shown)


kriszyp:
I have had some weird flakiness with the benchmarks, but it is green right now :thinking: .

cb1kenobi:
> I have had some weird flakiness with the benchmarks, but it is green right now 🤔 .

Me too. I restarted the GitHub runner on my server, but I don't think it helped. Not sure what's up.

cb1kenobi (approved):
Looks great!

Issue #411, opened 2026-02-17
Essential benchmarks periodically crash in CI


cb1kenobi:
The benchmark suite causes Node.js to crash or timeout after 300s. When it crashes, sometimes we get a `corrupted size vs. prev_size in fastbins` and sometimes there is no message.

Crash no message: https://github.com/HarperFast/rocksdb-js/actions/runs/22039501401/job/63677950148?pr=400
Crash w/ message: https://github.com/HarperFast/rocksdb-js/actions/runs/22000176787/job/63570524698
Timeout: https://github.com/HarperFast/rocksdb-js/actions/runs/22087164757/job/63824278089

I have yet to reproduce it locally on a Mac. Benchmarks run on self-hosted runner, which runs in a Docker container on a Ubuntu 24.04.3 LTS machine.

It is unclear if this is crash is caused by `rocksdb-js` or `lmdb-js`.

PR #412, opened 2026-02-17
Synchronize benchmarks

cb1kenobi:
Vitest's benchmark uses `tinybench` and apparently the `teardown` callback can be async, but it won't block the next benchmark from running! I added a global `activeBenchmark` promise to synchronize the benchmarks so the setup of the next benchmark won't start until the previous teardown has completed.

Fixes https://github.com/HarperFast/rocksdb-js/issues/411

### DIFF
--- a/benchmark/setup.ts
+++ b/benchmark/setup.ts
@@ -15,6 +15,12 @@ const vitestBench = workerData?.benchmarkWorkerId
 		}
 	: (await import('vitest')).bench;
 
+/**
+ * A promise that blocks the next benchmark from starting until the current benchmark completes.
+ * `vitest bench` uses `tinybench` and for
+ */
+let activeBenchmark = Promise.resolve();
+
 export type LMDBDatabase = lmdb.RootDatabase<any, string> & { path: string };
 
 export interface BenchmarkContext<T> extends Record<string, any> {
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
I think I had fixed an issue with the bench not waiting for async `teardown` as well with https://github.com/HarperFast/rocksdb-js/commit/dee8a23d916b00a16bd87383e3d4d12fec7c068e#diff-48ac35a023ee06816a4df1d39061d6ca5773cb3ef1277c2947137ee7db5c48caR83 (changing the async `rm` to `rmSync`), so this makes sense.

cb1kenobi:
> I think I had fixed an issue with the bench not waiting for async `teardown` as well with [dee8a23#diff-48ac35a023ee06816a4df1d39061d6ca5773cb3ef1277c2947137ee7db5c48caR83](https://github.com/HarperFast/rocksdb-js/commit/dee8a23d916b00a16bd87383e3d4d12fec7c068e#diff-48ac35a023ee06816a4df1d39061d6ca5773cb3ef1277c2947137ee7db5c48caR83) (changing the async `rm` to `rmSync`), so this makes sense.

Yup, I don't get why teardown doesn't wait... or why it's not documented.

PR #416, opened 2026-02-18
Unref timeouts

cb1kenobi:
Unref timeout timers in benchmarks. This is more of a defensive improvement.

I audited all other setTimeouts and they should be fine without the unref.


kriszyp (approved):
Sorry I suggested this _after_ I approved 😅 . Looks good though.

PR #425, opened 2026-02-24
Fix memory leak and block cache size type

kriszyp:
This PR fixes a memory leak in reseting/deleting the log entry batch and a leak with removing transactions from the map of id->transactions. #429
It also fixes the type of the block cache size to be 64-bit; often we will use block cache that is bigger than 4GB. #430 
I also (think I) fixed some pointer ownership issues with maintaining a reference to the transaction log store, which started occurring after proper cleanup was taking place.

### DIFF
--- a/src/binding/transaction_log_handle.cpp
+++ b/src/binding/transaction_log_handle.cpp
@@ -39,7 +39,7 @@ void TransactionLogHandle::addEntry(
 		// store was closed/destroyed, try to get or create a new one
 		DEBUG_LOG("%p TransactionLogHandle::addEntry Store was destroyed, re-resolving \"%s\"\n", this, this->logName.c_str());
 		store = dbHandle->descriptor->resolveTransactionLogStore(this->logName);
-		this->store = store; // update shared_ptr to point to new store
+		this->store = store; // update weak_ptr to point to new store
 	}
 
 	// check if transaction is already bound to a different log store


src/binding/transaction_log_handle.h
```diff
@@ -17,7 +17,7 @@ struct TransactionLogHandle final : Closable {
 	/**
 	 * The transaction log store.
 	 */
-	std::weak_ptr<TransactionLogStore> store;
+	std::shared_ptr<TransactionLogStore> store;
```
cb1kenobi: This should be revert to a weak_ptr, no?

src/binding/transaction_log_handle.cpp
```diff
@@ -39,7 +39,7 @@ void TransactionLogHandle::addEntry(
 		// store was closed/destroyed, try to get or create a new one
 		DEBUG_LOG("%p TransactionLogHandle::addEntry Store was destroyed, re-resolving \"%s\"\n", this, this->logName.c_str());
 		store = dbHandle->descriptor->resolveTransactionLogStore(this->logName);
-		this->store = store; // update weak_ptr to point to new store
+		this->store = store; // update shared_ptr to point to new store
```
cb1kenobi:
```suggestion
		this->store = store; // update weak_ptr to point to new store
```

cb1kenobi (approved):

PR #470, opened 2026-03-17
`db.transaction()` retry on `ERR_BUSY`

cb1kenobi:
When using `db.transaction()` or `db.transactionSync()` and a transaction log, if the transaction fails due to a `ERR_BUSY` error, the log and the data in RocksDB will be inconsistent. The transaction should be retried.

Currently, `db.transaction()` will automatically abort the transaction if the commit fails leaving no opportunity for the transaction to be retried.

This PR adds a retry mechanism that will re-run the transaction. If the commit fails, it will retry up to `maxRetries` (default 3). If it still can't commit, or the `retryOnBusy` option is `false`, then it throws the `ERR_TRANSACTION_ABANDONED` error.

The transaction callback function now accepts a 2nd argument: `attempt`. This value starts at `1` and goes to `maxRetries`. Users will need to use `attempts === 1` to ensure `addEntry()` is only called once.

### DIFF
--- a/src/transaction-log-reader.ts
+++ b/src/transaction-log-reader.ts
@@ -53,6 +53,10 @@ Object.defineProperty(TransactionLog.prototype, 'query', {
 		if (start === undefined && !startFromLastFlushed) {
 			// if no start timestamp is specified, start from the last committed position
 			position = size;
+			if (position === 0) {
+				// if the file is empty, start after the header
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
This looks good to me, I think this is a solid way to handle this.
(I think I am still a little skeptical that throwing an error is the best way to communicate that something should be retried, but its fine, this works).
Good job getting this together.

PR #481, opened 2026-03-26
Fix `SIGBUS` crash when reading a recreated log file after purge

cb1kenobi:
## Problem
A `SIGBUS` (bus error / core dump) could crash the Node.js process when reading transaction log data after calling `purgeLogs()` on btrfs or ext4 filesystems.

**Root cause:** The transaction log store retains `currentSequenceNumber` across a purge — so the very next write recreates the log file at the same sequence number with only a tiny amount of content (a file header + one entry, ~36 bytes). When `getMemoryMap()` is called for the current log file it always maps `maxFileSize` (16 MB) rather than the actual file size, so the JavaScript reader held a 16 MB `Buffer` backed by a tiny file. On Linux (btrfs/ext4), accessing any virtual address whose backing page lies beyond the physical end of the file — i.e. byte 4096 or later when the file is only 36 bytes — causes the kernel to raise `SIGBUS`.

The crash was observed via a `callListenerCallback` stack trace: a committed event listener would call `log.query()`, which internally read from the memory-mapped buffer at an offset derived from the previous (pre-purge) `lastCommittedPosition`.

> _Note: `/tmp` (tmpfs) zero-fills pages beyond the file end for `MAP_SHARED` reads and so never triggers `SIGBUS` there; the crash only manifests on disk-backed filesystems._

## Fix
`transaction_log_file_posix.cpp` — two-step `mmap` strategy

Instead of a single `MAP_SHARED` mapping of `maxFileSize`:

1. Allocate a `MAP_PRIVATE | MAP_ANONYMOUS` region for the full `fileSize`. Every page defaults to zero — so reads beyond the actual file content safely return `0.0` timestamps and the reader stops gracefully.
2. Overlay the real file content at the start of that region using `MAP_SHARED | MAP_FIXED`. This gives readers live, up-to-date data while leaving the anonymous zero pages intact beyond the file's current end.

When the file grows after the initial mapping, the overlay is extended in-place via a second `MAP_FIXED` remap — no teardown needed.

A `mmapOverlaySize` field (added to `TransactionLogFile` in `transaction_log_file.h`) tracks how far the file-backed overlay currently extends.

## Also fixed in this branch (1b23e9df)
- TSFN / `napi_ref` resource leaks — `removeListenersByOwner()` and `DBDescriptor::close()` now explicitly call `napi_release_threadsafe_function` and `napi_delete_reference` before erasing listener entries, preventing active TSFNs from keeping the event loop alive after a database handle is closed.
- `callListenerCallback` memory leaks — `ListenerData*` is now freed when the TSFN teardown path passes `env == nullptr`, and the `argv` array is managed via `std::unique_ptr` so it is always released on every exit path.

## Test
A new regression test ("should not SIGBUS when reading a recreated log file after purge") reproduces the exact crash scenario deterministically:

1. Writes 12 transactions with 400-byte entries, pushing `positionInLogFile` to ~4969 bytes (past the 4 KiB page boundary).
2. Purges all log files.
3. Writes one small entry, recreating the log at the same sequence number with only ~36 bytes on disk.
4. Fetches the 16 MB memory map for that sequence number and reads byte 4096 — `SIGBUS` without the fix, 0 with it.

However, when the test is run via `vitest`, the `SIGBUS` crash is not surfaced, but the error can be easily reproduced by copying the following code into a script:

```javascript
import { RocksDatabase } from './dist/index.mjs';

const db = RocksDatabase.open('/var/tmp/rocksdb-sigbus-test');

const log = db.useLog('foo');

// Build a log file with `positionInLogFile` well past the 4 KiB page
// boundary (4096 bytes). Each entry costs:
//   TRANSACTION_LOG_ENTRY_HEADER_SIZE + payloadSize bytes
// The file itself starts with TRANSACTION_LOG_FILE_HEADER_SIZE bytes.
// With a 400-byte payload: 12 entries →
//   13 + 12 * (13 + 400) = 4969 bytes — past the page boundary.
const payload = Buffer.alloc(400, 0x55);
for (let i = 0; i < 12; i++) {
	await db.transaction(async (txn) => {
		log.addEntry(payload, txn.id);
		await txn.put(`key${i}`, payload);
	});
}

// Capture the committed position and verify the offset is beyond 4 KiB.
const positionBuf = log._getLastCommittedPosition();
const positionInLogFile = positionBuf.readUInt32LE(0);
const sequenceNumber = positionBuf.readUInt32LE(4);
console.log({
	positionInLogFile,
	expect: 4096,
});

// Purge every log file. The store retains `currentSequenceNumber` so
// the very next write recreates the file at the same sequence number.
db.purgeLogs({ before: Date.now() + 1000 });

// One small commit recreates the log at `sequenceNumber` with only
// TRANSACTION_LOG_FILE_HEADER_SIZE + TRANSACTION_LOG_ENTRY_HEADER_SIZE
// + 10 ≈ 36 bytes — far less than one 4 KiB memory page.
await db.transaction(async (txn) => {
	log.addEntry(Buffer.alloc(10), txn.id);
	await txn.put('small', Buffer.alloc(10));
});

// Fetch the memory map for the recreated (tiny) log file.
//
// Without the fix, POSIX maps maxFileSize (16 MB) via MAP_SHARED over
// a ~36-byte file. On btrfs/ext4, accessing any byte at or beyond the
// last 4 KiB page of the actual file (i.e. byte ≥ 4096) triggers
// SIGBUS because there is no kernel backing for those pages.
//
// With the fix, an anonymous zero-filled mapping covers the full
// maxFileSize region and only the real file content is overlaid at the
// start via MAP_FIXED, so reads past the file return 0 harmlessly.
const mmap = log._getMemoryMapOfFile(sequenceNumber);
console.log(`mmap is ${mmap ? 'defined' : 'undefined'}`);

// This single-byte read at the start of the second 4 KiB page is the
// exact access that fires SIGBUS without the fix on btrfs/ext4.
const byteAtPageBoundary = mmap.readUInt8(4096);
console.log({
	byteAtPageBoundary,
	expect: 0,
});

console.log('IT WORKS!');
```

### DIFF
--- a/src/binding/transaction_log_file_posix.cpp
+++ b/src/binding/transaction_log_file_posix.cpp
@@ -115,62 +115,20 @@ std::shared_ptr<MemoryMap> TransactionLogFile::getMemoryMap(uint32_t fileSize) {
 		return nullptr;
 	}
 
-	uint32_t actualFileSize = this->size.load(std::memory_order_relaxed);
-
-	if (this->memoryMap && this->memoryMap->mapSize >= fileSize) {
-		// The existing anonymous mapping is large enough. If the file has grown
-		// since the last time we overlaid it, extend the MAP_SHARED overlay so
-		// readers immediately see the new data without requiring a new mmap.
-		if (this->fd >= 0 && actualFileSize > this->mmapOverlaySize) {
-			uint32_t newOverlaySize = std::min(actualFileSize, fileSize);
-			void* updated = ::mmap(this->memoryMap->map, newOverlaySize, PROT_READ,
-				MAP_SHARED | MAP_FIXED, this->fd, 0);
-			if (updated != MAP_FAILED) {
-				this->mmapOverlaySize = newOverlaySize;
-				DEBUG_LOG("%p TransactionLogFile::getMemoryMap Extended file overlay to %u bytes\n",
-					this, newOverlaySize);
-			} else {
-				DEBUG_LOG("%p TransactionLogFile::getMemoryMap WARNING: overlay extension failed: %s\n",
-					this, ::strerror(errno));
-			}
-		}
-		this->memoryMap->fileSize = fileSize;
-		return this->memoryMap;
-	}
-
-	// Allocate an anonymous read-only region of the full requested size.
-	// All bytes default to zero, so readers that iterate past the end of
-	// actual file content will see 0.0 timestamps and stop gracefully
-	// instead of triggering a SIGBUS (which MAP_SHARED alone would cause
-	// when the file is smaller than fileSize on Linux).
-	void* anonMap = ::mmap(NULL, fileSize, PROT_READ, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-	if (anonMap == MAP_FAILED) {
-		DEBUG_LOG("%p TransactionLogFile::getMemoryMap ERROR: anonymous mmap failed: %s\n",
-			this, ::strerror(errno));
-		return nullptr;
-	}
-	DEBUG_LOG("%p TransactionLogFile::getMemoryMap Created anonymous map %p (size=%u, actualFileSize=%u)\n",
-		this, anonMap, fileSize, actualFileSize);
-
-	// Overlay the real file content at the start of the anonymous mapping so
-	// that readers see up-to-date data. MAP_FIXED replaces the corresponding
-	// pages of the anonymous region with file-backed shared pages.
-	uint32_t overlaySize = 0;
-	if (this->fd >= 0 && actualFileSize > 0) {
-		overlaySize = std::min(actualFileSize, fileSize);
-		void* fileMap = ::mmap(anonMap, overlaySize, PROT_READ, MAP_SHARED | MAP_FIXED, this->fd, 0);
-		if (fileMap == MAP_FAILED) {
-			DEBUG_LOG("%p TransactionLogFile::getMemoryMap ERROR: file overlay mmap failed: %s\n",
-				this, ::strerror(errno));
-			::munmap(anonMap, fileSize);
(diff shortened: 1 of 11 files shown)


src/binding/transaction_log_file_posix.cpp
```diff
@@ -115,20 +115,62 @@ std::shared_ptr<MemoryMap> TransactionLogFile::getMemoryMap(uint32_t fileSize) {
...
-		DEBUG_LOG("%p TransactionLogFile::getMemoryMap new memory map: %p\n", this, map);
-		if (map == MAP_FAILED) {
-			DEBUG_LOG("%p TransactionLogFile::getMemoryMap ERROR: mmap failed: %s", this, ::strerror(errno));
+	uint32_t actualFileSize = this->size.load(std::memory_order_relaxed);
+
+	if (this->memoryMap && this->memoryMap->mapSize >= fileSize) {
+		// The existing anonymous mapping is large enough. If the file has grown
+		// since the last time we overlaid it, extend the MAP_SHARED overlay so
+		// readers immediately see the new data without requiring a new mmap.
+		if (this->fd >= 0 && actualFileSize > this->mmapOverlaySize) {
```
kriszyp: I am a little confused as to how this works. We have to re-overlay the memory map if the file grows in size? But we don't call `getMemoryMap` every time we read from the transaction log after writes/growth, we are constantly reading new bytes past the position of the size of file when the memory map was created. I am probably missing something though.
And is this necessary if we have addressed the root cause of maintaining monotonicity of absolute position and bumping to a new sequence number on purge of the current log file?

cb1kenobi:
Abandoned in favor of https://github.com/HarperFast/rocksdb-js/pull/486.

PR #485, opened 2026-03-29
Use a mmap overlay to avoid SIGBUS, increment log file sequence numbe…

cb1kenobi:
Use a mmap overlay to avoid SIGBUS, increment log file sequence number if current log file is purged.

This PR supersedes https://github.com/HarperFast/rocksdb-js/pull/481.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ build
 deps
 dist
 node_modules
+.clangd
 .claude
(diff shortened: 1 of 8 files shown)


cb1kenobi:
Abandoned in favor of https://github.com/HarperFast/rocksdb-js/pull/486.

PR #486, opened 2026-03-29
Fix memory map SIGBUS error and race conditions

cb1kenobi:
This started out as a fix for a SIGBUS error that occurred when `purgeLogs()` was called followed by a transaction being committed (to add a log entry), and a transaction log query is evaluated. On mac and linux, memory maps are not filled with zeros, so reading beyond the memory map size is what triggers the SIGBUS.

1. Use a second memory map as an overlay to handle reading bytes beyond the allocated memory map page size
2. Advance the current log file sequence number when the current transaction log file is purged to ensure the transaction log sequences + position are always monotonic
3. Prevent a log store from purging a log file if it has any uncommitted transactions associated with that log file

These fixes exposed a pretty involved race condition on Windows that somehow has yet to occur on mac/linux. When a log store is purging, it's possible to close the log store while there are in-flight async transactions. This leads to `writeBatch()` executing when the log store is in a closed state. This didn't inherently throw any errors, but it does lead to orphan transaction log files (close wipes the sequence files map). On Windows, `purgeLogs()` deletes all known log files, then deletes the log store directory. However if the log store directory has an orphan file, then Node.js crashes deleting the non-empty directory.

To fix the race condition:
1. We need `DBDescriptor::purgeTransactionLogs()` to not force a store to close if it has in-flight associated transactions
2.  If `writeBatch()` is called once a store is closed, throw an error
3. A new atomic counter is used to keep track of the number of active associated transactions. We use an atomic counter instead of relying on shared_ptr ref counts which would interfere with log store lifecycles.
4. A new mutex to guard the incrementing of the transaction count and is store closing check

This PR supersedes https://github.com/HarperFast/rocksdb-js/pull/481 and https://github.com/HarperFast/rocksdb-js/pull/485.


kriszyp (commented):
This looks good, but have a few questions as I try to understand this:

> Use a second memory map as an overlay to handle reading bytes beyond the allocated memory map page size

Does this incur extra memory consumption? Does that actually allocate 16MB of zero-ed memory (presumably 16MB of physical RAM)? Is the overlay still necessary with the other fixes? I thought the SIGBUS was only occurring due to an actual bug in where the position in the log file pointing to a bad location?

> Advance the current log file sequence number when the current transaction log file is purged to ensure the transaction log sequences

Is it not sufficient acquire the write mutex and increment the sequence number if it is current log file? If there pending transactions, wouldn't they use the new sequence number after acquiring the write mutex?

cb1kenobi:
> This looks good, but have a few questions as I try to understand this:
>> Use a second memory map as an overlay to handle reading bytes beyond the allocated memory map page size
>
> Does this incur extra memory consumption? Does that actually allocate 16MB of zero-ed memory (presumably 16MB of physical RAM)? Is the overlay still necessary with the other fixes? I thought the SIGBUS was only occurring due to an actual bug in where the position in the log file pointing to a bad location?

Not especially. The second memory map (fileMap) is a fixed memory map that shares the same base address as the anonymous map.

I thought about whether this could this be removed and I think technically it could, but why? The SIGBUS is a legit issue.

The overhead of this solution is small: `maxFileSize` + one page table entry per 4k page. If we grew the `maxFileSize` to something big like 128MB, the small overhead would grow too, but would still be small relatively speaking.

>>Advance the current log file sequence number when the current transaction log file is purged to ensure the transaction log sequences
> Is it not sufficient acquire the write mutex and increment the sequence number if it is current log file? If there pending transactions, wouldn't they use the new sequence number after acquiring the write mutex?

I thought of this. My initial thinking was if you delete the log file, then just write the transaction to the next log file, however, order matters. Out of order writes leads to data loss. Deleting a log file that has a pending transaction is bad.

kriszyp:
> The overhead of this solution is small: maxFileSize + one page table entry per 4k page.

So if we have a 20 node network, with 20 different log stores per database, and 5 different databases, we are looking at 5x20x16MB, so over a GB of memory? And we run _many_ instances on each host (sometimes over a hundred).  That doesn't seem small. Memory is _extremely_ valuable to Harper, every KB counts.

Also, having a transaction log position that is out of sync is catastrophic, and could lead to data loss or corruption. A SIGBUS fault seems like the appropriate error response when we read beyond the expected size.

> Deleting a log file that has a pending transaction is bad.

Yes definitely. But, I think it may be important to clarify what we mean by "pending transaction"
* An open transaction that has not yet been `commit()`'ed: This transaction hasn't committed itself to any transaction log file or sequence number yet, and will only do so with a write mutex, so I believe that this isn't really a concern for purging logs as long as we purge logs (or increment sequence number) with a write mutex lock, right?
* A transaction that is or has written to the transaction log, but hasn't yet committed to the RocksDB database: We definitely do not want to purge a log file with data written in this state. However, I actually don't think this really matters much because it is a subset of the next state:
* A transaction that is/has been written to the transaction log, but has not been flushed yet (specifically has not the RocksDB database has not been flushed and the txn.state update has not been written yet): Any transaction in this state really needs to be protected from purging logs, as deleting a transaction log with entries that have been committed, but not flushed still has potential for data loss. So I don't think it is so much the committed state we need to check, but the `lastWrittenFlushedPosition` that we need to verify.

kriszyp:
> The overhead of this solution is small: maxFileSize + one page table entry per 4k page.

Looking into this more, without `MAP_POPULATE`, I don't think the anonymous allocation actually uses any (physical) memory immediately, it is just virtual memory allocation (memory is zero'ed on access via page faults). But it would be good to verify that with tests (I don't see any performance regression in the benchmarks, which is good). Also, according to Claude, repeated calls to mmap are going to be triggering extensive TLB shootdowns which are particularly problematic for high core count situations (using `mremap` would be better, but still some cost). So I'm still a bit skeptical that this is necessary/desirable complexity to add (maybe should be a separate PR to evaluate?).

src/binding/transaction_log_store.cpp
```diff
@@ -356,22 +362,13 @@ void TransactionLogStore::doPurge(std::function<void(const std::filesystem::path
...
-				break;
-			}
-		}
-		if (!shouldPurge) {
+		// only purge files that are entirely before the last flushed position,
+		// guaranteeing all their transactions have been committed to RocksDB
+		auto lastFlushedPosition = this->getLastFlushedPosition();
+		if (sequenceNumber > lastFlushedPosition.logSequenceNumber ||
+			(sequenceNumber == this->lastCommittedPosition->logSequenceNumber &&
+				lastFlushedPosition.positionInLogFile == this->lastCommittedPosition->positionInLogFile)
```
kriszyp: It seems like it should be this:
```suggestion
			(sequenceNumber == lastFlushedPosition.logSequenceNumber &&
				logFile->size > lastFlushedPosition.positionInLogFile)
```

cb1kenobi: Doh, I had the `logSequenceNumber` check and then autosuggest popped up a bad recommendation. Good catch!

cb1kenobi:
@kriszyp we are green!

kriszyp (approved):
Let's ship it!
I would like to come back to this and reassess if we really need `pendingTransactionCount` and overlays, but let's release this so we can get 5.0.1 out.

PR #512, opened 2026-04-16
Fix dead lock when purging transaction logs

cb1kenobi:
Fixes two related issues. One of them was a deadlock that caused the transaction log worker test to sometimes timeout.

**Issue 1: `DBDescriptor::close()` deadlock**

When closing a descriptor, it would hold `transactionLogMutex` while calling `store->close()` which acquires `writeMutex`. This deadlock occurred when the main thread was closing the descriptor while another thread holds `writeMutex` and RocksDB flush callback is blocked waiting for `transactionLogMutex` to free.

To fix it, update RocksDB flush listeners to hold the `transactionLogMutex` only while we collect the stores to close, release the lock, then signal the database is flushed.

Also, when closing the DB descriptor, hold the `transactionLogMutex` only while we collect the stores to close, then release the lock, then close the stores.

**Issue 2: `resolveTransactionLogStore()` returns stores that are closing/closed**

`purgeTransactionLogs()` calls `tryClose()`, `isClosing` is set to true, but the store is still in the stores map. `resolveTransactionLogStore()` sees the store still in the map and doesn't care if it's closing/closed, so it happily returns it.

The fix is to simply check if the store is closing, and if so, return a new store.

Disclaimer, Claude burned a lot of tokens to find and fix the deadlock.

### DIFF
--- a/src/binding/db_descriptor.cpp
+++ b/src/binding/db_descriptor.cpp
@@ -48,8 +48,16 @@ public:
 			DEBUG_LOG("%p TransactionLogEventListener::OnFlushBegin flushedSequence=%llu\n",
 				desc.get(), (unsigned long long)flushedSequence);
 
-			std::lock_guard<std::mutex> lock(desc->transactionLogMutex);
-			for (auto& [name, store] : desc->transactionLogStores) {
+			// Collect stores while holding the lock, then call methods without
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Great!

PR #514, opened 2026-04-18
Add stress test for opening many database instances

### LINKED ISSUE #473: Stress test hundreds of database instances
Harper opens hundreds of database instances where each one can experience thousands of transactions. We don't really have a test that stresses multiple database instances, including different column families, then throw in the occasional close.

cb1kenobi:
Harper opens a lot of database instances and column families, so this stress test will test this scenario.

`stress-test` and `benchmark` were not correctly added to the `tsconfig` and thus types weren't being enforced. After adding these, I needed to add types to the benchmark `setup.ts`.

Fixes #473

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -37,6 +37,11 @@ jobs:
         with:
           gh-token: ${{ secrets.GH_TOKEN }}
 
+      - name: Build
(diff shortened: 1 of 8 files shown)


kriszyp (approved):
Looks good!

PR #520, opened 2026-04-22
Read-only mode

### LINKED ISSUE #490: Add support for opening database in read-only mode
This could be useful in situations where we want to access a RocksDB from another process.

cb1kenobi:
Adds support for opening a RocksDB database in read-only mode. In read-only mode, all write operations are forbidden. Read operations, including those in transactions, as well as reading the transaction log, are allowed.

You can open same database multiple times in both read/write and read-only mode in the same process. Separate processes can only open a database in read-only mode.

The `DBRegistry` tracks `DBDescriptor`s by distinct path + readonly mode. Transaction logs are tied to a database path, thus there's a new `TransactionLogStoreRegistry` that orchestrates the transaction log stores so that they can be shared across `DBDescriptors`.

Fixed a number of bugs in `DBRegistry`. `DBRegistry::OpenDB()` adds the `DBRegistryEntry` to the `databases` map before `DBDescriptor::open()` is called and if it throws a C++ exception, then we have a stale entry in the registry that needs to be cleaned up. There was a null dereference in the `RegistryStatus()` function. Lastly, it was possible that the database entry's condition (used for coordinating access to the RocksDB database path) to be used after the entry is freed.

I updated the readme to document optimistic vs pessimistic. Also added a missing transaction mode unit test.

Fixes #490


src/binding/transaction.cpp
```diff
@@ -561,6 +563,11 @@ napi_value Transaction::PutSync(napi_env env, napi_callback_info info) {
 	NAPI_GET_BUFFER(argv[1], value, nullptr);
 	UNWRAP_TRANSACTION_HANDLE("Put");
 
+	if ((*txnHandle)->dbHandle->descriptor->readOnly) {
```
kriszyp: Does the RocksDB `Put()` method not return a status indicating that you can't write a read-only database? I am curious why we need to reimplement a check for read-only if Rocks does this for us? (or does it do in an inconvenient way?)

src/binding/db_registry.h
```diff
@@ -11,6 +11,28 @@
 
 namespace rocksdb_js {
 
+/**
+ * Lightweight key for the registry map. Composed of the three fields that
+ * uniquely identify a database instance: path, transaction mode, and whether
```
kriszyp: Might update this comment since transaction mode isn't in there.

src/binding/transaction.cpp
cb1kenobi: Yes, `Put()` does return a non-ok status in readonly mode:
```
Put failed: Not implemented: Not supported operation in read only mode.
```
I figured a quick check with a sensible error message made more sense.

src/binding/db_registry.h
cb1kenobi: Fixed

src/binding/database.cpp
```diff
@@ -1166,6 +1177,7 @@ napi_value Database::UseLog(napi_env env, napi_callback_info info) {
 	NAPI_METHOD_ARGV(1);
 	NAPI_GET_STRING(argv[0], name, "Name is required");
 	UNWRAP_DB_HANDLE_AND_OPEN();
+	THROW_IF_READONLY();
```
kriszyp: Why can't we access logs? Being able to read transaction logs in read-only mode is important, I believe. And this is the first/primary error that occurs when you try to open `harper` is read-only mode.

src/binding/transaction.cpp
kriszyp: This seems like a reasonable, descriptive, and appropriate error message to me. It seems like using RocksDB's error messages gives us better consistency of error messaging. It is easy to search for, or for AI to understand that, the error message directly maps to a RocksDB error. Remapping errors to different error messages can kinda obscure that connection. And there other RocksDB errors that we directly preserve. And it seems like leaning on RocksDB's errors keeps the code simpler, I would guess.

kriszyp (commented):
This looking good. The `useLog` error is the biggest blocker if you try to run Harper with this right now.

src/binding/database.cpp
cb1kenobi: Fixed!

kriszyp (approved):
Looks great, nice work, thank you!

kriszyp (approved):
Great!

Issue #446, opened 2026-03-04
Optimize `getRange()`/iteration


kriszyp:
We would like to optimize the `getRange` and iteration to avoid interacting with JS properties in the native C++ side (expensive), and to utilize the shared buffer mechanism where possible.

PR #524, opened 2026-04-25
Optimize getRange() iterator with shared buffers and bitmap flags

kriszyp:
From the `getRange() > small range (100 records, 50 range)` benchmarks:
main: 3.52K ops/sec
PR: 16.96K ops/sec
(4.8x improvement)

Mirrors the existing getSync() fast-path design to eliminate per-iteration NAPI overhead. Constructor now accepts a flag bitmask and three key offsets into the shared default key buffer instead of a JS options object, and Next() returns a primitive signal while writing key/value bytes into the shared KEY_BUFFER/VALUE_BUFFER and lengths into a new shared ITERATOR_STATE buffer (Uint32Array). The JS wrapper builds the {done, value} result.

Microbenchmark (50k entries, msgpack values, full forward scan, 10 runs):
  before: 2.71 us/entry,  370k entries/sec
  after:  0.87 us/entry, 1.14M entries/sec  (~3.1x faster)

The slow path (oversized key/value or stable-buffer decoder) still allocates fresh per-iteration buffers and is selected via flag. getCount() continues to use the original initFromNapiObject path.

Addresses: #446

(39MM Opus tokens)


src/dbi-iterator.ts
```diff
@@ -1,56 +1,88 @@
...
+			}
 		}
 
 		return { done: false, value: value as DBIteratorValue<T> };
 	}
 
 	return(value?: any): IteratorResult<DBIteratorValue<T>, any> {
 		if (this.iterator.return) {
-			return this.iterator.return(value);
+			this.iterator.return();
```
cb1kenobi: nit: We could do:

```suggestion
		this.iterator.return?.();
```

src/dbi-iterator.ts
```diff
@@ -1,56 +1,88 @@
...
 		}
 		return { done: true, value };
 	}
 
 	throw(err: unknown): IteratorResult<DBIteratorValue<T>, any> {
 		if (this.iterator.throw) {
-			return this.iterator.throw(err);
+			this.iterator.throw(err);
```
cb1kenobi: nit: We could do:

```suggestion
		this.iterator.throw?.(err);
```

src/dbi-iterator.ts
```diff
@@ -1,56 +1,88 @@
...
 		}
```
cb1kenobi: Now that we throw instead of returning error, which is correct, we have a test that catches the error and console.logs it, but doesn't `expect()` the error to match: https://github.com/HarperFast/rocksdb-js/blob/7f821ac1577026136bf8dbdf6ad2710162c60e0c/test/ranges.test.ts#L163

Please replace the `console.log(error);` with:

```
expect(error).to.beInstanceOf(Error);
expect(error.message).to.be('test');
```

src/store.ts
```diff
@@ -458,40 +479,94 @@ export class Store {
...
+		// the database) so it can skip a napi_instanceof check. The store
+		// holds the canonical NativeDatabase reference, so identity comparison
+		// is sufficient.
+		if (context !== this.db) {
+			flags |= ITERATOR_CONTEXT_IS_TRANSACTION_FLAG;
+		}
+
+		// Only pass the advanced ReadOptions object on the rare path where any
+		// of the underlying RocksDB iterator options are actually overridden.
+		const advancedOptions =
```
cb1kenobi, line 552: We don't seem to have a single unit test that sets any of these advanced options. We should at the very least, add a test that sets one of these so we can test the native code paths.

cb1kenobi (approved):
Looks awesome!

Issue #254, opened 2025-11-30
Batch write transaction log entries


kriszyp:
Currently each commit involves a write of transaction entries to the transaction log(s). This is accomplished with a write mutex surrounding a `writev` call to the transaction log file. Based on my performance testing, I would estimate that the `writev` calls are taking about 5μs each. And since these are in a mutex, they must be performed sequentially (regardless of how many threads we are using), limiting our transaction log performance to roughly 200K writes / second. There are some applications with write requirements closer to a million writes / second, so eventually improving this performance would be beneficial.

One possible optimization is to move the `writev` call outside the write mutex, and only "stage" the writes in the mutex. We would need to eliminate the O_APPEND flag and do writes with exact offsets so that multiple `writev` calls could be done in parallel. However, I would guess this is simply going to move the contention into the kernel, and may not significantly improve performance (although possibly worth trying).

I believe it is more likely that higher performance could be achieved by batching parallel transaction writes. Here is rough outline of a proposed commit process with transaction log write batching:
1. On the start of the commit, queue all of the transaction entries to be written into a queue in TransactionLogStore (using a mutex for this queue). Record the ending position of these entries once they are written to the transaction log (computed from the size of entries and existing queued entries).
2. Perform the RocksDB commit.
3. Check to see if the transaction log has written this commit's queued entries. If so, commit is done.  
4. If not, acquire write mutex.
6. Again, check to see commit's queued entries have been written and if so, commit is done. Otherwise, construct an iovecs of all queued entries (from this commit and any other commits that have entries in the store queue), and then remove all the entries transferred to the iovecs. Then perform the `writev` call for all the queued entries.

PR #525, opened 2026-04-25
perf(txnlog): reduce writeBatch overhead with three targeted fixes

kriszyp:
Realistic write load with workers > write variable records with transaction log benchmark:
* `main`: 203.90 ops/sec
* This PR: 518.48 ops/sec

This was originally intended to implement #254, but Claude did some experimentation and profiling and concluded that `writev` calls did not have much overhead and there were more significant gains to be made elsewhere:

- Cache fileLastWriteTime in open() to avoid a stat() syscall on every commit for the maxAgeThreshold age check
- Stack-allocate iovec array (capacity 8) in writeEntriesV1 to eliminate a heap allocation per commit for single-entry batches
- Replace std::set<LogPosition> uncommittedTransactionPositions with a sorted std::vector for better cache locality and no per-node allocation; add positionInsert/positionErase helpers to preserve no-duplicate semantics

Also adds benchmark/perf-txnlog-driver.mjs, a tight microbenchmark driver for profiling the transaction log write path.

Together these move the hot path from ~143K to ~156K commits/sec (8 threads, disableWAL=true), with writeBatch's share of perf samples dropping from ~5% to ~1.7%.


cb1kenobi (approved):
What's up with `benchmark/perf-txnlog-driver.mjs`? A Claude remnant?

cb1kenobi (approved):

PR #528, opened 2026-04-28
Set database name on encoder

cb1kenobi:
I noticed [Harper setting the database name on the encoder](https://github.com/HarperFast/harper/commit/02abd63f9eee89403f35c6e0ae2afcd0e02faebc#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R116) and figured we might as well do it in rocksdb-js.


kriszyp (approved):
Smart.

PR #533, opened 2026-04-29
Fix `TransactionOptions` type and enable `retryOnBusy` to `saveStructures`

cb1kenobi:
Found a few inconsistencies with the `TransactionOptions` type.

Also, with https://github.com/HarperFast/rocksdb-js/pull/531 being closed, I thought it was a good idea to add the `retryOnBusy` flag to the `saveStructures` callback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,7 +399,7 @@ Synchronous version of `remove()`.
 
 ## Transactions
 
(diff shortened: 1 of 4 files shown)


kriszyp (approved):
Ah yes, that was indeed the important part of my old PR that should be saved, thank you!

PR #543, opened 2026-05-06
rocksdb-js CLI

cb1kenobi:
The `rocksdb-js` package can be installed globally and used to open existing databases, view stats, create, update, and delete data.

Limitations: it can't read data encoded by Harper's `RecordEncoder`.


kriszyp (approved):
This looks good to me. I think my suggestion is that it would be nice to be able to print entry values as UTF8 strings as an alternative/option to hex; often times there are strings in values that make them highly identifiable, even if there is some incomprehensible data mixed.

I'm guessing this hasn't had AI review yet (no claude PR, and I dunno if the DLC cross-model review step was followed), so here is one. I don't see anything to concerning here, but feel free to address anything you want:

**Overview:** Adds a 545-line interactive REPL CLI (`bin/rocksdb-js.mjs`) for inspecting and manipulating RocksDB databases. It supports column family management, range queries, transaction log inspection, and a JS sub-REPL. All 20 CI checks pass.

---

## Bugs

**1. `queryCommand` — O(N) individual reads defeats the range scan (performance bug)**

```js
for await (const { key } of currentDB.getRange({ start, end, values: false })) {
    const value = await currentDB.get(key);  // separate round-trip per key
```

`getRange` is called with `values: false` then each value is fetched individually. This is far slower than letting the iterator carry the value. Should be:
```js
for await (const { key, value } of currentDB.getRange({ start, end })) {
```

**2. `completer` returns the `COMMANDS` object when no command matches**

```js
return [hits.length ? hits : COMMANDS, line];
```

`readline`'s completer callback expects `[string[], string]`. Returning the `COMMANDS` object (a plain object, not an array of strings) when there are no matches will produce broken behavior. Should be `Object.keys(COMMANDS)`.

**3. `putCommand` — guard condition is always false**

```js
if (!key || value === undefined) {
```

`value` is `valueParts.join(' ')`, which is always a string (at minimum `''`). This guard never fires when the value is missing — should be `!value` or `value === ''`.

**4. `dropCommand` — confusing/contradictory behavior for the `default` column**

```js
if (column === 'default') {
    console.log('You cannot drop the default column family');
    const answer = await ask(`Are you sure you want to clear all data? (y/N) `);
    // ...
    const { time } = await run(() => dbs.default.drop());
```

It says "You cannot drop the default column family" then proceeds to call `.drop()` on it. If the intent is to prevent dropping `default`, the function should return early. If `.drop()` here means something safe for the default family, the messaging is wrong. Either way this is a logic/UX mismatch.

**5. `completer` — `currentDB` may be null during startup prompts**

```js
if (command === 'log') {
    const logs = currentDB.listLogs();  // currentDB is null before DB is opened
```

Tab completion is active from process start, including during the "database doesn't exist, create it?" prompt. If the user presses Tab at that point while typing `log`, this throws.

---

## Minor Issues

- **`purge-logs` is destructive but has no confirmation prompt** — unlike `clear` and `drop`, it runs immediately. Warrants at least a `Are you sure?` check.
- **`entry` in `logCommand` is used as a number but comes in as a string** — `entry < 1` works via JS coercion, but explicit `Number(entry)` would make intent clear.
- **Help text hardcodes spaces for alignment** — fragile to maintain as commands change; a data-driven formatter (pad to max command length) would be more robust.
- **`--version` flag**: the version is printed as the very first line before flag processing. Fine as-is but slightly non-standard (`--version` usually prints *only* the version). Not blocking.

---

## Positives

- Good use of `AbortController` for graceful Ctrl+C handling with re-prompt loop.
- `finally` block in `main` correctly closes all open column family handles and the readline interface.
- The `repl` command properly saves/restores readline history and exposes `db` as a context property.
- Detecting the "already open" lock error and offering the `--readonly` flag is excellent UX.
- Separation of command dispatch via `COMMANDS` map is clean.

---

## Summary

The `queryCommand` range-then-individual-get pattern and the `completer` returning a non-array are the two bugs I'd want fixed before merge. The `putCommand` guard and `dropCommand` messaging are also worth correcting. The rest are polish.

kriszyp (approved):
-Sorry that was meant to be an approval :).-, ugh it was already one :laughing:

PR #544, opened 2026-05-06
Compaction support

cb1kenobi:
I don't think compaction was ever being triggered, even on close. This PR adds compaction on close as well as an explicit `db.compact()` API that can be periodically invoked.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -53,6 +53,13 @@ jobs:
       matrix:
         node: [20, 22, 24]
         os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - node: 20
+            pnpm-version: '10'
+          - node: 22
+            pnpm-version: latest
+          - node: 24
+            pnpm-version: latest
 
     steps:
       - name: Checkout repository
@@ -61,7 +68,7 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
-          version: latest
+          version: ${{ matrix.pnpm-version }}
 
       - name: Use Node.js ${{ matrix.node }}
         uses: actions/setup-node@v6
@@ -100,6 +107,12 @@ jobs:
         with:
           version: latest
 
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: 'pnpm'
+
       - name: Install Bun
         uses: oven-sh/setup-bun@v2
         with:
@@ -135,6 +148,12 @@ jobs:
         uses: pnpm/action-setup@v4
         with:
           version: latest
+
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: 'pnpm'
 
       - name: Install Deno
         uses: denoland/setup-deno@v2
--- a/README.md
+++ b/README.md
@@ -193,6 +193,27 @@ const entriesRemoved = db.clearSync();
 console.log(entriesRemoved); // 10
 ```
 
+### `db.compactRange(options?): Promise<void>`
+
+Compacts a range of keys in the database. In RocksDB, deleted keys are not immediately removed from
+the database. Instead, they are marked as deleted and a tombstone is written. This function
+triggers a manual compaction which removes the tombstones and reclaims space.
+
+- `start?: Key` The start key of the range to compact.
+- `end?: Key` The end key of the range to compact.
+
+```typescript
+await db.compactRange();
+```
+
+### `db.compactRangeSync(options?): void`
+
+Synchronous version of `compactRange()`.
+
+```typescript
+db.compactRangeSync();
+```
+
 ### `db.destroy(): void`
 
 Completely removes a database based on the `db` instance's path including all data, column families,
@@ -222,6 +243,23 @@ Synchronous version of `db.drop()`.
 const db = RocksDatabase.open('path/to/db');
 db.dropSync();
 db.close();
+```
+
+### `db.flush(): Promise<void>`
+
+Flushes all in-memory data to disk asynchronously.
+
+```typescript
+await db.flush();
+```
+
+### `db.flushSync(): void`
+
+Flushes all in-memory data to disk synchronously. Note that this can be an expensive operation, so
+it is recommended to use `flush()` if you want to keep the event loop free.
+
+```typescript
+db.flushSync();
 ```
 
 ### `db.get(key: Key, options?: GetOptions): MaybePromise<any>`
@@ -875,23 +913,6 @@ await Promise.all([
 Note: If the `callback` throws an error, Node.js suppress the error. Node.js 18.3.0 introduced a
 `--force-node-api-uncaught-exceptions-policy` flag which will cause errors to emit the
 `'uncaughtException'` event. Future Node.js releases will enable this flag by default.
-
-### `db.flush(): Promise<void>`
-
-Flushes all in-memory data to disk asynchronously.
-
-```typescript
-await db.flush();
-```
-
-### `db.flushSync(): void`
-
-Flushes all in-memory data to disk synchronously. Note that this can be an expensive operation, so
-it is recommended to use `flush()` if you want to keep the event loop free.
-
-```typescript
-db.flushSync();
-```
 
 ## Transaction Log
 
--- a/README.md
+++ b/README.md
@@ -193,25 +193,30 @@ const entriesRemoved = db.clearSync();
 console.log(entriesRemoved); // 10
 ```
 
-### `db.compactRange(options?): Promise<void>`
+### `db.compact(options?): Promise<void>`
 
 Compacts a range of keys in the database. In RocksDB, deleted keys are not immediately removed from
 the database. Instead, they are marked as deleted and a tombstone is written. This function
 triggers a manual compaction which removes the tombstones and reclaims space.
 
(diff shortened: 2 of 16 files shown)


kriszyp (approved):
Just to be clear, this is to enable explicit manual compation, right? Standard automated background compaction should be running on a background thread based on a bunch heuristics, and I believe this is working? And I think I have observed it working when running benchmarks where old SSTs are clearly being deleted. (We would not have anywhere close to enough internal knowledge of RocksDB to run this manually compared to RocksDB's intrinsic internal knowledge)

Claude's analysis also agrees it is already running by default:

Compaction is on by default in RocksDB — no explicit enablement needed. Looking at your config in DBDescriptor::open() (line 677+):
• IncreaseParallelism(options.parallelismThreads) (line 684) — actively sets max_background_jobs to enable background compaction threads. This is the key call that ensures compaction has threads to run on.
• disable_auto_compactions is never set → defaults to false → auto compaction is active.
• enable_blob_garbage_collection = true (line 698) — blob GC runs as part of compaction, so this also relies on it running.
• No CompactionStyle is set → defaults to level-based, which auto-compacts aggressively.
Your settings actively enable and support compaction. It will run automatically in the background.

kriszyp:
AI Review (again, just in case you haven't already done it):

# PR #544: Compaction Support — Code Review

## Overview
Adds explicit `compact()`/`compactSync()` API and triggers `CompactRange` on close. The PR premise ("compaction was never triggered") is partially incorrect — RocksDB runs background compaction automatically via `IncreaseParallelism` and heuristics. What this PR actually adds is *forced full-range compaction*, which is a meaningful difference.

---

## Issues

**1. Double compaction on every close (correctness/performance)**

`DBDescriptor::close()` calls `CompactRange` on all column families, then resets `this->db`, which triggers `DBDeleter::operator()`, which calls `CompactRange` again on the default CF. Every close compacts twice.

**2. `DBDeleter` only compacts the default column family**

`db->CompactRange(..., nullptr, nullptr)` in `DBDeleter` only compacts the default CF. If the database has custom column families (which it does — `close()` iterates them), they're not compacted here. This is inconsistent with `close()`.

**3. Full compaction on every close is expensive by default**

`CompactRange` on close blocks and can take minutes on large databases. This is a very aggressive default. Users with large datasets will see dramatically increased close times with no way to opt out. Suggest making this opt-in (`CloseOptions { compact: bool }`) or removing it from `close()` entirely and relying on the explicit API.

**4. Memory leak in `Compact()` on NAPI error**

```cpp
auto state = new AsyncCompactState(env, *dbHandle);

napi_valuetype startType;
NAPI_STATUS_THROWS(::napi_typeof(env, argv[2], &startType));  // leaks state on failure
```
After `new AsyncCompactState`, any `NAPI_STATUS_THROWS` that triggers before `delete state` will leak. The `napi_create_reference` and `napi_create_async_work` calls both have this risk.

**5. Test assertions may be flaky**

`rocksdb.estimate-live-data-size` is an estimate. The "compact on close" test reopens the db and compares estimates — this may not reliably reflect tombstone removal, especially with small datasets. Consider using `rocksdb.total-sst-files-size` which is less approximate.

**6. `compactSync()` event loop warning is undersold**

The docs say "can be expensive" — but for large DBs it can block for minutes. Should say: "Blocks the Node.js event loop for the duration of compaction. Not suitable for production use on large databases."

---

## Minor

- `AsyncCompactState` has a trailing blank line in the header (`database.h:110`)
- The `compact()` TS method duplicates the encode-key-to-buffer logic for start/end — could extract to a small helper
- The key range test end `'bbc'` works but is fragile — a comment explaining why `'bbc'` is the exclusive end for `'bbb-*'` entries would help readers

---

## Verdict

The explicit `compact()`/`compactSync()` API is a clean, useful addition. The auto-compact-on-close behavior is the main concern — it changes close latency for all users silently. Recommend removing on-close compaction from the default path (or gating it behind an option), and fixing the `DBDeleter` double-compact.

cb1kenobi:
I did some testing. I inserted 40,000 records, table size is 5.4 GB. I delete all 40,000 records, table size still 5.4 GB. I stop Harper (forcing a compaction on close), start Harper back up, and size dropped all the way down to 1 MB. I did observe a 1-2 second delay when stopping Harper. For a much larger database, it may take a few seconds to shutdown.

kriszyp:
> I delete all 40,000 records

I don't think that's even close to enough to trigger a flush (less than a MB of tombstones), much less compaction. You should eventually get compaction at some point, but with default levels, that's a pretty small amount of pressure on the compaction triggers, I think. I doubt it would happen any time soon. Again, I am pretty sure I have witnessed compaction under more realistically substantive loads. Also, I would assume that's pretty much all blobs, which has a little different compaction/cleanup process than SSTs.

cb1kenobi:
40,000 records with 100k of data. That's over 5GB of data. If that doesn't trigger compaction, I don't know what does.

cb1kenobi:
BTW, I'm working on fixing the Deno tests and testing with Harper. Stay tuned.

kriszyp:
image

kriszyp:
I'm happy to have an manual mechanism for this. I approve and think this looks good.

However, I'd assume the testing you did suggested that it would be nice for the real automated compaction to more aggressively reclaim this space. Manual compaction is neither desirable, reliable, or feasible, and Harper certainly doesn't have adequate acknowledge of the internals of a database to try to trigger that. This isn't a real solution to that problem.

Here are some suggestions that Claude had for this use case to more aggressively reclaim space (it reiterated that the sizes in this test wouldn't even remotely come close to being big enough to trigger compaction, it is designed to trigger on vastly larger amounts of SSTs): 

# RocksDB Compaction Tuning for Blob Space Reclamation

Context: `min_blob_size = 2048`, bulk deletes of large records (5GB+), need automatic space reclamation without manual `CompactRange` calls.

## Recommended Changes to `db_descriptor.cpp` (`cfOptions`)

### 1. CompactOnDeletionCollectorFactory (most targeted)

Triggers compaction automatically when delete ratio in an SST file exceeds threshold.
Directly addresses bulk-delete case — tombstone flush → immediate compaction scheduled → blob GC runs.

```cpp
#include "rocksdb/utilities/table_properties_collectors.h"

cfOptions.table_properties_collector_factories.push_back(
    rocksdb::NewCompactOnDeletionCollectorFactory(
        4096,  // sliding_window_size: number of keys to observe
        256    // deletion_trigger: deletes in window needed to trigger compaction
    )
);
```

### 2. periodic_compaction_seconds (safety net)

Forces SST files to be re-compacted after a time window even without size pressure.
Blob GC only runs during SST compaction, so this ensures orphaned blob data is eventually collected
even for cases that slip past CompactOnDeletion.

```cpp
cfOptions.periodic_compaction_seconds = 3600; // 1 hour
```

### 3. Lower blob_garbage_collection_force_threshold

Default is 1.0 (blob file must be 100% dead to force GC). Lowering to 0.5 means blob files get
rewritten once half their references are dead, reclaiming space in fewer compaction rounds.

```cpp
cfOptions.blob_garbage_collection_force_threshold = 0.5;
```

## Why all three

- `CompactOnDeletion`: handles the immediate trigger after bulk deletes
- `periodic_compaction_seconds`: safety net for cases that slip through (e.g., spread-out deletes)
- Lower blob GC threshold: ensures the blob data is actually reclaimed in the first compaction
  that touches those SST files, rather than waiting for 100% dead blob files

IMO, these suggestions are actually probably way too aggressive. Again, 5GB/40K records is tiny in our business. But I could see this being something we chase more aggressively in free instances.

cb1kenobi:
@kriszyp I decided to NOT compact on close, but to rather make it a config option. I added a `compact` command to the CLI. I don't think Harper needs to manually invoke compaction, but I like know that we can and that it works.

cb1kenobi:
So, couple things came up in this PR:

1. pnpm@11 released and dropped Node 20. I've updated the PR workflow to use pnpm@10 for Node 20, otherwise pnpm@latest for Node 22/24.
2. ~Apparently in Deno calling `napi_get_value_int32()` with a `Buffer` returns `napi_ok` and a value of `0`. This is bad because the shared buffer mechanism uses a number to signal that `getSliceFromArg()` should use the shared buffer. Unfortunately, we need to check if the arg is a buffer first, then check if it's a number.~

kriszyp:
> Apparently in Deno calling napi_get_value_int32() with a Buffer returns napi_ok and a value of 0. This is bad because the shared buffer mechanism uses a number to signal that getSliceFromArg() should use the shared buffer. Unfortunately, we need to check if the arg is a buffer first, then check if it's a number.

Can we check if it is buffer if the length is zero? The benchmark didn't run after this commit, so unclear if this has a performance impact. Some buffer interactions are expensive, so not sure. Also, kinda seems like a separate PR from this, but if you need this for passing tests.

cb1kenobi:
> Can we check if it is buffer if the length is zero? 

A buffer with length of zero is still a valid buffer. Deno's `napi_get_value_int32()` should not return `napi_ok` of the value is a buffer. I didn't bother to go back and get metrics, but I can't fathom the `napi_is_buffer()` call has a ton of overhead. I added a comment to `getSliceFromArg()` noting the Deno bug. This function is only called from `db.get()` and `txn.get()`, which should also be using the shared buffer, right?

> Also, kinda seems like a separate PR from this

I reverted my fix for `getSliceFromArg()` and the compact functions in favor of `NAPI_GET_BUFFER()`.

kriszyp (approved):
Still approve, looks great!

PR #545, opened 2026-05-08
Make transaction log read paths lock-free under concurrency

### LINKED ISSUE #462: Explore using atomics to reduce mutex locks
Mutexes are expensive and where we are using them to synchronize access to a variable, we should see if that variable can be an atomic eliminating the need for the mutex. First place to investigate would be the `TransactionLogStore`.

harper-joseph:
Eliminates dataSetsMutex and weak_ptr::lock() contention from the steady-state read path, unlocking high-QPS workloads with many concurrent subscribers (e.g., CDC replication).

Read-path changes (TransactionLogStore + TransactionLogHandle):

- Per-handle file snapshot cache (cachedFiles + filesVersion). Handles refresh the snapshot only when the store's filesVersion atomic advances (rotation, registration, purge); steady-state reads walk the local snapshot lock-free.
- Lock-free findPosition: walks cachedFiles newest-to-oldest, skipping files by their stored timestamp. Falls back to the store-level slow path only when the snapshot is empty.
- Lock-free getLogFileSize fast path: reads logFile->size atomic via the cached snapshot. Falls through to the slow path only when sequenceNumber=0 or the file is in the snapshot but not yet opened.
- Cached shared_ptr<TransactionLogStore> on the handle replaces the per-call weak_ptr::lock() CAS. Each read-method now does an isClosing.load() check instead. addEntry re-resolves to a fresh store and clears cachedFiles when isClosing is observed.
- currentSequenceNumber is now atomic (was plain uint32_t) so handle fast-path readers can compare without acquiring dataSetsMutex.

Per-file index changes (TransactionLogFile):

- Lock-free in-file timestamp index using a stable buffer + packed atomic state (low 32 bits = entry count, high 32 bits = position indexed up to). Replaces the std::map + indexMutex serialization. Acquire/release ordering on indexState publishes new entries safely to lock-free readers.
- Slow-path index extension serializes only on indexExtendMutex (per-file), not the global dataSetsMutex.
- Removed eager ensureIndexUpToDate at registerLogFile — saves up to maxFileSize/13 × 16 bytes per recovered file at startup. The lazy slow path handles the first reader instead.
- Inlined extendIndexLocked into findPositionByTimestamp (only caller).

Test coverage:

Adds 9 new regression tests covering the cases the changes affected:
- per-handle cache invalidation across many rotations
- concurrent first-time readers on a freshly-opened (unindexed) log
- multiple handles on the same log staying consistent
- iterator resume across rotations
- crash-free behavior under concurrent reads + purgeLogs(destroy:true)
- queries after attempting to purge earlier files
- ...and more

Bench:

Adds benchmark/worker-transaction-log-read.bench.ts — 4 access patterns (bulk forward scan, bulk forward scan with concurrent writer, high-frequency short-range queries, cursor-advance tail scan with writer) at 8 workers, comparing rocksdb-js against lmdb with matched lazy-durability semantics (noSync) and Harper-realistic LMDB options (snapshot:false, numeric keys).

### DIFF
--- a/src/binding/transaction_log_handle.h
+++ b/src/binding/transaction_log_handle.h
@@ -3,6 +3,7 @@
 
 #include <memory>
 #include <string>
+#include <vector>
 #include "db_handle.h"
 #include "transaction_log_store.h"
 
@@ -16,8 +17,15 @@ struct TransactionLogHandle final : Closable {
 
 	/**
 	 * The transaction log store.
+	 *
+	 * Held as a strong reference rather than a weak_ptr so steady-state
+	 * read methods don't pay the atomic refcount of weak_ptr::lock() per
+	 * call — under high concurrent QPS, that CAS on a single shared
+	 * cache line is a real bottleneck. The handle keeps the store alive
+	 * for as long as the handle itself is alive; addEntry re-resolves
+	 * to a fresh store if this one has been marked closing.
 	 */
-	std::weak_ptr<TransactionLogStore> store;
+	std::shared_ptr<TransactionLogStore> store;
 
 	/**
 	 * The name of the transaction log store.
@@ -33,6 +41,23 @@ struct TransactionLogHandle final : Closable {
 	 * The transaction id.
 	 */
 	uint32_t transactionId;
+
+	/**
+	 * Per-handle snapshot of all log files in the store, sorted by
+	 * sequence number ascending. Refreshed only when the store's
+	 * `filesVersion` counter advances (rotation, registration, purge);
+	 * the steady state hits this cache for every query without ever
+	 * touching dataSetsMutex.
+	 *
+	 * The shared_ptr keeps cached files alive even if they're purged
+	 * from the store's sequenceFiles map — their mmap and on-disk path
+	 * remain valid (Unix unlinks are deferred until last open closes).
+	 *
+	 * Each handle has its own cache, so concurrent subscribers don't
+	 * contend on shared state.
+	 */
+	std::shared_ptr<const std::vector<std::shared_ptr<TransactionLogFile>>> cachedFiles;
+	uint64_t cachedFilesVersion = 0;
 
 	/**
 	 * Creates a new transaction log handle.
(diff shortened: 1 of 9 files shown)


src/binding/transaction_log_handle.h
```diff
@@ -16,8 +17,15 @@ struct TransactionLogHandle final : Closable {
...
+	 *
+	 * Held as a strong reference rather than a weak_ptr so steady-state
+	 * read methods don't pay the atomic refcount of weak_ptr::lock() per
+	 * call — under high concurrent QPS, that CAS on a single shared
+	 * cache line is a real bottleneck. The handle keeps the store alive
+	 * for as long as the handle itself is alive; addEntry re-resolves
+	 * to a fresh store if this one has been marked closing.
 	 */
-	std::weak_ptr<TransactionLogStore> store;
+	std::shared_ptr<TransactionLogStore> store;
```
cb1kenobi, line 28: I'm skeptical that this will work. You are now bounding the life of the `TransactionLogStore` to the life of the `TransactionLogHandle`. When you call `db.purgeLogs({ destroy: true })`, it won't be able to fully close a `TransactionLogStore` and almost certainly the files will fail to be cleaned up on Windows.

The current design allows the `TransactionLogStore` to be destroyed while `TransactionLogHandles` reference it. The `TransactionLogHandle` life is bound to V8's garbage collection and `TransactionLogStore` is not.

With that said, there were some recent race conditions that where fixed by adding `isClosing` checks. I suppose it's possible the initial reason for the `weak_ptr` no longer exists.

cb1kenobi:
Ignore the failed tests. They are failing due to pnpm 11 dropping Node 20 and Bun/Deno not supporting sqlite. I fixed it in https://github.com/HarperFast/rocksdb-js/pull/548 and so it'll fix your tests when it lands.

kriszyp (commented):
I am curious what is meant by "read path"? The most common "read path" in Harper for most application is long-lived iterators, used by replication. And the dominant iteration involves _zero_ native/C++ calls, it is entirely in JavaScript. And this is intentional because it basically guarantees that the dominant read path not only does _not_ use locks, but it _can not_ use locks. And that's why we get many millions of iterations per second. Iteration is faster than is even possible with native calls, much less mutexes, I believe.
So what read path are we are talking about? Is this more for random access queries on the transaction log (this is more common with MQTT applications)?

PR #547, opened 2026-05-08
Update `getKeysCount()` to return exact count

cb1kenobi:
`getKeysCount()` returns estimated count if no `start` or `end`, but if we need the exact count, then we need to set `start` to `null`, which is a hack. Instead, make `getKeysCount()` always return the exact count and add a new `getEstimatedKeyCount()` method.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -53,6 +53,13 @@ jobs:
       matrix:
         node: [20, 22, 24]
         os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - node: 20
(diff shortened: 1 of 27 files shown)


kriszyp (approved):
FWIW, I wonder if we really should preserve this behavior. I think we did this "optimization" before we had direct access to `rocksdb.estimate-num-keys` through the `getIntProperty`. Now that we do, it kind of seems more consistent/intuitive that getKeysCount() is always exact (instead of only exact if there is a key range), and  `rocksdb.estimate-num-keys` is the way to get the estimate (as it clearly communicates).

PR #557, opened 2026-05-12
Drop Node 20, add Node 26 to CI, pin action versions, notify Harper on new release

cb1kenobi:
This PR targets `main` and does a bunch of housekeeping to keep up with the actively supported Node.js releases. All `uses` actions now have updated/pinned versions.

After publishing a new `rocksdb-js` release, it signals the `harper` repo to create a PR.

This PR also unblocks the merging of several dependencies that dropped Node.js 20.


.github/workflows/pr.yml
```diff
@@ -142,21 +135,21 @@ jobs:
...
 
       - name: Install pnpm
-        uses: pnpm/action-setup@v4
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # 6.0.8
         with:
           version: latest
 
       - name: Use Node.js 24
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # 6.4.0
```
dawsontoth: If you put a `v` before all of these, like
```
        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
```
I think renovate will detect it better and manage it in the future, keeping the sha and the version in sync. It's possible it'll work without the v! But I'm used to the v. 😄

dawsontoth (approved):

.github/workflows/pr.yml
cb1kenobi: This repo uses Dependabot instead of Renovate, but fixed for the greater good!

PR #558, opened 2026-05-12
Ensure data integrity on shutdown

cb1kenobi:
When the rocksdb-js binding is unloaded, call the `DBRegistry::Shutdown()` (which calls `PurgeAll()`) instead of calling `PurgeAll()`. This allows `DBDescriptors` to gracefully close.

I added 2 new brutal tests:
1. Terminate the worker while it's writing data without the WAL, then immediately call `shutdown()`
2. Call `shutdown()` while a worker thread is writing data without the WAL

On the second test, there was some race issues with sync methods. If we try to close a `DBDescriptor` while a sync operation is in-flight, rocksdb-js crashes. So I added a lightweight atomic counter for sync methods only and `DBDescriptor::close()` will simply wait until the counter returns to zero before closing.


kriszyp (commented):
Did this have AI review before submission? It looks mostly good to me, but Claude did have suggestions:
1. Add ACQUIRE_OPERATIONS_LOCK() to CompactSync (and audit Destroy).
2. Rename the macro's internal locals away from _descriptor / _operationGuard.
3. Replace the delay(100) / delay(50) in tests with explicit progress signals from the worker.
4. Add a small test that asserts "Database is closing" is thrown after shutdown().

cb1kenobi:
Code has been AI reviewed and ready for peer review!

kriszyp (approved):
Looks great! Thank you!

PR #559, opened 2026-05-13
Fixed bug with transaction logs not binding to transactions

cb1kenobi:
This PR started out with addressing the confusing "Log already bound to a transaction" error message. Logs are not bound to transactions, but rather transactions are bound to log stores.

Upon fixing this, I noticed that adding an entry to a bound transaction log instances was throwing an error that the `transactionId` was not set. The README example is correct, but the code didn't pass the `transactionId` into the `TransactionLogHandle` instance. Furthermore, we had zero tests for bounded transaction log instances.

This PR now lets you do:
```typescript
await db.transaction(async (txn) => {
  const log = txn.useLog('foo');
  log.addEntry(Buffer.from('hello')); // no transaction id required
});
```

And if you try to add an entry to the wrong store, it errors with a message like `Transaction 123 is already bound to the log store "foo"`.

After all these fixes, I found a huge bug where the "already bound" error was being thrown because when the store is closing and addEntry re-opens it, the new store instance didn't match the previous store because it was comparing the memory address and not the store names. This simple incorrect comparison caused it to think the store is not the same and throw.

### DIFF
--- a/src/binding/transaction.cpp
+++ b/src/binding/transaction.cpp
@@ -635,7 +635,8 @@ napi_value Transaction::UseLog(napi_env env, napi_callback_info info) {
 	// check if transaction is already bound to a different log store
 	auto boundStore = (*txnHandle)->boundLogStore.lock();
 	if (boundStore && boundStore->name != name) {
-		::napi_throw_error(env, nullptr, "Log already bound to a transaction");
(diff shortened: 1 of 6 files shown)


kriszyp (approved):
Looks good!

PR #573, opened 2026-05-19
Handle partial writev returns in transaction-log writeBatchToFile

### LINKED ISSUE #572: Partial `writev` in `writeBatchToFile` causes audit-log framing corruption
## Summary

`TransactionLogFile::writeBatchToFile` in `src/binding/transaction_log_file_posix.cpp` does not handle short `writev` returns. The POSIX spec allows `writev` to return fewer bytes than requested (and Linux does so under `EINTR`, mid-write `ENOSPC`, NFS/FUSE backed volumes, etc.). The current loop advances by iovec *count* rather than bytes written, so on a short return it silently drops the tail of one iovec and skips the rest of the batch, while still reporting the correct total-bytes-written to the caller.

The result is a corrupt transaction log: the file's on-disk bytes are contiguous, but a truncated entry's header claims more data than was actually written. The next batch appends immediately after the partial bytes, and the JS reader (`src/transaction-log-reader.ts`) advances by the (now wrong) length field, landing inside the next entry. Subsequent reads see garbage; eventually `getFloat64` throws `RangeError: Offset is outside the bounds of the DataView`.

The Windows emulation in `transaction_log_file_windows.cpp:352-403` detects partial `WriteFile` returns and `break`s out of the loop, but the caller `writeEntriesV1` has the same bookkeeping bug regardless of backend — it pre-increments `batch.currentEntryIndex` for every queued iovec before the write, so `batch.isComplete()` reports completion even when bytes were dropped.

## Investigation

Full investigation in HarperFast/harper#610 (comment): https://github.com/HarperFast/harper/issues/610#issuecomment-4489382836

Salient file/line refs:
- `src/binding/transaction_log_file_posix.cpp:238-266` — buggy `writev` loop.
- `src/binding/transaction_log_file_windows.cpp:352-403` — Windows partial-write break.
- `src/binding/transaction_log_file.cpp:187-220` — `writeEntriesV1` pre-increments index without rollback.
- `src/binding/transaction_log_entry.h:71-82` — vestigial `currentEntryBytesWritten` / `currentEntryHeaderWritten` resume state (never read for resume; `currentEntryHeaderWritten` is never set true).

## Proposed fix

1. POSIX `writeBatchToFile`: track byte progress through the iovec array, advance into a partial iovec's remainder by adjusting `iov_base` and `iov_len`, retry. On `EINTR`, retry; on `written == 0`, bail to avoid infinite loop; on other negative returns, surface as `-1` to caller.
2. Windows `writeBatchToFile`: convert the partial-write `break` into a retry loop on the same iovec.
3. Layer 2 (caller): once `writeBatchToFile` is write-everything-or-fail, `writeEntriesV1`'s pre-increment is safe — `bytesWritten < 0` already throws `DBException`. Drop the unused `currentEntryBytesWritten` / `currentEntryHeaderWritten` fields.

## Tests

- Unit test: exercise `writev` partial-return path (pipe with small buffer, or similar) and confirm all bytes make it through.
- End-to-end: write many small audit-log-like entries and read back via `query()` to confirm framing integrity.

🤖 Filed by Claude (Opus 4.7)

kriszyp:
## Summary

Fix audit-log framing corruption caused by short `writev()` returns in `TransactionLogFile::writeBatchToFile`. The previous POSIX loop advanced through iovecs by *count* rather than by *bytes written*, so any short return silently dropped the tail of one iovec and skipped the rest of the batch — but still reported the partial byte count to the caller. The on-disk file looked contiguous, a truncated entry's header claimed more data than was actually written, and the next batch appended right after the partial bytes. The reader's length-prefixed advance landed inside the next entry; downstream this surfaced as `RangeError: Offset is outside the bounds of the DataView` during replication catch-up.

Both backends now retry-until-complete-or-fail:

- POSIX: extracted writev loop to `TransactionLogFile::writevAll` in `src/binding/writev_all.cpp`. Tracks byte progress through a mutable iovec copy, trims `iov_base` / `iov_len` on partial writes, retries on `EINTR`, bails on `written == 0`.
- Windows: WriteFile partial returns now retry on the same iovec instead of breaking out of the loop with an incorrect partial count. Any error now returns `-1` unconditionally (was: return partial count if any bytes had been written, which the caller treated as success).

With `writeBatchToFile` now write-everything-or-fail (negative on error, which `writeEntriesV1` already turns into a thrown `DBException`), the half-built partial-resume state on `TransactionLogEntryBatch` is dead code. Removed `currentEntryBytesWritten` (only ever read by debug logs) and `currentEntryHeaderWritten` (never set true anywhere).

## Purpose

- Closes HarperFast/rocksdb-js#572
- Investigation: HarperFast/harper#610 (https://github.com/HarperFast/harper/issues/610#issuecomment-4489382836)

## Where to focus review attention

1. **writev loop correctness** (`src/binding/writev_all.cpp`): handles `EINTR` (retry), no infinite loop on `written == 0` (returns `-1`), advances by *bytes* not *count*, correctly chunks at `MAX_IOVS = 1024`.
2. **Windows parallel fix** (`src/binding/transaction_log_file_windows.cpp:352-407`): inner `while (remaining > 0)` retry loop, `static_cast<DWORD>(remaining)` matches existing pattern, error path returns `-1` instead of partial (changed semantic).
3. **Caller safety**: `writeEntriesV1` pre-increments `batch.currentEntryIndex` before the write; this is now safe because a partial write surfaces as `-1` and throws `DBException`. The removed `currentEntryBytesWritten` / `currentEntryHeaderWritten` fields were never used for resume.
4. **Test coverage of the partial path**:
   - Standalone C++ unit test (`test/binding/writev_partial_test.cpp`, run via `scripts/test-writev-partial.mjs`) compiles against `writev_all.cpp` with `-DROCKSDB_JS_WRITEV_ALL_STANDALONE` so it doesn't pull in N-API/RocksDB. Forces partial `writev` by sending 1 MiB through a default-sized pipe and asserts byte-exact round-trip; also covers `> MAX_IOVS` iovecs into a regular file, zero iovcnt, and bad fd.
   - JS integration tests in `test/transaction-log.test.ts` exercise the high-iovcnt batching path end-to-end (`> MAX_IOVS` entries in one batch, plus varied-size entries) and assert byte-exact round-trip via `query()`.
   - JS test (`test/writev-partial.test.ts`) wraps the C++ test as a vitest case so it runs as part of `pnpm test`.
5. **ODR**: `writev_all.cpp` defines `TransactionLogFile::writevAll`. Production builds include the full `transaction_log_file.h`; the standalone test build uses a minimal forward declaration guarded by `ROCKSDB_JS_WRITEV_ALL_STANDALONE`. The two builds are separate binaries, so the same translation unit doesn't see both definitions.
6. **C++ style**: tabs, brace placement, naming all match surrounding code in `transaction_log_file_posix.cpp` / `_windows.cpp`.

## Out of scope (intentionally not fixed)

- POSIX uses `O_APPEND` (`src/binding/transaction_log_file_posix.cpp:93`). If a partial write happens and `writevAll` returns `-1` (so the in-memory `size` is not advanced), the next process restart will `fstat` and pick up the stale partial-bytes-and-then-next-batch as the file size. The in-memory readers don't see this because they use the cached `size`, but the post-crash recovery path will. This is a pre-existing concern (same on `main`) and orthogonal to the writev correctness fix; not addressed here.

## Test plan

- [x] `node scripts/test-writev-partial.mjs` — standalone C++ partial-write test passes
- [x] `pnpm build:binding` — clean build
- [x] `pnpm test` — full suite passes (440 passed, 1 skipped)
- [x] `pnpm lint` / `pnpm fmt:check` / `pnpm type-check` — clean
- [ ] CI verification on Linux/macOS/Windows runners

## Review notes

- Cross-model review via Gemini confirmed loop correctness (POSIX + Windows), caller safety with vestigial fields removed, ODR setup, test coverage. No issues raised; one informational note about the orthogonal `O_APPEND` recovery concern documented above as out-of-scope.

---

PR generated by Claude (Opus 4.7).


src/binding/transaction_log/transaction_log_entry.h
```diff
@@ -70,17 +71,6 @@ struct TransactionLogEntryBatch final {
...
-	/**
-	 * The number of bytes already written from the current entry's data
-	 * (excluding the transaction header).
-	 */
-	uint32_t currentEntryBytesWritten = 0;
-
-	/**
-	 * Whether the transaction header for the current entry has been written.
-	 */
-	bool currentEntryHeaderWritten = false;
```
cb1kenobi, line 82: Ha, this wasn't even used!?

src/binding/transaction_log/transaction_log_entry.h
```diff
@@ -70,17 +71,6 @@ struct TransactionLogEntryBatch final {
 	 */
 	uint32_t currentEntryIndex = 0;
 
-	/**
-	 * The number of bytes already written from the current entry's data
-	 * (excluding the transaction header).
-	 */
-	uint32_t currentEntryBytesWritten = 0;
```
cb1kenobi, line 77: This was purely informational and ok to remove.

src/binding/transaction_log_file_posix.cpp
```diff
@@ -236,33 +236,9 @@ bool TransactionLogFile::removeFile() {
...
-
-		totalWritten += written;
-		offset += toWrite;
-		remaining -= toWrite;
-	}
-
-	return totalWritten;
+	// Implementation of writevAll lives in writev_all.cpp so the same loop
+	// can be linked into a standalone unit-test executable.
+	return TransactionLogFile::writevAll(this->fd, iovecs, iovcnt);
```
cb1kenobi: I don't think we need this `writevAll()` function. We should keep the logic in this method.

src/binding/writev_all.cpp
```diff
@@ -0,0 +1,89 @@
...
+	if (iovcnt <= 0) {
+		return 0;
+	}
+
+	// writev has a limit on the number of iovecs (IOV_MAX, typically 1024 on macOS)
+	// and may return short on partial writes (EINTR, ENOSPC, NFS/FUSE, etc.).
+	// We track byte progress through the iovec array and advance into a partial
+	// iovec's remainder by adjusting iov_base/iov_len so a short writev does not
+	// silently drop the tail of an entry.
+	constexpr int MAX_IOVS = 1024;  // IOV_MAX on most systems
```
cb1kenobi: Can we `#include <limits.h>` and use `IOV_MAX`?

test/writev-partial.test.ts
```diff
@@ -0,0 +1,14 @@
+import { execFileSync } from 'node:child_process';
+import { join } from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+// Compiles and runs the standalone C++ unit test that exercises
+// TransactionLogFile::writevAll's partial-write retry path against a pipe.
+// Skipped on Windows; the Win32 writeBatchToFile uses WriteFile, not writev.
+describe.skipIf(process.platform === 'win32')('writevAll partial-write retry', () => {
+	it('writes all bytes through a pipe even when writev returns short counts', () => {
+		const script = join(process.cwd(), 'scripts', 'test-writev-partial.mjs');
```
cb1kenobi: If we need tests to test scenarios in the C++ code that we can't test using the JavaScript API, then we should adopt a real C++ unit test runner like `gtest`. This would actually be advantageous as we could test many other scenarios that are challenging to test via JavaScript. Clang can instrument the C++ code and allow us to capture code coverage reports. I would much rather get a proper solution in place than band-aid some script to invoke the compiler and run a program. I can add gtest and give us a structure for adding C++ tests.

package.json
```diff
@@ -49,6 +49,7 @@
     "rebuild": "node-gyp rebuild",
     "rebuild:debug": "node-gyp rebuild --coverage --debug --verbose",
     "test": "cross-env CI=1 node --expose-gc ./node_modules/vitest/vitest.mjs --reporter=verbose",
+    "test:writev-partial": "node scripts/test-writev-partial.mjs",
```
cb1kenobi: We probably shouldn't specifically create a script for this test.

test/binding/writev_partial_test.cpp
```diff
@@ -0,0 +1,181 @@
...
+		if (n <= 0) break;
+		args->sink->insert(args->sink->end(), buf, buf + n);
+	}
+	return nullptr;
+}
+
+// writevAll handles short writev returns when the pipe buffer fills. We send
+// 1 MiB across 256 iovecs through a pipe whose default capacity is well under
+// the total, so the kernel will return short counts.
+void testPartialPipeWrite() {
```
cb1kenobi: I don't think this test is actually testing a partial `writev`s, at least not on macOS. On macOS, blocking `writev` to a pipe with an active reader completes in a single syscall, so there's no short return. If it was non-blocking, then you'd have to have a retry loop that checks  a returned error as `EAGAIN`.

The only/best thing to do is use preprocessor macros to define `writev()`. In release builds, `writeBatchToFile()` would call `ROCKSDB_JS_WRITEV()` which is defined as `#define ROCKSDB_JS_WRITEV ::writev` and in test builds would be some mocked function that returns a partial write. It would have to be crafty as to only run the mocked version for certain tests, but the real one for other tests, so you'd need some global state and a way to change it (maybe env vars?).

cb1kenobi (approved):
This is insanely cool.

PR #575, opened 2026-05-20
Expose write buffer options and retune defaults for transactional load

kriszyp:
## Summary

- Exposes `writeBufferSize`, `maxWriteBufferNumber`, `dbWriteBufferSize`, and `maxWriteBufferSizeToMaintain` as DB open options through both the TS API and the native binding.
- Retunes defaults toward "flush fast, queue deep" for `OptimisticTransactionDB` workloads with `atomic_flush = true`.
- Drops `parallelismThreads ?? 1` in the TS layer so the native `max(1, hardware_concurrency() / 2)` default actually applies.

## Why

Under burst load, a Harper instance hit repeated `OptimisticTransactionDB` conflict-check failures:

> Transaction could not check for conflicts for operation at SequenceNumber X as the MemTable only contains changes newer than SequenceNumber Y. Increasing the value of the max_write_buffer_size_to_maintain option could reduce the frequency of this error

Root cause: `src/binding/db_descriptor.cpp` hardcoded `dbOptions.db_write_buffer_size = 32MB` while `atomic_flush = true` was enabled, so under load every CF flushed atomically once the global 32MB sum was reached, before enough write history had accumulated for snapshot-based conflict checking. `max_write_buffer_size_to_maintain` was never set explicitly. The TS layer also defaulted `parallelismThreads` to `1`, starving background flush/compaction on multi-core hosts.

## Defaults changed

| Knob | Before | After | Notes |
|---|---|---|---|
| `writeBufferSize` (per-CF) | RocksDB default 64MB | **16MB** | Smaller memtable → faster, more frequent flushes |
| `maxWriteBufferNumber` (per-CF) | RocksDB default 2 | **16** | Deep queue absorbs bursts without write-stalls |
| `dbWriteBufferSize` (global) | hardcoded 32MB | **0 (disabled)** | Per-CF settings drive flushing; removes the cap that was at the root of the conflict-check failures |
| `maxWriteBufferSizeToMaintain` | unset (struct 0; OptTxnDB auto-applied formula) | **-1 (explicit formula)** | Resolves to `maxWriteBufferNumber * writeBufferSize` = 256MB with the new defaults |
| `parallelismThreads` (TS default) | `?? 1` | not defaulted | Falls through to native `max(1, hardware_concurrency() / 2)` |

## Where to focus the review

- **`max_write_buffer_size_to_maintain = -1` semantics** — relies on the RocksDB header documentation that `-1` means "use the `max_write_buffer_number * write_buffer_size` formula". Worth double-checking against the runtime behavior of RocksDB 11.1.1 (the version vendored in `deps/rocksdb/`).
- **Memory ceiling change** — with the new defaults, worst-case memtable memory per DB is roughly `N * 256MB` where N is the number of column families. For a Harper install with 10–20 tables that's a 2.5–5GB ceiling, vs. the previous ~32MB cap. This only materializes under sustained write pressure where flush can't keep up; under normal load, `atomic_flush` keeps actual usage much lower. Users on memory-constrained hosts can now set `dbWriteBufferSize` explicitly to restore a cap — but we should be sure we're comfortable shipping this as the default.
- **Backward compatibility** — removing the 32MB global cap is a behavior change for existing users. Anyone relying on it for memory pressure relief will see baseline RAM grow.
- **API naming** — the four new options map 1:1 to RocksDB option names (jargon-heavy but unambiguous for advanced users).

## Test plan

- [x] `pnpm build` — native + TS bundle build cleanly
- [x] `pnpm check` — type-check, lint, format all pass
- [x] `pnpm test` — full suite (442 tests) passes including new `test/db-options.test.ts`
- [x] New tests verify the option plumbing and that a small `writeBufferSize` actually produces SST output after a flush
- [ ] Validate under real Harper workload that the conflict-check failures stop reproducing

## Notes

This PR was generated by an AI agent (Claude Opus 4.7) and reviewed by Gemini 3 Pro before opening.

### DIFF
--- a/src/binding/database/database.cpp
+++ b/src/binding/database/database.cpp
@@ -1168,6 +1168,10 @@ napi_value Database::Open(napi_env env, napi_callback_info info) {
 	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "noBlockCache", dbHandleOptions.noBlockCache));
 	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "readOnly", dbHandleOptions.readOnly));
 	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "parallelismThreads", dbHandleOptions.parallelismThreads));
+	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "writeBufferSize", dbHandleOptions.writeBufferSize));
+	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "maxWriteBufferNumber", dbHandleOptions.maxWriteBufferNumber));
+	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "dbWriteBufferSize", dbHandleOptions.dbWriteBufferSize));
+	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "maxWriteBufferSizeToMaintain", dbHandleOptions.maxWriteBufferSizeToMaintain));
 	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "transactionLogMaxAgeThreshold", dbHandleOptions.transactionLogMaxAgeThreshold));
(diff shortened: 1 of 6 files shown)


cb1kenobi (approved):
This is awesome, great work!

PR #579, opened 2026-05-21
Fix transaction log shutdown race with flush callbacks

cb1kenobi:
Fix transaction log shutdown race with flush callbacks

During `db.shutdown()`, `DBDescriptor::close()` calls `Unregister()` before `db.reset()`, so RocksDB can still invoke `OnFlushBegin/OnFlushCompleted` on background threads while stores are closing.

Hold a `shared_ptr` to the registry entry in `GetStores()` so `Unregister` cannot destroy the entry between releasing `entriesMutex` and locking `storesMutex`. Reject flush handler work on stores that have already set `isClosing`, matching `writeBatch()`.

This PR should have been apart of https://github.com/HarperFast/rocksdb-js/pull/559.

### DIFF
--- a/src/binding/transaction_log_store.cpp
+++ b/src/binding/transaction_log_store.cpp
@@ -664,7 +664,7 @@ void TransactionLogStore::databaseFlushBegin(rocksdb::SequenceNumber rocksSequen
  */
 void TransactionLogStore::databaseFlushed(rocksdb::SequenceNumber rocksSequenceNumber) {
 	if (this->isClosing.load(std::memory_order_relaxed)) {
-		throw rocksdb_js::DBException("Transaction log store is closed");
+		return;
 	}
 
 	LogPosition latestSequencePosition = { 0, 0 };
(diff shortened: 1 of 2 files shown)


src/binding/transaction_log_store.cpp
```diff
@@ -630,6 +630,10 @@ bool operator>(const LogPosition a, const LogPosition b) {
  * log file (until it expires), even after crash.
  */
 void TransactionLogStore::databaseFlushBegin(rocksdb::SequenceNumber rocksSequenceNumber) {
+	if (this->isClosing.load(std::memory_order_relaxed)) {
+		throw rocksdb_js::DBException("Transaction log store is closed");
```
kriszyp:
```suggestion
		return;
```
I don't see any try/catch in the caller, and `OnFlushBegin` is called by RocksDB internal event delegation (I don't think we can expect it to try/catch).

src/binding/transaction_log_store.cpp
```diff
@@ -659,6 +663,10 @@ void TransactionLogStore::databaseFlushBegin(rocksdb::SequenceNumber rocksSequen
  * after restart or crash
  */
 void TransactionLogStore::databaseFlushed(rocksdb::SequenceNumber rocksSequenceNumber) {
+	if (this->isClosing.load(std::memory_order_relaxed)) {
+		throw rocksdb_js::DBException("Transaction log store is closed");
```
kriszyp:
```suggestion
		return;
```

src/binding/transaction_log_store.cpp
cb1kenobi: Good catch!

kriszyp (approved):

PR #583, opened 2026-05-22
Native unit tests

cb1kenobi:
Introduces unit tests for the native C++ code using [GoogleTest](https://google.github.io/googletest/). This allows us to easily create new C++ unit tests and run them via `pnpm test:native`. We also get code coverage report (stdout and html) when we run `pnpm coverage:native`.

FWIW, RocksDB uses GoogleTest for its unit tests.

Note that we can only test code that does NOT rely on Node API. Mocking Node API is out-of-scope for this PR. We are still able to write tests for internal things (such as the transaction log internals) and pretty much every surface that wraps RocksDB.

A few files including `macros.h`, `util.cpp`, and `util.h` contained code that used Node API and code that didn't. These files were split up which triggered a reorg of the native code. On the plus side, the directory structure looks nicer.


kriszyp (approved):
Impressive, looks good!

PR #584, opened 2026-05-26
Expose RocksDB WriteBufferManager with cost-to-cache support

kriszyp:
Adds a process-wide `WriteBufferManager` singleton in `DBSettings`, lazy-created on first DB open and shared across every database in the process. Memtable memory is bounded by a single configurable limit rather than each DB managing its own budget. With `costToCache: true`, memtable charges show up as pinned entries in the block cache, giving a single observability metric for read cache + memtable footprint.

## Why

In Harper Fabric's shared-host deployment, each container opens many RocksDB databases (one per Harper user database/system DB). Memtable memory accumulates per-DB and per-column-family — for a customer like Canopy with 14 tables, the maintain window alone (`max_write_buffer_size_to_maintain` × CFs) can account for ~1.8 GB of resident anonymous memory. Today there's no process-level cap. WriteBufferManager fixes that.

## What this changes

New `RocksDatabase.config()` options:
| Option | Type | Default | Description |
|---|---|---|---|
| `writeBufferManagerSize` | number | `0` | Bytes the WBM caps across all DBs in the process. `0` disables. Runtime-mutable via `SetBufferSize`. |
| `writeBufferManagerCostToCache` | boolean | `false` | Charge memtable bytes against the shared block cache. Fixed at first WBM creation. |
| `writeBufferManagerAllowStall` | boolean | `false` | Stall writes once the cap is reached. Runtime-mutable via `SetAllowStall`. |

## Implementation notes

- `DBSettings::getInstance()` now uses C++11 magic statics — the previous `unique_ptr` lazy-init raced on concurrent first-call from libuv worker threads. Independent of WBM, this also fixes a pre-existing risk.
- WBM configuration fields are `std::atomic` so JS-thread `Config()` writes are safe against worker-thread reads in `getWriteBufferManager()`.
- WBM creation is mutex-protected with double-checked locking on the atomic size.
- `costToCache` is fixed at first creation (already-open DBs hold `shared_ptr` refs; the cache reservation manager is configured at construction). Subsequent attempts to change it emit a stderr warning.
- `allowStall` IS updateable on a live WBM via `SetAllowStall` — `Config()` propagates runtime changes.
- Setting `writeBufferManagerSize = 0` at runtime means "no new attachments," not teardown. RocksDB asserts `SetBufferSize(new_size > 0)`.

## What's verified by tests

The 11 tests cover the structural claims and document the limitations:

- ✓ Negative size rejected with `RangeError`
- ✓ Default-off path still works (no WBM attached, writes succeed)
- ✓ WBM attaches to opened DBs; multiple DBs share one WBM
- ✓ Runtime resize via `SetBufferSize` doesn't disrupt writes
- ✓ **Active memtable memory IS reclaimed immediately after flush** — `cur-size-active-mem-table` drops by 10x
- ✓ With `costToCache`, memtable charges show up in `block-cache-usage` (≥1 MB delta for 2 MB writes)
- ✓ WBM charges appear as pinned entries (the cache cannot LRU-evict them)
- ✓ The structural cap on memtable memory is `writeBufferManagerSize`, NOT `blockCacheSize`
- ✓ Cache usage grows with memtable bytes (visible to operators via a single property)
- ✓ Older WBM charges are released when newer memtables push them out of RocksDB's maintain window

## Honest caveats (covered in test comments)

- The "memory returns dynamically when load drops" claim is partial. Active memtable charges release on flush; maintain-window-pinned memtables release only when newer writes push them out. With OptimisticTransactionDB defaults, ~128 MB / CF stays pinned until churned.
- OS-level RSS reduction is glibc-dependent. RocksDB's accounting drops correctly; whether glibc returns memory to the kernel depends on arena fragmentation. Tests assert against RocksDB properties, not RSS.

## Cross-model review findings addressed

**Gemini** — flagged the singleton race in `getInstance()` (fixed: magic statics) and data races on the WBM config fields (fixed: `std::atomic` + mutex). Suggested per-DB opt-out which is left for a follow-up.

**Codex** — caught that `allowStall` changes after WBM creation were silently dropped (fixed: now propagates via `SetAllowStall`).

## Related

Complements #575 (per-CF write buffer options) — that PR exposes the per-CF knobs, this one adds the process-wide cap that makes them safe to tune up.

Generated by Claude. — *Claude Sonnet 4.7*

### DIFF
--- a/src/binding/database/db_settings.cpp
+++ b/src/binding/database/db_settings.cpp
@@ -1,5 +1,4 @@
 #include "database/db_settings.h"
-#include <cstdio>
 #include "napi/macros.h"
 #include "core/platform.h"
 #include "napi/helpers.h"
@@ -116,30 +115,11 @@ napi_value DBSettings::Config(napi_env env, napi_callback_info info) {
 	}
 
 	int64_t writeBufferManagerSize = 0;
-	status = rocksdb_js::getProperty(env, params, "writeBufferManagerSize", writeBufferManagerSize, true);
-	if (status == napi_ok) {
-		if (writeBufferManagerSize < 0) {
-			::napi_throw_range_error(env, nullptr, "Write buffer manager size must be a positive integer or 0 to disable");
-			return nullptr;
-		}
-		const size_t newSize = static_cast<size_t>(writeBufferManagerSize);
-
-		// Take the mutex to serialize with WBM construction and SetBufferSize.
-		// We update the atomic *inside* the lock so observers in
-		// getWriteBufferManager() see a consistent state (lock held =>
-		// no concurrent WBM creation in flight).
-		std::lock_guard<std::mutex> lock(settings.writeBufferManagerMutex);
-		settings.writeBufferManagerSize.store(newSize, std::memory_order_relaxed);
-
-		// If the manager was already created, adjust its buffer size in place
-		// (SetBufferSize is atomic). RocksDB asserts new_size > 0, so when
-		// "disabling" via size=0 we leave the existing manager alone —
-		// already-open DBs keep their reference, and subsequent opens skip
-		// the manager entirely (see getWriteBufferManager). Runtime size=0
-		// is a "no new attachments" signal, not a teardown.
-		if (newSize > 0 && settings.writeBufferManager) {
-			settings.writeBufferManager->SetBufferSize(newSize);
-		}
+	const bool wbmSizeProvided =
+		rocksdb_js::getProperty(env, params, "writeBufferManagerSize", writeBufferManagerSize, true) == napi_ok;
+	if (wbmSizeProvided && writeBufferManagerSize < 0) {
+		::napi_throw_range_error(env, nullptr, "Write buffer manager size must be a positive integer or 0 to disable");
+		return nullptr;
 	}
 
 	// `costToCache` is immutable after WBM creation — the cache reservation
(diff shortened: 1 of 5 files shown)


src/binding/database/db_settings.h
```diff
@@ -15,19 +17,48 @@ class DBSettings final {
...
+	 *
+	 * Uses C++11 magic-static initialization so concurrent first-callers
+	 * from libuv worker threads don't race during construction.
+	 */
 	static DBSettings& getInstance() {
-		if (!instance) {
-			instance = std::unique_ptr<DBSettings>(new DBSettings());
-		}
-		return *instance;
+		static DBSettings instance;
```
cb1kenobi, line 60: Cool!

src/binding/database/db_settings.cpp
```diff
@@ -72,6 +115,64 @@ napi_value DBSettings::Config(napi_env env, napi_callback_info info) {
...
+
+		// Propagate allowStall to the live manager if it exists — this is
+		// the RocksDB-supported runtime knob.
+		if (wbmAlreadyCreated && allowStallProvided) {
+			settings.writeBufferManager->SetAllowStall(newAllowStall);
+		}
+
+		// costToCache can't be retro-applied; warn if the user attempts it.
+		if (wbmAlreadyCreated && costToCacheProvided && newCostToCache != oldCostToCache) {
+			std::fprintf(stderr,
```
cb1kenobi: A library should not write to stdio. Instead, we can throw and the user can retry the `config()` call. In order to throw, we need to reorder this function as to do this check before mutating any state. I've got a PR that targets the branch to show what I'm talking about: https://github.com/HarperFast/rocksdb-js/pull/585.

cb1kenobi (approved):

PR #585, opened 2026-05-26
Don't write warnings to stderr

cb1kenobi:
Suggested change to https://github.com/HarperFast/rocksdb-js/pull/584

- Parses `writeBufferManagerSize`, `writeBufferManagerCostToCache`, `writeBufferManagerAllowStall` up front.
- Folds all WBM mutations (size + costToCache + allowStall + SetBufferSize + SetAllowStall) into one critical section.
- Checks the costToCache invariant first inside the lock and throws via `napi_throw_error` if violated — before `SetBufferSize`, `SetAllowStall`, or any atomic store runs. A rejected change leaves the live WBM and settings untouched.
- Drops the `#include <cstdio>` along with the `fprintf(stderr, …)` call.

### DIFF
--- a/src/binding/database/db_settings.cpp
+++ b/src/binding/database/db_settings.cpp
@@ -1,5 +1,4 @@
 #include "database/db_settings.h"
-#include <cstdio>
 #include "napi/macros.h"
 #include "core/platform.h"
 #include "napi/helpers.h"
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Cool

PR #586, opened 2026-05-26
Update readme for new write buffer manager config

cb1kenobi:
Documents the new config settings added in https://github.com/HarperFast/rocksdb-js/pull/584.


kriszyp (approved):
Great, thank you!

PR #599, opened 2026-05-28
fix: treat dropped column family as empty in clear()

### LINKED ISSUE #598: clear() should return OK when column family was already dropped
## Summary

`DBHandle::clear()` in `src/binding/database/db_handle.cpp` calls `compactRange` on the column family handle. If the CF has already been dropped (e.g., by a concurrent `drop_table` operation), `compactRange` returns `kColumnFamilyDropped` and `clear()` propagates this as a JavaScript error (`ERR_COLUMN_FAMILY_DROPPED` / "Clear failed: Column family dropped...").

From the caller's perspective, calling `clear()` on a CF that is already dropped should be a **no-op success**: the goal of `clear()` is to leave the CF empty, and a dropped CF is already effectively empty (it doesn't exist). Failing here is unnecessarily strict and forces callers to add ad hoc error handling for a case they can't distinguish from a more serious failure.

## Current behaviour

```cpp
rocksdb::Status DBHandle::clear() {
    rocksdb::Status status = this->descriptor->compactRange(cf, nullptr, nullptr);
    if (!status.ok()) {
        return status;  // kColumnFamilyDropped propagated as error
    }
    return rocksdb::DeleteFilesInRange(...);
}
```

## Desired behaviour

If `compactRange` returns `kColumnFamilyDropped`, return `Status::OK()`. The caller's intent (empty CF) is already satisfied — there is nothing to clear. Subsequent writes to the same handle will still fail with `ERR_COLUMN_FAMILY_DROPPED`, allowing callers to detect and handle the dropped-CF state at write time (which is the right place).

## Observed in production

Harper (harper-pro) `runIndexing()` calls `clear()` before rebuilding a `@computed @indexed` attribute. When `clear()` raises `ERR_COLUMN_FAMILY_DROPPED`, the outer error handler only logs and exits, leaving `isIndexing` stuck with no persistent failure marker. The node then 503s all searches on that attribute indefinitely across restarts.

## Companion fix

Harper issue HarperFast/harper#<N>: `runIndexing` outer `catch` should persist `indexingFailed = true`.

kriszyp:
## Summary

- `DBHandle::clear()` now returns `OK` when `compactRange` reports `kColumnFamilyDropped` rather than propagating the error
- A dropped CF is already empty, so `clear()` is a no-op in this case

## Context

When a concurrent `drop_table` races with `clear()` (e.g. `runIndexing` clearing an index CF just as the table is dropped), RocksDB returns `kColumnFamilyDropped` from `compactRange`. Before this fix, that error surfaced to JavaScript callers as `ERR_COLUMN_FAMILY_DROPPED`, which forced them into an unnecessary error state. Subsequent writes to the same handle will still receive the error at write time and can handle recovery there.

Closes #598

> Generated by Claude Sonnet 4.6

### DIFF
--- a/src/binding/database/db_handle.cpp
+++ b/src/binding/database/db_handle.cpp
@@ -35,6 +35,12 @@ rocksdb::Status DBHandle::clear() {
 		nullptr
 	);
 	if (!status.ok()) {
+		// A dropped column family is effectively already empty — clear is a no-op.
+		// Callers that subsequently write to the same handle will receive
+		// kColumnFamilyDropped at write time and can handle recovery there.
+		if (status.IsColumnFamilyDropped()) {
+			return rocksdb::Status::OK();
+		}
 		return status;
 	}
 	// it appears we do not need to call WaitForCompact for this to work
--- a/test/clear.test.ts
+++ b/test/clear.test.ts
@@ -43,6 +43,13 @@ describe('Clear', () => {
 			10_000
 		);
 
+		it('should succeed when column family has been dropped', () =>
+			dbRunner({ dbOptions: [{ name: 'test' }] }, async ({ db }) => {
+				db.putSync('key', 'value');
+				db.dropSync();
+				await expect(db.clear()).resolves.toBeUndefined();
+			}));
+
 		it('should only remove entries in one column family', () =>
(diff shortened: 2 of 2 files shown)


cb1kenobi (approved):
Looks good, but this PR is blocked until you approve https://github.com/HarperFast/rocksdb-js/pull/600 :)

PR #612, opened 2026-06-02
Harden transaction-log readers against corrupt entry lengths

kriszyp:
## Summary

Hardens the two transaction-log **read** paths so an already-corrupt `.txnlog` fails with a precise, bounded error instead of OOMing or crash-looping.

A torn/truncated entry can leave a 4-byte big-endian length header claiming far more data than is physically present. (The *writer* bug that produces this is fixed separately in #573; this PR is the complementary read-side hardening, since corrupt files already exist in the field on un-upgraded nodes and #573 can't retroactively repair them.) Two reader defects turned such a file into a fatal:

- **`parse-transaction-log.ts`** — `Buffer.allocUnsafe(length)` on a bogus ~2 GB length, then the short-read error path hex-dumped the oversized buffer → `Cannot create a string longer than 0x1fffffe8` / heap OOM during startup "replaying transaction logs" replay.
- **`transaction-log-reader.ts` `next()`** — added the bogus length to `position`, then the advance-to-next-log path dereferenced an undefined buffer (`getLogMemoryMap(...)!` returning `undefined`) → `Cannot read properties of undefined (reading 'dataView')` on every commit during replication catch-up.

## Changes

- `parse-transaction-log.ts`: validate declared length against bytes remaining **before** allocating; throw a bounded `Corrupt entry at offset…` error. Cap the short-read error's hex preview to the first 64 bytes actually read.
- `transaction-log-reader.ts`: bound each entry against the committed `size` (or the physical mapped buffer when reading uncommitted); add a truncated-entry-header guard; throw a bounded `RangeError` on overrun. Convert the unguarded `getLogMemoryMap(...)!` into an `undefined` check that stops iteration cleanly (matching the existing guarded pattern a few lines above).
- Tests: new `corruption handling` block covering the parser (overrun length → bounded error, no oversized string) and the reader iterator (overrun length → bounded `RangeError`, valid data still readable afterward).

## Where to focus review

- **Throw vs fail-soft (design decision):** a corrupt *length* throws loudly (genuine corruption — surfacing it beats silently truncating the log view and risking replication divergence), whereas a transient *missing memory map* (rotation / FS race) stops cleanly with `done`. Is that distinction right for the replication-catch-up consumer, or should a corrupt length also stop cleanly?
- **The bound:** committed reads are bounded by `size` (a true entry boundary, so no false positives); uncommitted reads only by the physical buffer length. This was tightened from buffer-length-only after cross-model review — please sanity-check the `readUncommitted` branch.
- Mid-buffer subtle corruption (a length smaller than the bound but still misframed) is intentionally *not* caught here — that's the writer's job (#573 + the follow-up truncate-on-open PR).

## Test plan

- [x] `test/transaction-log.test.ts` — full file (53 tests) passes, incl. new corruption cases
- [x] Full vitest suite — 458 passed / 1 skipped, 0 failures
- [x] `pnpm type-check`, `pnpm lint`, `pnpm fmt` clean

A companion PR will address the writer side (truncate-to-last-valid-frame on open), closing #573's documented out-of-scope `O_APPEND` gap.

🤖 Generated by Claude (Opus 4.7) for @kriszyp. Cross-model reviewed via Gemini before opening.

### DIFF
--- a/src/parse-transaction-log.ts
+++ b/src/parse-transaction-log.ts
@@ -1,7 +1,7 @@
 import { constants } from './load-binding.js';
 import { closeSync, openSync, readSync, type Stats, statSync } from 'node:fs';
 
-const { TRANSACTION_LOG_TOKEN } = constants;
+const { TRANSACTION_LOG_TOKEN, TRANSACTION_LOG_ENTRY_HEADER_SIZE } = constants;
 
 interface LogEntry {
 	data: Buffer;
@@ -46,10 +46,13 @@ export function parseTransactionLog(path: string): TransactionLog {
 		const bytesRead = readSync(fileHandle, buffer, 0, numBytes, fileOffset);
 		fileOffset += bytesRead;
 		if (bytesRead !== numBytes) {
+			// Only hex-dump the bytes actually read, capped at 64. A corrupt length
+			// can request a huge read; dumping the whole (largely uninitialized)
+			// buffer can exceed V8's max string length and mask the real error.
+			const previewBytes = Math.min(bytesRead, 64);
+			const preview = buffer.subarray(0, previewBytes).toString('hex');
 			throw new Error(
-				`Expected to read ${numBytes} bytes but only read ${bytesRead}, file offset: ${fileOffset}, file size: ${size}, file path: ${path}, buffer: ${buffer.toString(
-					'hex'
-				)}`
+				`Expected to read ${numBytes} bytes but only read ${bytesRead}, file offset: ${fileOffset}, file size: ${size}, file path: ${path}, buffer (first ${previewBytes} bytes): ${preview}`
 			);
 		}
 		return buffer;
@@ -81,6 +84,18 @@ export function parseTransactionLog(path: string): TransactionLog {
 			}
 			const length = read(4).readUInt32BE(0);
 			const flags = read(1).readUInt8(0);
+			// Validate the declared length against the bytes actually remaining
+			// before allocating. A torn/corrupt entry can claim a length far larger
+			// than the file (e.g. a partial write that left header bytes pointing
+			// past the data), which would otherwise drive an unbounded allocUnsafe
+			// and OOM the process during log replay.
+			const remaining = size - fileOffset;
+			if (length > remaining) {
+				const entryOffset = fileOffset - TRANSACTION_LOG_ENTRY_HEADER_SIZE;
+				throw new Error(
+					`Corrupt entry at offset ${entryOffset}: declared length ${length} exceeds ${remaining} bytes remaining (file size: ${size})`
+				);
+			}
 			const data = read(length);
 			entries.push({ timestamp, length, flags, data });
 		}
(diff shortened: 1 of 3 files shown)


src/parse-transaction-log.ts
```diff
@@ -99,31 +84,23 @@ export function parseTransactionLog(
...
+			// Validate the declared length against the bytes actually remaining
+			// before allocating. A torn/corrupt entry can claim a length far larger
+			// than the file (e.g. a partial write that left header bytes pointing
+			// past the data), which would otherwise drive an unbounded allocUnsafe
+			// and OOM the process during log replay.
+			const remaining = size - fileOffset;
+			if (length > remaining) {
+				const entryOffset = fileOffset - TRANSACTION_LOG_ENTRY_HEADER_SIZE;
+				throw new Error(
+					`Corrupt entry at offset ${entryOffset}: declared length ${length} exceeds ${remaining} bytes remaining (file size: ${size})`
```
cb1kenobi:
```suggestion
					`Corrupt entry at offset ${entryOffset.toString(16)}: declared length ${length} exceeds ${remaining} bytes remaining (file size: ${size})`
```

src/parse-transaction-log.ts
```diff
@@ -59,10 +46,13 @@ export function parseTransactionLog(
...
+			// Only hex-dump the bytes actually read, capped at 64. A corrupt length
+			// can request a huge read; dumping the whole (largely uninitialized)
+			// buffer can exceed V8's max string length and mask the real error.
+			const previewBytes = Math.min(bytesRead, 64);
+			const preview = buffer.subarray(0, previewBytes).toString('hex');
 			throw new Error(
-				`Expected to read ${numBytes} bytes but only read ${bytesRead}, file offset: ${fileOffset}, file size: ${size}, file path: ${path}, buffer: ${buffer.toString(
-					'hex'
-				)}`
+				`Expected to read ${numBytes} bytes but only read ${bytesRead}, file offset: ${fileOffset}, file size: ${size}, file path: ${path}, buffer (first ${previewBytes} bytes): ${preview}`
```
cb1kenobi:
```suggestion
				`Expected to read ${numBytes} bytes but only read ${bytesRead}, file offset: ${fileOffset.toString(16)}, file size: ${size}, file path: ${path}, buffer (first ${previewBytes} bytes): ${preview}`
```

cb1kenobi (changes requested):
Can we output the file offsets in hex? This makes it easier when viewing the logs in a hex editor.

cb1kenobi (approved):

kriszyp:
Thanks @cb1kenobi — good call, hex offsets line up much more cleanly with a hex editor. The `parse-transaction-log.ts` offsets went in via your committable suggestions (8dcdc3e0); I've now applied the same convention to the `transaction-log-reader.ts` error messages in cb40f478, so `position` prints in hex there too (sizes/lengths left decimal to match). Ready for another look when you get a chance.

— Claude (assisting Kris)

PR #614, opened 2026-06-02
Recover torn transaction-log tails on open

kriszyp:
## Summary

Closes the `O_APPEND` gap that the writev fix (#573) explicitly left **out of scope**. On POSIX the log is opened `O_RDWR|O_CREAT|O_APPEND` and on open `size = fstat().st_size`, so a crash that interrupts an append mid-write leaves a **torn partial entry at the tail** of the active log file (a 13-byte header claiming more data than was written). On restart the reader misframes on it — the same corruption class #573 prevents for the live writer, but #573 can't repair bytes already on disk.

On store load, the active (highest-sequence) log file is scanned and, if its tail is a torn partial entry with no valid framing after it, **truncated back to the last valid entry boundary and `fsync`'d**.

This is the writer-side companion to the reader hardening in #612 (the two are independent and can land in either order).

## Design

- **`transaction_log_recovery.{h,cpp}`** — a *pure* (no I/O) scan, `scanTransactionLogForRecovery(data, fileSize)`, classifying the file as `Clean` / `TruncateTail(validEnd)` / `MidFileCorruption(breakOffset)`. Unit-tested standalone.
- **`recoverTail()`** (`transaction_log_file.cpp`) — reads the file once, runs the scan, and truncates / logs accordingly.
- **`truncateFile()`** — POSIX `ftruncate`+`fsync`; Windows is a documented no-op (it pre-extends + zero-pads, so torn tails surface as the zero-padding end marker the scan treats as `Clean`).
- Wired into `TransactionLogStore::load()` to run **once, on the active file only** (rotated files are immutable).

## Safety properties (the load-bearing parts to review)

- **Truncates to the last intact frame found by scanning, never to a stored offset.** The durable `txn.state` position is the last *flushed* position (conservative); the log legitimately extends past it with committed, already-replicated entries — truncating to it would destroy replication history.
- **Entry length is bounded only by the physical file size, never the rotation threshold.** A single entry can legitimately exceed it (the first entry in a fresh file is always written whole), so a rotation-size cap would misclassify a large committed entry as broken and truncate it. *(This was a data-loss bug caught in cross-model review and fixed.)*
- **Mid-file corruption is left intact and logged loudly, never truncated** (boot-anyway) — dropping it would discard committed entries. Resumed framing is detected by a run of ≥8 frames **or** a shorter run that lands exactly on EOF, so trailing committed entries are never truncated away. *(The "lands on EOF" signal was added in review to fix a false-negative that could truncate <8 trailing committed entries.)*

## Where to focus review

- `validFramingResumes` heuristic — is "≥8 frames OR a run reaching EOF exactly" the right resync signal? It fails *safe* (a false positive leaves a torn tail unrepaired rather than dropping data), but please sanity-check the EOF-coincidence reasoning.
- The `load()` wiring: recovery runs after the registration loop and refreshes `nextLogPosition` from the corrected size.
- Alignment: the byte-by-byte scan calls `readDoubleBE`/`readUint32BE` at arbitrary offsets — verified alignment-safe (they assemble bytes individually, like the existing `findPositionByTimestamp`).

## Test plan

- [x] Native GoogleTest (`pnpm test:native`): clean, zero-pad tail, torn tail (overrun + partial header), large entry > rotation size, mid-file corruption, and a break followed by <8 entries reaching EOF — 20 tests pass
- [x] JS integration (`test/transaction-log.test.ts`): corrupt an active log's tail on disk, reopen, assert the torn bytes are truncated, committed entries preserved, log still writable; plus a clean-file-untouched case
- [x] Full vitest suite: 458 passed / 1 skipped, 0 failures; `pnpm type-check` / `lint` / `fmt` clean

🤖 Generated by Claude (Opus 4.7) for @kriszyp. Cross-model reviewed via Gemini; two data-loss findings were fixed before opening (see Safety properties).


src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -122,6 +125,70 @@ void TransactionLogFile::open(const double latestTimestamp) {
...
+			// truncating would discard committed/replicated transactions. Surface
+			// it so an operator can repair the file; the reader's per-entry bounds
+			// checks refuse to return the broken frame.
+			DEBUG_LOG("%p TransactionLogFile::recoverTail Mid-file corruption at offset %u in %s (size=%u), leaving intact\n",
+				this, scan.validEnd, this->path.string().c_str(), fileSize);
+			::fprintf(stderr,
+				"[rocksdb-js] WARNING: transaction log %s has a framing break at offset %x with %u byte(s) of further "
+				"data; leaving it intact to avoid discarding committed entries. Reads past this point will fail until "
+				"the file is repaired.\n",
+				this->path.string().c_str(), scan.validEnd, fileSize - scan.validEnd);
```
cb1kenobi: We should not write to stderr. Comment out this `fprintf` and we'll replace it with the process-wide event emitter when it's ready (https://github.com/HarperFast/rocksdb-js/pull/621).

```suggestion
//			::fprintf(stderr,
//				"[rocksdb-js] WARNING: transaction log %s has a framing break at offset %x with %u byte(s) of further "
//				"data; leaving it intact to avoid discarding committed entries. Reads past this point will fail until "
//				"the file is repaired.\n",
//				this->path.string().c_str(), scan.validEnd, fileSize - scan.validEnd);
```

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -122,6 +125,70 @@ void TransactionLogFile::open(const double latestTimestamp) {
...
+				this, this->path.string().c_str(), fileSize, scan.validEnd);
+			if (this->truncateFile(scan.validEnd)) {
+				this->size.store(scan.validEnd, std::memory_order_relaxed);
+				if (this->lastFlushedSize > scan.validEnd) {
+					this->lastFlushedSize = scan.validEnd;
+				}
+				::fprintf(stderr,
+					"[rocksdb-js] transaction log %s had a torn tail; dropped %u partial byte(s) back to the last valid "
+					"entry (new size=%u).\n",
+					this->path.string().c_str(), fileSize - scan.validEnd, scan.validEnd);
```
cb1kenobi: Same here:


```suggestion
//				::fprintf(stderr,
//					"[rocksdb-js] transaction log %s had a torn tail; dropped %u partial byte(s) back to the last valid "
//					"entry (new size=%u).\n",
//					this->path.string().c_str(), fileSize - scan.validEnd, scan.validEnd);
```

cb1kenobi (changes requested):
When validating the transaction log, we should be checking that each block's timestamp is monotonically increasing and that the timestamps are plausible Unix epoch values (not too far in the past, not far in the future).

cb1kenobi:
I've created PR https://github.com/HarperFast/rocksdb-js/pull/640 that removes the fprintf calls and replaces it with the process-wide global event emitter.

cb1kenobi (approved):

PR #619, opened 2026-06-03
fix(build): link zlib so the macOS linker can resolve RocksDB's compressor

kriszyp:
## Summary

`librocksdb.a`'s compression object (`BuiltinZlibCompressor`) references zlib's `_deflate`/`_inflate` but does **not** bundle zlib. `binding.gyp` links `librocksdb.a` without the `libz.a` shipped beside it in the RocksDB prebuild, so zlib is only resolvable by luck of the link.

Whether the linker pulls that compression object in depends on the rest of the target's object set: GNU `ld` (Linux) currently leaves it out, so builds pass — but macOS's linker can pull it, and then the build fails with `Undefined symbols: _deflate for architecture arm64`. It's a **latent gap** on `main` today (no target trips it yet); the Verification Table branch (#526) happens to trip it, which is how it surfaced.

Harden it proactively: link `deps/rocksdb/lib/libz.a` after `librocksdb.a` (GNU `ld` is order-sensitive) in both the main binding and the native-tests targets, under the `linux`/`mac` condition. Windows is unaffected (links `rocksdb.lib`).

## Notes

- `libz.a` ships alongside `librocksdb.a` in the RocksDB prebuild bundle used by CI.
- Extracted from #526 (Verification Table) so the build hardening can land on `main` independently.

## Test plan

- [ ] CI green across Linux/macOS/Windows × Node/Deno/Bun (this is the platform matrix that exercises the link).


cb1kenobi (approved):
Awesome!

PR #625, opened 2026-06-04
Add RocksDB backup and restore support

cb1kenobi:
## Summary

Adds RocksDB backup/restore support via RocksDB's `BackupEngine`.

- **`db.backup(backupDir, options?): Promise<number>`** — instance method (needs the open DB); creates a whole-database backup (all column families) and resolves the new backup id. Creates parent directories as needed. Defaults `flushBeforeBackup` to `true` when the DB was opened with `disableWAL` so unflushed data isn't lost.
- **`backups` namespace** (path-based, no open DB): `restore`, `list`, `delete`, `purge`, `verify`. Mirrors how `constants`/`stats` are exported.
- **CLI `backups` command** (`bin/rocksdb-js.mjs`): list/backup/restore/verify/delete/purge from the REPL. Restore prompts for confirmation, is blocked in `--readonly` mode, and closes/reopens all column family handles around the destructive restore. Delete/purge report recovered disk space. Tab completion covers the backup dir path, subcommands, and backup ids.

C++: new `src/binding/database/backup.cpp` — `Database::Backup` (instance, registered on the Database class) plus 5 module-level N-API functions registered in `binding.cpp`. All async (libuv thread pool), following the existing `Flush`/`BaseAsyncState` pattern.

## Purpose

rocksdb-js had no way to back up or restore a database. This exposes RocksDB's purpose-built incremental, checksum-verified backup engine to JS.

## Where to focus

- **C++ lifetime under concurrent close** (`backup.cpp`): the async backup state holds its own `shared_ptr<DBDescriptor>` so the underlying `rocksdb::DB` stays alive for the whole copy. `DBHandle::close()` resets the handle's descriptor after a bounded 5s wait, so relying on `handle->descriptor` from the worker thread would be a use-after-free for backups that outrun that wait. Worth a careful look.
- **Destructive restore + path guard** (`backup.ts`): `restore`'s default mode purges the target dir; `backups.restore` resolves both paths before the `backupDir === dbDir` guard so trailing-slash/relative variants can't slip a destructive purge onto the backup dir itself.
- **Backup granularity**: a backup is the whole physical database, not per-`Store`/column-family — reflected in the API and docs.
- **CLI restore failure handling** (`bin/rocksdb-js.mjs`): if the destructive restore throws after the DB handles are closed, the CLI exits instead of reopening, because a read-write reopen would silently create a new empty database over the partially-restored directory. Exiting the REPL on a failed restore is a deliberate trade-off — flag if you'd rather it attempt a guarded reopen.

## Testing

`test/backup.test.ts` (18 tests): round-trip, parent-dir creation, incremental + list, metadata, restore-by-id vs latest, delete/purge/verify (incl. missing-id → `ERR_NOT_FOUND`), `disableWAL` auto-flush, custom `walDir`, non-incremental (`shareTableFiles: false`), incremental restore mode, path-variant guard, a close-during-backup regression, backup from a read-only handle, and restore-while-a-read-only-handle-is-open (POSIX only; Windows cannot delete open files). Full suite: 509 passed / 1 skipped.

The CLI has no automated test coverage (pre-existing); the `backups` command was manually exercised end-to-end: list/create/verify/delete/purge, restore round-trip (data verified rolled back), declined confirmation, `--readonly` blocking, same-directory guard, invalid ids/subcommands/keep-counts, and tab completion at every argument position (via a PTY).

## Notes

- Generated by an LLM (Claude Opus 4.8; CLI command by Claude Fable 5).
- **Cross-model review (API/C++ work):** only the Gemini/AntiGravity (`agy`) review ran — the `codex` CLI wasn't installed in that environment. The Gemini review's blocker (close-during-backup UAF) and significant findings (path-guard bypass, missing tests) are all addressed in this branch.
- **Cross-model review (CLI command):** both Codex and Gemini reviewed; all findings addressed (failed-restore reopen hazard, uint32 purge keep-count wrap, dirSize race hardening).
- **Open suggestions left unaddressed (reviewer's call to weigh in):** (1) `IOStatus`→`Status` assignment slices I/O subcodes — kept deliberately since `createRocksDBError` only reads `status.code()`, matching existing patterns; (2) a type-only import cycle `load-binding.ts ↔ backup.ts` (erased at compile, lints/types clean), consistent with the existing `load-binding ↔ store` cycle; (3) the `BaseAsyncState` cancellation path doesn't `napi_delete_reference` — a pre-existing project-wide pattern, not introduced here.

Required for https://github.com/HarperFast/harper/issues/131


kriszyp (approved):
This looks awesome!

Issue #628, opened 2026-06-04
Use-after-free in getUserSharedBuffer external ArrayBuffer lifetime


cb1kenobi:
## Summary
`UserSharedBufferFinalizeData` holds the underlying `UserSharedBufferData` via a `std::weak_ptr`, while the only strong reference lives in `ColumnFamilyDescriptor::userSharedBuffers`. When the last `DBHandle` closes and `DBDescriptor::close()` runs `this->columns.clear()`, the data is freed even though external V8 `ArrayBuffer`s still hold raw pointers into it. Subsequent JS access is a use-after-free.

## Where
- `src/binding/database/db_descriptor.h:410` - `std::weak_ptr<UserSharedBufferData> sharedData;`
- `src/binding/database/db_descriptor.cpp:205` - `this->columns.clear()` in `DBDescriptor::close()` drops the only strong ref
- `src/binding/database/db_descriptor.cpp:1365` - `napi_create_external_arraybuffer(env, userSharedBuffer->data, …)` hands V8 a raw pointer with no lifetime guarantee
- `src/binding/database/db_descriptor.h:460` - `releaseUserSharedBuffer`'s `use_count() <= 2` heuristic

## Trigger
1. Main and a worker both open the same path+CF (shared `DBDescriptor`).
2. Both call `getUserSharedBuffer(key, …)`. The binding creates one `UserSharedBufferData` in `userSharedBuffers[key]` and hands each thread an external `ArrayBuffer` over the same raw `data` pointer.
3. Main closes its handle. Worker closes its handle. Registry sees `use_count <= 1` and calls `descriptor->close()`.
4. `columns.clear()` runs `~ColumnFamilyDescriptor`, which destroys the `userSharedBuffers map`. The strong `shared_ptr<UserSharedBufferData>` drops to refcount 0 (finalize data only holds a `weak_ptr`). `~UserSharedBufferData()` runs `delete[] this->data`.
5. Either retained `ArrayBuffer` still references the freed `data`. Any access is UAF.

## Observed crash
`BUG IN CLIENT OF LIBMALLOC: memory corruption of free block`, SIGTRAP (exit 133). Trapping frame is downstream in V8 turbofan (`v8::internal::compiler::ZoneStats::NewEmptyZone` → `operator new`) — V8's compiler had reallocated over the freed region and tripped a free-list invariant.

[...]

PR #629, opened 2026-06-04
Fix use-after-free in `getUserSharedBuffer()` when DB closes with retained `ArrayBuffers`

cb1kenobi:
## Summary

User-shared `ArrayBuffer`s returned from `db.getUserSharedBuffer()` could outlive the underlying storage when the last `DBHandle` for a column family closed. `DBDescriptor::close()` cleared `columns`, destroying the
`ColumnFamilyDescriptor` and freeing the `UserSharedBufferData::data` heap allocation while JS still held the external `ArrayBuffer`, leaving a dangling pointer. Any subsequent read/write — or another open/close cycle that re-allocated over the freed bytes — would corrupt the heap and SIGSEGV / SIGTRAP. This was reproducible from Harper schema-sync workflows.

The fix moves the strong reference to `UserSharedBufferData` into the external `ArrayBuffer`'s N-API finalize data so the storage lives as long as any JS reference does. The map entry in `ColumnFamilyDescriptor::userSharedBuffers` becomes one of N+1 strong refs (one per live `ArrayBuffer` + the map); the finalizer evicts the map entry when it sees `use_count() <= 2` (map + itself). On `DBDescriptor::close()`, the map clear no longer frees storage; storage is released at GC time when the last `ArrayBuffer` finalizer runs.

## What to look at

- [`src/binding/database/db_descriptor.h`](src/binding/database/db_descriptor.h):
  `UserSharedBufferFinalizeData::sharedData` changed from `weak_ptr` to `shared_ptr`. `ColumnFamilyDescriptor::releaseUserSharedBuffer` now takes `const shared_ptr&` so the existing `use_count() <= 2` invariant is accurate (previously the by-value parameter inflated the count by one and the map-entry eviction in that path was effectively dead — entries only cleared on CF teardown).
- [`src/binding/database/db_descriptor.cpp`](src/binding/database/db_descriptor.cpp):
  finalizer drops the `weak_ptr::lock()` dance, passes the strong ref directly, and falls through to `delete finalizeData` which now releases storage when this was the last retained `ArrayBuffer`.
- The `weak_ptr<DBHandle>` and `weak_ptr<ColumnFamilyDescriptor>` are kept for opportunistic cleanup (listener removal, map-entry eviction); only the storage is pinned strongly. This is the minimum needed to fix the UAF.

## Tests

Four child-process repro modes added under
`test/fixtures/fork-user-shared-buffer-uaf.mts` and exercised by
`test/user-shared-buffer.test.ts`:

- `buffer` — main + worker close all handles, then main hammers the retained ArrayBuffer
- `buffer-race` — hammer loop runs concurrent with the close (highest pre-fix repro rate: ~50% on macOS-arm64)
- `schema-sync` — Harper-like cycle: holder worker pins the descriptor, sync worker repeatedly opens/closes the same CF, main spins on the shared buffer
- `stale-iterator` — moved to its own `Descriptor teardown races` describe; exercises iterator-vs-close, not the buffer UAF, but it's the same descriptor-lifetime class of bug

Each mode is spawned in a child process (a failing build SIGSEGVs/SIGTRAPs rather than throwing) and looped 3× per `it()` to keep CI detection rate high for the underlying race. The in-process "should not crash if you close while holding a reference to the buffer" test was extended to read/write the retained buffer after close as a fast regression guard.

Pre-fix repro rate (direct, 10 runs each):
- buffer-race: 5/10 SIGTRAP
- buffer / schema-sync: occasional

Post-fix: 0 failures across 15 vitest runs of the file, 10 direct `buffer-race` runs, full `pnpm test` (464 pass), `pnpm test:native` (12/12), `pnpm check`.

🤖 Generated by Claude (claude-opus-4-7 with 1M context).

Ticket: https://github.com/HarperFast/rocksdb-js/issues/628


kriszyp (approved):
Great job!

A few notes from KrisAI review (minor, nothing blocking):

* Potential message-drop in runStaleIteratorMode: After the iterator loop finishes, the code registers worker.once('message', ...) to wait for { closed: true }. The worker may have already sent that message while the main thread was busy in the iterator loop (it runs in a separate OS thread). If the message arrives before the listener is registered, the test will hang. The close worker + message could be restructured to avoid this window.
* terminateWorker import in user-shared-buffer.test.ts — appears unused after the stale-iterator inline test was moved to expectReproSurvives. Worth checking — TypeScript should flag this, but if noUnusedLocals isn't strict it'll silently stay.
* Cosmetic: generateDBPath is imported on a separate line from the other ./lib/util.js exports (line 3 vs line 2). Easy to consolidate.

Issue #626, opened 2026-06-04
Expose transaction log stats


cb1kenobi:
There's currently no insights into the loaded transaction log stores, log files, memory maps, etc. This information would be helpful to get a sense of non-RocksDB usage and whether any transaction log settings should be tuned.

cb1kenobi:
Fixed in https://github.com/HarperFast/rocksdb-js/pull/636

PR #636, opened 2026-06-09
Expose transaction log statistics

cb1kenobi:
## Summary

Adds observability for the transaction log subsystem, which previously exposed none of its internal state as metrics, and reworks how stat names are catalogued.

- **`log.getStats()`** — a detailed per-log snapshot: file/transaction gauges, **memory-map + POSIX overlay usage**, recovery positions (next / committed / flushed + `replayGapBytes`), purge/retention gauges, and lifetime counters.
- **`db.getStats()`** — gains a summarized, aggregate `txnlog.*` key set across all of the database's logs, present **regardless of whether RocksDB statistics are enabled**.
- **`db.getStat('txnlog.…')`** — resolves the aggregate value for an individual txnlog key (unknown `txnlog.*` keys return `undefined` rather than throwing).
- **Stat catalog moved to docs** — removed the `stats.tickers` / `stats.histograms` arrays (and the dead `getStat*Names` decls); every stat is now documented in [`docs/stats.md`](docs/stats.md) with type/internal/essential columns. `stats.StatsLevel` is unchanged.

Implemented via a NAPI-free `TransactionLogStore::collectStats(StoreStats&)` reused by both the per-log method and the DB-level aggregate; the txnlog summary key set is defined once as an X-macro shared by `getStats()`/`getStat()`.

## ⚠️ Breaking change

`stats.tickers` and `stats.histograms` (on the `stats` export) are **removed**. Anything enumerating those arrays must instead consult `docs/stats.md`. Values are still readable via `db.getStats()` / `db.getStat()`. Flagging since this is a public-API removal — confirm nothing downstream depends on those arrays.

## Where to focus review

- **Concurrency in `collectStats`** (`transaction_log_store.cpp`): reads `getLastFlushedPosition()` *before* `dataSetsMutex` (honoring `dataSetsMutex → flushedStateMutex` ordering), copies each file's `memoryMap` under `fileMutex` (the lock that guards it), and calls `getLastWriteTime()` under `dataSetsMutex` only (matching `doPurge()`). Please sanity-check the lock ordering.
- **Memory semantics**: `memory.mappedBytes` is virtual address space (full `maxFileSize` map on POSIX), not resident; `memory.overlayBytes` (POSIX-only) is the closer proxy. Documented in `docs/stats.md` + the type.
- **`advanceSequence()` helper**: folds the six identical `currentSequenceNumber = nextSequenceNumber++` sites into one helper that also bumps the `rotations` counter — behavior unchanged.
- **`replayGapBytes`** sums the unflushed span across files (approximation; includes file headers).

## Notes

- No timing histograms in this pass (write-batch latency, purge duration) — deferred.
- A native GoogleTest for `collectStats` was intentionally skipped: the `rocksdb-js-native-tests` target only links `transaction_log_file.o`, not the store; `collectStats` is covered end-to-end by `test/transaction-log-stats.test.ts` instead.
- The two cross-model reviews (Codex/Gemini) from the Harper lifecycle were **not run by the agent** — blocked locally as external data egress. Being run separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8)


kriszyp (approved):
This looks great!

cb1kenobi:
Fixes #626

PR #637, opened 2026-06-09
fix(transaction-log): stop the read path from truncating the append-owned file size (#1148)

kriszyp:
## Summary
`findPositionByTimestamp` (the txn-log point read) mutated `this->size` — the atomic written-extent counter **owned by the append path** — when it read a zero timestamp while indexing. Under concurrent appends that zero is a not-yet-visible memory-map artifact, not EOF, so truncating froze the index and made point reads intermittently return `null` for committed entries, breaking the out-of-order duplicate detection in `Table.commit` that depends on them.

## Purpose
Root cause for HarperFast/harper#1148 — the intermittent, load-sensitive transaction-log point-read misses — and therefore the load-sensitive #1137 commutative-op double-apply that depends on this read.

## Root cause — a regression introduced by the anonymous mmap overlay
The truncation (`this->size = lastIndexedPosition` on a zero timestamp) **predates the overlay** (authored late 2025) and was correct under a **direct page-cache mmap**: on POSIX, reads of `[0, size)` return real data, so the only zero it ever hit was a genuine zero-padded EOF (the Windows pre-allocation case its comment describes). The **anonymous-overlay scheme (`d0b8fcf5`, 2026-05-22)** later introduced anonymous zero pages in the `[lastOverlaySize, size)` window — the just-appended tail not yet re-overlaid by `updateMemoryMapOverlay`, which necessarily runs *after* `size += ` (it derives its extent from `size`). Under concurrent appends a reader observes the bumped `size`, reads that not-yet-exposed tail through the overlay, gets an anonymous zero, and the (previously-correct) truncation misreads it as EOF — freezing the index and intermittently hiding committed entries.

So **#1148 is a regression from the overlay**, not a latent flaw in the point read. Note the irony: the overlay's anonymous-zero tail is a *defense* against out-of-bounds reads (which a correctly-bounded reader never makes) — and that defense is exactly what converted a would-be loud SIGBUS into silent corruption.

## Which fix to land — reviewer's call (@cb1kenobi)
The PR currently implements **(B)**. Three viable approaches — your decision on which to land:

**(A) Bound the index by the committed position.** Eliminates the over-read at the source for committed reads: the index never reads past the committed position, so it never touches the not-yet-exposed tail. Caveat (cross-model review found): must exempt `readUncommitted` queries and fix resumable-iterator positioning (return the committed boundary, not file size) — otherwise they miss uncommitted-tail entries. Adds a bit of JS↔native plumbing for the `readUncommitted` flag.

**(B) Read-side defensive fix — _currently implemented_.** Index the full extent; never let the read path truncate the append-owned `size`; return `lastIndexedPosition` on a not-yet-exposed tail (so the committed-bounded TS scan via `loadLastPosition` picks it up next call). Correct for all callers — committed / `readUncommitted` / resumable / replay — and contained, but symptom-handling rather than removing the over-read.

**(C) Revert the overlay.** Read the file's page cache directly (the pre-overlay behavior under which the point read was correct), removing the publish-before-expose window at its source. Larger change: must re-solve the stable, growable POSIX mapping the overlay provides — remap/extend on growth (fails *loud* (SIGBUS) rather than silently if a read ever races ahead of the remap), or pre-size the file to `maxFileSize` sparse (Windows-style).

## Where to look
- `findPositionByTimestamp` (`transaction_log_file.cpp`): the guarded truncation and the `lastIndexedPosition`-vs-`0xFFFFFFFF` return. Codex caught that *not truncating* alone still left a one-call transient miss (`lower_bound` → past-EOF for the unindexed tail); the `lastIndexedPosition` return closes that.
- `hasAppendedSinceOpen` (set in the `writeEntries` dispatcher, reset in `openFile`): the startup-replay vs steady-state discriminator. Uses default seq_cst (a reader observing the bumped `size` also observes the flag) — release/acquire was suggested by the Gemini/agy review but seq_cst chosen for verifiability on this cold path.

## Validation
The race is load-sensitive and not deterministically reproducible single-threaded, so coverage is the full rocksdb-js suite (460 passing / 1 skipped, incl. concurrency + out-of-order/`exactStart`) plus a new consistency guard test. The end-to-end symptom is harper#1137 (red on `main`), best confirmed by running the harper resources suite against this branch.

Cross-model review: Codex (found+fixed the transient-tail miss; flagged the `readUncommitted`/resumable caveat on approach A) and Gemini/agy (confirmed correctness, memory ordering, Windows reopen, out-of-order `exactStart`; no blockers).

🤖 Generated with Claude (Opus 4.7)


cb1kenobi (approved):
Good find! Tricky one.

PR #640, opened 2026-06-09
Add JSON-escape helper and ListenerData::fromStrings factory

cb1kenobi:
## Summary
Two native call sites (`new-transaction-log` event in `db_handle.cpp`, and the new mid-file-corruption warning carried in from this branch's base in `transaction_log_file.cpp`) were each building JSON-array listener payloads inline. The escaping was buggy in both — `db_handle.cpp` only handled `"`, and the new `transaction_log_file.cpp` emission did no escaping at all, so a Windows path or a name with `\` or any control char would yield malformed JSON that the listener trampoline can't parse.

Centralized into:
- `core/json.h` — header-only `appendJsonString` / `jsonString`, RFC 8259-compliant.
- `ListenerData::fromStrings(initializer_list<string_view>)` — wraps escape + array assembly + the new/move boilerplate. Defined inline so files that only need this factory (e.g. recovery code in the native GoogleTest binary) get the implementation without linking the rest of `event_emitter.cpp`.

Both call sites collapse to one line.

## Where to look
- **`test/native/event_emitter_stub.cc`** — `transaction_log_file.cpp` is also linked into the GoogleTest binary, which has no N-API runtime. The stub supplies bodies for the two EventEmitter member symbols pulled in transitively via `emitGlobalEvent` (`EventEmitter::notify` and `releaseAll()` from the destructor). `notify` is a no-op that still takes ownership of the payload, matching the production contract. Production and test now run the same recovery path.
- **Open Codex finding (not addressed):** event name `log.warn` doesn't match the documented namespaced-key convention (`AGENTS.md` calls for `transactionLog:warning` for internal events). The name was set in the base branch's WIP, not by this refactor — flagging in case it should be renamed before either PR lands.
- **Lifetime of `string_view`s in `fromStrings({ msg.str() })`** — the `std::string` temporary lives to end-of-full-expression, which outlives the call. Safe but worth a glance.

## Testing
- New native GTest (`test/native/json_test.cc`) covers empty / plain ASCII / `"` / `\` / named control chars / `\u00XX` escapes / UTF-8 pass-through / append behavior / mixed escapes (9 cases).
- Existing `transaction-log.test.ts` `'new-transaction-log'` event test exercises the factory end-to-end.
- All 30 native GTests, 474 Vitest tests, `pnpm check` (type + lint + fmt) green.

🤖 Generated with Claude Opus 4.7 (1M context) via Claude Code

### DIFF
--- a/src/binding/core/json.h
+++ b/src/binding/core/json.h
@@ -20,7 +20,8 @@ namespace rocksdb_js {
 inline void appendJsonString(std::string& out, std::string_view s) {
 	out.reserve(out.size() + s.size() + 2);
 	out += '"';
-	for (unsigned char c : s) {
+	for (char ch : s) {
+		const unsigned char c = static_cast<unsigned char>(ch);
 		switch (c) {
 			case '"':  out += "\\\""; break;
 			case '\\': out += "\\\\"; break;
--- a/src/binding/core/json.h
+++ b/src/binding/core/json.h
@@ -1,0 +1,57 @@
+#ifndef __CORE_JSON_H__
+#define __CORE_JSON_H__
+
+#include <cstdio>
+#include <string>
+#include <string_view>
+
+namespace rocksdb_js {
+
+/**
+ * Appends a JSON string literal (including surrounding double quotes) for
+ * `s` to `out`, escaping per RFC 8259: `"`, `\`, and control characters
+ * 0x00–0x1F. Bytes >= 0x20 are passed through verbatim, which means a
+ * valid UTF-8 input produces a valid UTF-8 JSON literal without
+ * re-encoding multi-byte sequences as `\uXXXX`.
+ *
+ * Inline so it can be used both from the N-API layer and from native
+ * GoogleTest binaries that don't link the N-API code.
+ */
+inline void appendJsonString(std::string& out, std::string_view s) {
+	out.reserve(out.size() + s.size() + 2);
+	out += '"';
+	for (unsigned char c : s) {
+		switch (c) {
+			case '"':  out += "\\\""; break;
+			case '\\': out += "\\\\"; break;
+			case '\b': out += "\\b";  break;
+			case '\f': out += "\\f";  break;
+			case '\n': out += "\\n";  break;
+			case '\r': out += "\\r";  break;
+			case '\t': out += "\\t";  break;
+			default:
+				if (c < 0x20) {
+					char buf[7];
+					std::snprintf(buf, sizeof(buf), "\\u%04x", c);
+					out.append(buf, 6);
+				} else {
+					out += static_cast<char>(c);
+				}
+		}
+	}
+	out += '"';
+}
+
+/**
+ * Convenience wrapper around `appendJsonString` that returns a freshly
+ * built JSON string literal.
+ */
+inline std::string jsonString(std::string_view s) {
+	std::string out;
+	appendJsonString(out, s);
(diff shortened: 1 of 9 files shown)


src/binding/core/json.h
```diff
@@ -0,0 +1,57 @@
...
+ * re-encoding multi-byte sequences as `\uXXXX`.
+ *
+ * Inline so it can be used both from the N-API layer and from native
+ * GoogleTest binaries that don't link the N-API code.
+ */
+inline void appendJsonString(std::string& out, std::string_view s) {
+	out.reserve(out.size() + s.size() + 2);
+	out += '"';
+	for (unsigned char c : s) {
+		switch (c) {
```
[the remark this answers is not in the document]
cb1kenobi: Fixed

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -152,19 +153,25 @@ void TransactionLogFile::recoverTail() {
...
-				this->path.string().c_str(), scan.validEnd, fileSize - scan.validEnd);
+
+			std::ostringstream msg;
+			msg << "Transaction log " << this->path.string()
+				<< " has a framing break at offset " << std::hex << scan.validEnd << std::dec
+				<< " with " << (fileSize - scan.validEnd)
+				<< " byte(s) of further data; leaving it intact to avoid discarding committed entries. "
+					"Reads past this point will fail until the file is repaired.";
+
+			emitGlobalEvent("log.warn", ListenerData::fromStrings({ msg.str() }));
```
kriszyp, line 171: In global_events.h says it should be "transactionLog:warning"? I am fine with this event name, just noting the inconsistency, whatever you prefer.

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -152,19 +153,25 @@ void TransactionLogFile::recoverTail() {
...
 
-		case RecoveryScan::Kind::MidFileCorruption:
+		case RecoveryScan::Kind::MidFileCorruption: {
 			// Leave the file intact: entries are still framed after the break, so
 			// truncating would discard committed/replicated transactions. Surface
 			// it so an operator can repair the file; the reader's per-entry bounds
 			// checks refuse to return the broken frame.
 			DEBUG_LOG("%p TransactionLogFile::recoverTail Mid-file corruption at offset %u in %s (size=%u), leaving intact\n",
 				this, scan.validEnd, this->path.string().c_str(), fileSize);
-			::fprintf(stderr,
```
kriszyp, line 162: So if there is no event handler, then there is no warning, completely silent? You think that is ok?

kriszyp (approved):
This looks cool. A couple questions for you to look at, but I approve.

src/binding/transaction_log/transaction_log_file.cpp
cb1kenobi: AI suggested `transactionLog:warning` for me also, but I was thinking more from the consumer perspective. It's a better UX to listen for all `log.warn` events than a bunch of different, but like, events.

src/binding/transaction_log/transaction_log_file.cpp
cb1kenobi: I think that's fine. I've documented the `log.warn` event with an example.

I also added a `DEBUG_LOG()` for debug builds.

PR #647, opened 2026-06-11
fix: free a column family name immediately on drop (ghost tables)

sleekmountaincat:
## Summary

Dropping a column family now unconditionally removes its by-name entry from the descriptor's registry map, so a subsequent open with the same name creates a fresh column family instead of reusing the dangling dropped handle. A new `columnsMutex` on `DBDescriptor` guards all concurrent accessors of the columns map, and `flush()`/`close()` now snapshot-and-pin the descriptors they iterate.

## Purpose

Root cause of Harper's "ghost table" failures (deleted tables resurrecting after restart, `Invalid column family specified in write batch` poisoning a whole database env, wedged table creation). Harper opens every table's column family from multiple worker threads, so the old `tryUnregisterColumnFamily` refcount gate (`use_count <= 2`) never fired: the dropped entry stayed registered, a same-name recreate got the dead handle, and one failed write contaminated the env's shared write path until process restart. Reproduced from scratch on a live 2-node Fabric cluster (harper-pro 5.0.27) with a plain table: drop + same-name recreate without a restart breaks the env every time.

## Where to focus review

1. **Intentional semantics change**: drop goes from deferred-while-referenced to immediate-name-free. Instances still holding the dropped column family can read it until they close; what's gone is the old "reopen by name resurrects it" behavior. The previous two-instance drop test encoded the deferred contract and was replaced with a regression test for the new one (`test/drop.test.ts`). This needs an explicit maintainer call.
2. **Lock ordering**: `databasesMutex` before `columnsMutex` everywhere both are held (`DBRegistry::OpenDB`, `close()` under `PurgeAll`); the drop path takes only `columnsMutex`, so it cannot interact with registry-lock holders that wait on in-flight operations at shutdown.
3. **`flush()` pinning** (`db_descriptor.cpp`): runs on libuv worker threads; it now copies the `shared_ptr`s under the mutex so a concurrent drop/close cannot destroy a handle mid-Flush. Previously this was an unsynchronized iteration that the new every-drop erase would have made a routine data race.
4. The last assertion in the new drop test documents that a write through a still-held dropped handle fails AND contaminates the env's write path; it must stay the final assertion.

## Validation

- Full vitest suite: 32 files, 459 passed, 1 skipped, 0 failed (macOS arm64 build; tsc, oxlint, oxfmt clean).
- End-to-end against live harper-pro 5.0.27: this fix backported to v1.4.2 (#648), built in-container, swapped into the running server. Drop + same-name recreate + insert works; rapid drop/recreate cycles work; no poison, no wedge; clean drops stay dropped across restart.
- Harper-side regression tests exist in HarperFast/harper (branch `fix/ghost-table-drop-atomicity`, `unitTests/resources/dropTableGhost.test.js`): they fail on the current binding with the incident error (`Could not access column family N`) and pass with this fix.

## Coordination

Sibling backport PR targets `v1.4.x` for a 1.4.3 patch release (harper-pro 5.0.x pins `^1.4.2`). The harper core PR depends on a release containing this fix.

Generated by an AI agent (Claude, investigation + fix); reviewed via adversarial agent review and a cross-model (Antigravity) review whose findings (columns-map data race, flush use-after-free, lock granularity) are addressed in this diff.

### DIFF
--- a/src/binding/database/database.cpp
+++ b/src/binding/database/database.cpp
@@ -205,11 +205,16 @@ napi_value Database::Columns(napi_env env, napi_callback_info info) {
 	NAPI_METHOD();
 	UNWRAP_DB_HANDLE_AND_OPEN();
 
-	const auto& columns = (*dbHandle)->descriptor->columns;
 	std::vector<std::string> columnNames;
-	columnNames.reserve(columns.size());
-	for (const auto& [name, _column] : columns) {
-		columnNames.push_back(name);
+	{
+		// Snapshot under the columns mutex; a concurrent drop on another
+		// thread erases from this map.
+		std::lock_guard<std::mutex> columnsLock((*dbHandle)->descriptor->columnsMutex);
+		const auto& columns = (*dbHandle)->descriptor->columns;
+		columnNames.reserve(columns.size());
+		for (const auto& [name, _column] : columns) {
+			columnNames.push_back(name);
+		}
 	}
 	std::sort(columnNames.begin(), columnNames.end());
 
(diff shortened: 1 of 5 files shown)


kriszyp (approved):

cb1kenobi (approved):
LGTM! Thanks!

PR #648, opened 2026-06-11
Backport fix/evict-dropped-cf-on-reopen to v1.4.x (for 1.4.3)

sleekmountaincat:
## Summary

Conflict-free cherry-pick of #647 onto the `v1.4.x` line: free a column family name immediately on drop so same-name recreates get a fresh column family instead of the dangling dropped handle. Fixes Harper's ghost-table failures on the 5.0.x fleet (harper-pro pins `^1.4.2`).

Intended to ship as **1.4.3** so existing 5.0.x deployments can pick it up via a harper-pro patch release, without waiting for 5.1 / the 2.x binding line.

## Validation (this exact branch)

- Built in-container against live **harper-pro 5.0.27** (linux x64 glibc, node 24.16, RocksDB 11.1.1 prebuilt) and swapped into the running server on a 2-node Fabric repro cluster: drop + same-name recreate + insert works, rapid drop/recreate cycles work, no env poisoning, no wedged creates, clean drops stay dropped across restart.
- Drop test suite passes locally against this branch's build.
- See #647 for the full description, the semantics-change discussion, and the complete suite run.

Note: the release itself is tag-driven; no package.json version bump included here, the release owner tags v1.4.3.

Generated by an AI agent (Claude). 

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #649, opened 2026-06-11
Backport fix/evict-dropped-cf-on-reopen to v1.5.x (for 1.5.1)

sleekmountaincat:
## Summary

Conflict-free cherry-pick of #647 onto `v1.5.x`: free a column family name immediately on drop so same-name recreates get a fresh column family instead of the dangling dropped handle. Fixes Harper's ghost-table failures.

Replaces #648 per @cb1kenobi: no further 1.4 releases now that 1.5 is out. harper-pro 5.0.x pins `^1.4.2`, which a **1.5.1** release satisfies, so the delivery path to existing 5.0.x deployments is unchanged (harper-pro patch release resolves it at build).

## Validation

- Cherry-pick applied with zero conflicts (same call sites as main and 1.4.x).
- Local build clean on this line; drop suite passes (6/6) including the new multi-handle regression test.
- The identical change was e2e-validated against live harper-pro 5.0.27 on a 2-node Fabric repro cluster via the 1.4.2 backport build (see #647 and #648 for details): drop + same-name recreate + insert works, no env poisoning, no wedged creates, clean drops stay dropped across restart.

See #647 for the full root-cause description and the drop-semantics discussion.

Note: release is tag-driven; no package.json bump here, release owner tags v1.5.1.

Generated by an AI agent (Claude).

### DIFF
--- a/src/binding/database/database.cpp
+++ b/src/binding/database/database.cpp
@@ -205,11 +205,16 @@ napi_value Database::Columns(napi_env env, napi_callback_info info) {
 	NAPI_METHOD();
 	UNWRAP_DB_HANDLE_AND_OPEN();
 
-	const auto& columns = (*dbHandle)->descriptor->columns;
 	std::vector<std::string> columnNames;
-	columnNames.reserve(columns.size());
-	for (const auto& [name, _column] : columns) {
-		columnNames.push_back(name);
+	{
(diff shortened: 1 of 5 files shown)


cb1kenobi (approved):
LGTM!

PR #656, opened 2026-06-16
Make column family drop idempotent

kriszyp:
## Summary
`Database::Drop` / `Database::DropSync` now treat RocksDB's `Status::InvalidArgument("Column family already dropped!")` as success instead of surfacing it. The by-name registry `unregisterColumnFamily` runs only when *this* handle actually performed the drop (`status.ok()`).

## Purpose
Harper opens one handle to a given column family **per worker thread**, and a `drop` is broadcast to every thread. When two handles to the same shared column family race to drop it, the first succeeds and the second hits RocksDB's "already dropped" error — surfacing as a hard `Drop failed`. The family being gone is the intended end state, so the drop should be idempotent. This is the root cause of the flaky `Drop number table` integration failure (paired harper PR below).

## Where to look
- `isColumnFamilyAlreadyDropped()` — the match is scoped to `status.IsInvalidArgument()` **and** the message `"Column family already dropped"`, so genuine `Drop failed: <other>` errors still surface.
- **Unregister only on `status.ok()`** (`Drop` and `DropSync`): on the tolerated already-dropped path another handle already dropped this family and owns the unregister, and the name may by then point to a *freshly-created* same-name column family — unregistering it there would corrupt the registry. (This guard came out of cross-model review.)

## Tests
3 new cases in `test/drop.test.ts`: sync + async redundant-drop is a no-op, and a stale handle's tolerated re-drop does not unregister a freshly-recreated same-name family.

---
Generated by an LLM (Claude Opus 4.8). Reviewed locally: `pnpm check` + full `drop.test.ts` (9/9) green.


src/binding/database/database.cpp
```diff
@@ -418,6 +418,18 @@ napi_value Database::Destroy(napi_env env, napi_callback_info info) {
 	NAPI_RETURN_UNDEFINED();
 }
 
+namespace {
```
cb1kenobi: Why the `namespace`? I don't think we need it.

cb1kenobi (changes requested):

kriszyp:
Good catch! Switched to `static` to match the existing `doClear`/`doClearSync` convention in this file (f6982c0b).

*— Claude Sonnet 4.6*

cb1kenobi (approved):

PR #660, opened 2026-06-17
Bound transaction-log reader limit by mapped buffer length

### LINKED ISSUE #623: transaction-log-reader: bound `limit` by `logBuffer!.length` to prevent silent truncation when `size` exceeds physical buffer
## Summary

In `src/transaction-log-reader.ts`, the `limit` variable used to bound entry reads is currently set as:

```ts
const limit = readUncommitted ? logBuffer!.length : size;
```

If `size` is somehow larger than `logBuffer!.length` (e.g. due to file truncation or an inconsistency in how committed size is reported), setting `limit = size` allows `position` to advance past the end of the physical buffer. When this happens, `logBuffer!.subarray(entryStart, position)` silently returns a truncated buffer instead of throwing a `RangeError`, producing a misframed/garbage transaction that gets processed without error.

## Suggested Fix

Bound `limit` by the physical buffer length in all cases:

```ts
const limit = readUncommitted ? logBuffer!.length : Math.min(size, logBuffer!.length);
```

This ensures that even when `size` is over-reported, `position` can never legally exceed `logBuffer!.length`, and any overrun will be caught by the existing `RangeError` guard rather than silently proceeding with a truncated subarray.

## Context

Raised as a code-review comment by `gemini-code-assist` on PR #622 (backport of #612 — txnlog reader hardening). The PR already adds robust corruption detection; this is a follow-up hardening step to close the remaining gap where committed `size` could be inconsistent with the mapped buffer.

Ref: https://github.com/HarperFast/rocksdb-js/pull/622#discussion_r3350511472

cb1kenobi:
## Summary

Bounds the committed-read `limit` in the transaction-log reader by the physical buffer length to close a silent-truncation gap. Fixes #623.

## Purpose

In [src/transaction-log-reader.ts](src/transaction-log-reader.ts#L181), the committed-read bound was `limit = size` (the committed position). If committed `size` over-reports the physically mapped buffer — e.g. a truncated or torn log file — `position` could advance past the buffer end, and `logBuffer.subarray(entryStart, position)` silently returns a truncated, misframed transaction instead of throwing. The fix clamps the bound to `Math.min(size, logBuffer.length)` so any overrun trips the existing `RangeError` guard.

## Where to look

- **The clamp is a no-op in normal operation.** The native mmap (`getMemoryMap` in [transaction_log_file_posix.cpp](src/binding/transaction_log/transaction_log_file_posix.cpp#L133)) sizes the mapped region to `fileSize` and zero-pads beyond the committed extent, so `logBuffer.length >= size` normally. The clamp only engages when the buffer is physically shorter than the reported `size` (the corruption case), so there is no regression risk for legitimate reads.
- The regression test maps a buffer truncated mid-final-entry while committed `size` still reports the full extent; verified to **fail without the fix** (silent truncation) and **pass with it** (bounded `RangeError`).

## Review notes

- Codex cross-model review: no actionable issues.
- The Gemini (`agy`) reviewer was non-functional in this environment (returned empty output even for a trivial prompt), so the second cross-model review could not be completed. Flagging for the human reviewer.

Generated by an LLM (Claude Opus 4.8).

### DIFF
--- a/src/transaction-log-reader.ts
+++ b/src/transaction-log-reader.ts
@@ -173,12 +173,16 @@ Object.defineProperty(TransactionLog.prototype, 'query', {
 
 					// Corruption bound: a committed read can't legitimately extend past
 					// the committed `size` (a true entry boundary); an uncommitted read is
-					// bounded only by the physically mapped buffer. A torn/corrupt entry
+					// bounded only by the physically mapped buffer. In both cases the bound
+					// is also clamped to `logBuffer.length` — if committed `size` over-reports
+					// the mapped buffer (e.g. a truncated/torn file), an unclamped `size`
+					// would let `position` advance past the buffer and `subarray` silently
+					// return a truncated frame instead of throwing. A torn/corrupt entry
 					// can declare a length far past this bound — without the checks below,
 					// `position` runs past the buffer, `subarray` hands back a misframed
 					// (garbage) transaction, and the advance-to-next-log path can
 					// dereference an undefined buffer. Fail loudly with a bounded error.
-					const limit = readUncommitted ? logBuffer!.length : size;
+					const limit = readUncommitted ? logBuffer!.length : Math.min(size, logBuffer!.length);
 					if (position + TRANSACTION_LOG_ENTRY_HEADER_SIZE > limit) {
 						throw new RangeError(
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Correct bounds guard — clean fix, well-tested. LGTM! 🟢

*Reviewed by Claude Sonnet 4.6*

PR #661, opened 2026-06-18
Add opt-in entry counts to purgeLogs()

cb1kenobi:
## Summary

`purgeLogs()` returns the paths of the transaction log files it deleted but gives no indication of how many entries each file held. This adds an **opt-in** `includeEntryCounts` option:

- **omitted / `false`** (default): return shape is unchanged — `string[]` of deleted file paths.
- **`true`**: returns `Array<{ path: string; entries: number }>`, where `entries` is the number of v1 entry frames the deleted file held, counted in C++.

Opt-in by design: the count requires reading each file before it is removed, so callers that don't need it pay nothing, and existing `string[]` consumers are not broken.

## Purpose

Callers purging logs want to know how much was discarded (entry counts), not just which files. Computing the count in C++ avoids re-reading/parsing the files in JS.

## Where to look

- **`src/binding/transaction_log/transaction_log_recovery.{h,cpp}`** — new pure `countTransactionLogEntries()` that walks the v1 framing (timestamp/length/flags), stopping at the first zero-timestamp marker, EOF, or a broken/torn frame — i.e. the same entry set `parseTransactionLog()` reports. Unit-tested standalone (GTest).
- **`transaction_log_file.{h,cpp}`** — `TransactionLogFile::countEntries()` reads the file's on-disk image through a *fresh* handle (old purgeable files are never opened; POSIX `pread` / Windows `FILE_SHARE_READ` make a concurrent read safe) and is wrapped to **never throw** (incl. `std::bad_alloc` from the whole-file buffer) so a counting failure can't abort a purge.
- **`transaction_log_store.{h,cpp}`** — the purge visitor signature gains an `entryCount`, and a `countEntries` flag is threaded through `purge()`/`doPurge()`. Counting happens *before* `removeFile()` and only when both a visitor and the flag are present. The no-visitor `close()`/on-load retention purge paths are unaffected (signature change is compiler-enforced across all three callers).
- **`transaction_log_store_registry.cpp`** — reads the `includeEntryCounts` option and builds either string elements or `{ path, entries }` objects.
- **`src/load-binding.ts`, `src/database.ts`** — `PurgeLogsOptions.includeEntryCounts`, a `PurgedLog` type, and overloads so the return type narrows to `string[]` or `PurgedLog[]` (plus a general overload so a non-literal `boolean` still type-checks).

## Notes for the reviewer

- **Entry semantics:** `entries` counts framed records (one per `addEntry`), matching `parseTransactionLog().entries.length` — a multi-entry transaction contributes each of its frames. Confirm that's the count you'd expect.
- **Cross-model review:** Codex reviewed and surfaced two items (TS overload gap for a `boolean` variable; `std::bad_alloc` could escape `countEntries`) — both fixed in this PR. The Gemini leg could not run in my environment (the `agy` CLI returns empty headlessly and `gemini` isn't installed), so a second cross-model pass is still worth a human eye.

## Testing

- Native GTest: 6 new `TransactionLogCount` cases (header-only, clean, zero-pad tail, torn tail, partial header, large entry).
- Vitest: 3 new `purgeLogs` cases (entry counts when opted in, zero for header-only, unchanged `string[]` by default). Full suite green (548 passed).

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: Claude Opus 4.8)

### DIFF
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -1,4 +1,6 @@
 #include <cmath>
+#include <fstream>
+#include <limits>
 #include <sstream>
 #include <system_error>
 #include <vector>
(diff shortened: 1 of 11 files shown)


kriszyp (approved):
Zero default-path cost, clean opt-in design — entry count only scans on explicit request. LGTM! 🟢

*Reviewed by Claude Sonnet 4.6*

PR #664, opened 2026-06-18
Fix cross-thread use-after-free in DBRegistry::CloseDB

cb1kenobi:
## Summary
Fixes an intermittent heap-corruption crash in the benchmark suite (`malloc(): unaligned tcache chunk detected` on Linux/glibc).

**Root cause:** worker threads that open the same DB path share one process-global `DBDescriptor`. On env teardown each worker's `Database` finalizer calls `DBRegistry::CloseDB`. The old code cached a raw pointer into the registry map under `databasesMutex`, then dereferenced it *after releasing the lock* — so a concurrent `CloseDB` for the same path could erase the map node first, and the survivor called `close()` on a freed `DBDescriptor`, locking its destroyed mutex. glibc only detects the resulting corruption later, as the tcache abort.

**Fix:** `CloseDB` now makes the purge decision entirely under `databasesMutex` with no raw map pointer escaping the lock. Purge is gated on `use_count() <= 1` (only the registry ref remains) **and** `DBDescriptor::beginClose()` — an atomic claim split out of `close()` that also publishes the closing state while the lock is held. `close()` is now `beginClose()` + `finishClose()`.

## Why this is correct (please scrutinize)
- **open-vs-close serialization:** `OpenDB` holds `databasesMutex` for its body and bumps the descriptor's `use_count` under it. So our `use_count() <= 1` check serializes with opens: if an open raced ahead, the count is `> 1` and we skip; if we win, `beginClose()` publishes `isClosing()` under the same lock, so a subsequent `OpenDB` waits instead of being handed a descriptor we then close. The entry stays in the map (non-null, `isClosing()`) until `finishClose()` completes.
- **single-purge:** two concurrent `CloseDB` for the same path can't both purge — `beginClose()` is an atomic test-and-set, and the under-lock `shared_ptr` copy also pushes `use_count` past the threshold for the loser.

## Where to look
- `src/binding/database/db_registry.cpp` — the rewritten `CloseDB`.
- `src/binding/database/db_descriptor.{h,cpp}` — the `close()` → `beginClose()`/`finishClose()` split; confirm the other `close()` callers (`DestroyDB`, `Shutdown`, descriptor destructor) are unaffected (they still go through `close()`, which is idempotent).

## Verification
- Reproduced locally under macOS **Guard Malloc** (AddressSanitizer is unusable on this OS — it deadlocks at init): a tight shared-descriptor teardown repro crashed in round 0 with `mutex lock failed: Invalid argument`; `lldb` gave the exact freed-mutex `close()` stack. After the fix, 90 rounds clean.
- New regression test `test/concurrent-teardown.test.ts` (many same-path workers tearing down concurrently).
- Full suite green (537 passed) before and after; `tsc`/lint/fmt clean.

## Also included (ASan tooling)
- `ROCKSDB_ASAN=1` build toggle in `binding.gyp` and a manual-dispatch `benchmark-asan.yml` (Linux, `LD_PRELOAD` libasan) to catch this class of corruption on the self-hosted benchmark runner.
- `scripts/gmalloc-worker-repro.mjs` + AGENTS.md notes on the teardown invariant and the Guard-Malloc-on-macOS debugging path.

## Review status / open items
- **Codex** review completed; its only finding (the open-vs-close TOCTOU, P1) is addressed by publishing `closing` under the lock via `beginClose()`.
- **Gemini/agy** review could not be run in the headless agent environment (autonomous mode blocked, no print output, raw `gemini` unavailable) — a second human/model pass on the concurrency reasoning would be valuable.

🤖 Generated by Claude Code (Opus 4.8, 1M context).


kriszyp (approved):
Approving — this is a clean fix for a real cross-thread use-after-free in `DBRegistry::Close`. The teardown is correctly performed under lock, the object-lifetime race is genuinely closed (not just narrowed), and the repro test gives confidence it won't regress. No hot-path cost — it's all on the teardown path. Nice catch tracking this one down. 🙏

— Reviewed by Claude (claude-opus-4-8)

PR #665, opened 2026-06-18
Fix worker-env teardown use-after-free in per-database event notify

kriszyp:
## Summary

- `EventEmitter::notify` was snapshotting listener tsfns under `events.mutex`, releasing the lock, then calling `napi_call_threadsafe_function` outside the lock
- This opened a window where a dying worker env could free a tsfn between the snapshot and the call → SIGABRT in `napi_call_threadsafe_function` (HarperFast/harper#1370)
- Fix: call the tsfns **while still holding** `events.mutex`; every tsfn release path also takes that mutex, so env teardown cannot free a tsfn until the call loop returns

## Why the fix is safe

`napi_call_threadsafe_function` only enqueues to the owning env's event loop — the JS trampoline runs later on that env's thread, never re-entering the mutex. The tsfns are created with an unbounded queue (`max_queue_size=0`), so the call never blocks. No tsfn callback can re-enter the emitter mutex, and no code holds the emitter mutex and then acquires another mutex in the codebase — no deadlock potential.

The `napi_acquire_threadsafe_function` approach (prior code) does **not** close this: Node frees a worker env's tsfns regardless of the tsfn-level thread count when the env exits.

## Test

Regression test (`test/notify-teardown-uaf.test.ts`) spawns the fixture in a child process with a read-once env-gated seam (`ROCKSDB_JS_NOTIFY_DELAY_MS=25`) that widens the in-lock window to surface the race deterministically. The fixture runs a committer worker (streams async transaction commits) against 60 churning listener workers (register one committed listener each, then get `terminate()`d). No transactions in listener workers — this isolates the notify UAF from the unrelated close-vs-commit double-free. Verified: 0/32 crashes with fix, 2/2 without.

## Review notes

The `napi_tsfn_blocking` call mode is effectively non-blocking here given the unbounded queue — equivalent to `napi_tsfn_nonblocking` in this configuration. Open to switching to `napi_tsfn_nonblocking` for clarity, but it makes no practical difference.

*Generated by Claude Opus 4.8 (1M context)*

### DIFF
--- a/src/binding/core/test_seam.h
+++ b/src/binding/core/test_seam.h
@@ -4,20 +4,14 @@
 #include <cstdlib>
 
 // Deterministic test seams that widen a race window are gated on a millisecond
-// delay read once from an environment variable (0 = disabled). They are inert in
+// delay read from an environment variable (0 = disabled). They are inert in
 // production where the env var is unset.
 //
-// DEFINE_TEST_DELAY_MS(funcName, envName) defines a `static int funcName()` that
-// returns that delay, reading `envName` exactly once into a function-local static.
-// Each seam keeps its own named accessor (and single cached read) at the call site;
-// see EventEmitter::notify and TransactionHandle::close for usage.
-#define DEFINE_TEST_DELAY_MS(funcName, envName) \
-	static int funcName() { \
-		static const int delayMs = []() -> int { \
-			const char* value = ::getenv(envName); \
-			return value ? ::atoi(value) : 0; \
-		}(); \
-		return delayMs; \
-	}
+// Pass the env var name to testDelayMs() at the call site; see
+// EventEmitter::notify and TransactionHandle::close for usage.
+inline int testDelayMs(const char* envName) {
+	const char* value = ::getenv(envName);
+	return value ? ::atoi(value) : 0;
+}
 
 #endif
--- a/src/binding/core/test_seam.h
+++ b/src/binding/core/test_seam.h
@@ -1,0 +1,23 @@
+#ifndef __CORE_TEST_SEAM_H__
+#define __CORE_TEST_SEAM_H__
+
+#include <cstdlib>
+
+// Deterministic test seams that widen a race window are gated on a millisecond
+// delay read once from an environment variable (0 = disabled). They are inert in
+// production where the env var is unset.
+//
+// DEFINE_TEST_DELAY_MS(funcName, envName) defines a `static int funcName()` that
+// returns that delay, reading `envName` exactly once into a function-local static.
+// Each seam keeps its own named accessor (and single cached read) at the call site;
+// see EventEmitter::notify and TransactionHandle::close for usage.
+#define DEFINE_TEST_DELAY_MS(funcName, envName) \
+	static int funcName() { \
+		static const int delayMs = []() -> int { \
+			const char* value = ::getenv(envName); \
+			return value ? ::atoi(value) : 0; \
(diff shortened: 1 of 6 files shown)


src/binding/napi/event_emitter.cpp
```diff
@@ -1,11 +1,25 @@
...
+// Returns the configured artificial delay in ms (0 = disabled), read once from
+// ROCKSDB_JS_NOTIFY_DELAY_MS. Unset in production; see EventEmitter::notify.
+static int notifyTestDelayMs() {
+	static const int delayMs = [] {
+		const char* value = ::getenv("ROCKSDB_JS_NOTIFY_DELAY_MS");
+		return value ? ::atoi(value) : 0;
+	}();
+	return delayMs;
+}
+
```
cb1kenobi: This looks identical to `txnCloseTestDelayMs()` (https://github.com/HarperFast/rocksdb-js/blob/2d544a1074780d121eb2b3d774a7f8a0696b2136/src/binding/transaction/transaction_handle.cpp#L21-L27). I'd say move this function to a file in `core` and create a preprocessor macro that accepts the env variable name and returns the value.

cb1kenobi (changes requested):

src/binding/napi/event_emitter.cpp
kriszyp: Good call — done in 6065a802. Both seams now share a `DEFINE_TEST_DELAY_MS(funcName, envName)` macro in the new `core/test_seam.h`. I went with a function-*generating* macro (rather than an expression macro) so each seam keeps its own named accessor and single read-once static — call sites are unchanged. Both regression tests still repro/pass.

— KrAIs (Claude Opus 4.8)

src/binding/core/test_seam.h
```diff
@@ -0,0 +1,23 @@
...
+// Each seam keeps its own named accessor (and single cached read) at the call site;
+// see EventEmitter::notify and TransactionHandle::close for usage.
+#define DEFINE_TEST_DELAY_MS(funcName, envName) \
+	static int funcName() { \
+		static const int delayMs = []() -> int { \
+			const char* value = ::getenv(envName); \
+			return value ? ::atoi(value) : 0; \
+		}(); \
+		return delayMs; \
+	}
```
cb1kenobi: Do we really want to define this function multiple times just because the env var name is different? I would think we define the function as a utility function once, then call it as needed. No need for preprocessor macros.

```suggestion
int testDelayMs() {
	static const int delayMs = []() -> int {
		const char* value = ::getenv(envName);
		return value ? ::atoi(value) : 0;
	}();
	return delayMs;
}
```

Then to use it, call `testDelayMs("ROCKSDB_JS_NOTIFY_DELAY_MS")` or `testDelayMs("ROCKSDB_JS_TXN_CLOSE_DELAY_MS")`.

cb1kenobi (changes requested):

src/binding/core/test_seam.h
kriszyp: Done in 21ed21cc — replaced the macro with a plain `inline int testDelayMs(const char* envName)` in `test_seam.h`. Both call sites pass their env var name directly. Dropped caching since `getenv` overhead on a null-returning production env var is negligible for a test seam.

— KrAIs (Claude Opus 4.8)

cb1kenobi (approved):
Awesome, great find!

PR #666, opened 2026-06-18
Fix TransactionHandle::close() double-free when descriptor close races async-commit callback (harper#1370)

kriszyp:
## Summary

Fixes the close-vs-commit double-free reported in HarperFast/harper#1370. When a DB descriptor is closed by one env's JS thread (PATH A: `DBDescriptor::close()` → `T->close()`) while the same transaction's async commit complete callback fires on another env's JS thread (PATH B), both paths could pass the original `!this->txn` guard and both execute `delete this->txn`, causing a double-free through `~OptimisticTransaction → ~PointLockTracker → SIGABRT`.

**Two fixes:**

1. **One-shot `closed` gate** (`std::atomic<bool> closed{false}`): `closed.exchange(true)` at the start of `close()`. The first caller proceeds; the second returns immediately. Prevents both PATH A and PATH B from reaching `delete this->txn`.

2. **Cross-env `napi_delete_reference` guard** (`std::thread::id envThreadId`): stored at construction time (W's JS thread). When `close()` is invoked from a different env's JS thread (PATH A, M's thread), `napi_delete_reference` is skipped — calling it cross-env is UB and crashes. The ref is a weak ref (refcount=0) and is cleaned up by Node on env teardown.

Also adds a `ROCKSDB_JS_TXN_CLOSE_DELAY_MS` test seam (read-once static, zero-cost in production) that widens the PATH A / PATH B overlap window after `waitForAsyncWorkCompletion()` returns, and a deterministic 60-round regression test.

## Relationship to PR #664

No file overlap with PR #664 (`DBRegistry::CloseDB` raw-pointer race). Different bugs, different layers. The regression test's round-coordination design avoids the DBRegistry race (W only opens its handle after M already holds one), so both bugs can coexist without interfering with each other.

## Areas deserving attention

- **`envThreadId` guard**: the transaction is constructed on W's JS thread, so `envThreadId = W's thread ID`. PATH A runs on M's JS thread. The IDs differ → `napi_delete_reference` is correctly skipped when PATH A wins the gate. This path is exercised deterministically by the test (whenever PATH A wins).
- **`waitForAsyncWorkCompletion()` unblocking point**: unblocks when `signalExecuteCompleted()` is called at the end of the execute lambda — before the complete callback fires. The production window between these two events is real; the seam just widens it.

Cross-model review (quick, Gemini leg unavailable): no blockers found. Significant concerns were all correctness confirmations; two minor suggestions addressed (test seam comment clarified, `txnCloseTestDelayMs()` read once per call site).

*Generated by Claude Opus 4.8*

### DIFF
--- a/src/binding/transaction/transaction_handle.cpp
+++ b/src/binding/transaction/transaction_handle.cpp
@@ -210,9 +210,12 @@ void TransactionHandle::close() {
 	this->waitForAsyncWorkCompletion();
 
 	// Test seam: widen the PATH A vs PATH B race window (see txnCloseTestDelayMs).
+	// This window is real in production (PATH B fires after waitForAsyncWorkCompletion
+	// unblocks); the seam makes it wide enough to reproduce deterministically.
 	// Noop in production.
-	if (txnCloseTestDelayMs() > 0) {
-		std::this_thread::sleep_for(std::chrono::milliseconds(txnCloseTestDelayMs()));
+	const int testDelayMs = txnCloseTestDelayMs();
+	if (testDelayMs > 0) {
+		std::this_thread::sleep_for(std::chrono::milliseconds(testDelayMs));
 	}
 
 	// if the transaction was aborted (either via an error, explicit abort, or was pending), we need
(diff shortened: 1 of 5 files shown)


cb1kenobi (approved):
Tested and looks awesome!

PR #673, opened 2026-06-23
Guard against null transaction handle in Commit complete callback

cb1kenobi:
## Summary

While reviewing `Transaction::Commit`, we noticed the complete callback was inconsistent about guarding `state->handle`. The ok path and the IsBusy path already null-checked it, but:

- the **normal error path** dereferenced `state->handle->state` unconditionally
- two `DEBUG_LOG` calls dereferenced `state->handle->id` without a guard — including the `else` branch whose entire purpose is to handle the null case (it would have null-deref'd the moment it ran)

`state->handle` is **never actually null in practice** today — it's a strong `shared_ptr<TransactionHandle>` copy, constructed non-null (`UNWRAP_TRANSACTION_HANDLE` guarantees it) and never reassigned/reset, so the strong ref keeps the object alive for the whole async-work lifetime. These guards are therefore defensive: the goal is that a null handle should never be able to crash the process, rather than asserting on it.

## Changes

- `transaction.cpp`: guard the error-path `state->handle->state` check (matches the existing ok/IsBusy guards)
- `transaction.cpp`: make the entry-log `txnId` null-safe
- `transaction.cpp`: drop the null-deref `->id` from the "handle is null!" log
- `transaction_handle.cpp`: fix the field-init-order warning (`envThreadId` / `committedPosition`) surfaced by the debug build

## Impact

The `DEBUG_LOG` changes only affect debug builds (`DEBUG_LOG` is a no-op in production), so this is not a production crash fix — it makes the whole complete callback uniformly null-tolerant and silences a compiler warning.

## Testing

- `pnpm build:binding:debug` compiles cleanly (no warnings in the touched files)

### DIFF
--- a/src/binding/transaction/transaction.cpp
+++ b/src/binding/transaction/transaction.cpp
@@ -326,7 +326,7 @@ napi_value Transaction::Commit(napi_env env, napi_callback_info info) {
 			TransactionCommitState* state = reinterpret_cast<TransactionCommitState*>(data);
 
 			DEBUG_LOG("%p Transaction::Commit Complete callback entered (status=%d, txnId=%d)\n",
-				state->handle.get(), status, state->handle->id);
+				state->handle.get(), status, state->handle ? state->handle->id : 0);
 
 			state->deleteAsyncWork();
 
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Traced every branch of the commit-complete callback at head — the null-handle guard is complete (no surviving unguarded `state->handle` deref), resolve/reject is settled exactly once on each path, and there's zero production-path cost. Nice tight crash-safety fix. — Claude (Opus 4.8)

Issue #676, opened 2026-06-24
Optional: bound exactStart absent-timestamp lookups with a suffix-minimum index


kriszyp:
## Summary

`TransactionLog.prototype.query({ start, exactStart: true })` has **no upper bound** on its
forward scan. When the requested `start` timestamp is **absent** from the log, the iterator
scans from the computed start position all the way to **end-of-log** (continuing across log
files) before returning `done`, even though it can never match. On a large or multi-segment
log this turns a point lookup into an O(n) scan (multi-millisecond).

This is a **missing capability, not a correctness bug** — results are always correct. Filing as
a tracked, optional enhancement.

## Background — why the current index can't bound it

`findPositionByTimestamp` (`src/binding/transaction_log/transaction_log_file.cpp`) builds a
**running-maxima** index: it records an entry only when its timestamp exceeds the current max,
then answers with `lower_bound(t)`. That gives a correct, tight **lower bound to START a scan**
— "before the returned position, no entry has timestamp ≥ t" — which is exactly what
`query()`/`next()` relies on, in both normal and `exactStart` modes. It does **not** provide any
*upper* bound, so the scan has no stop condition and runs to EOF.

Position-derived stop heuristics (e.g. "stop at the next maxima position", "stop at
`nextUp(start)`") are **not safe**: on an out-of-order log a present `=== start` entry can sit
physically *after* such a position, so those bounds produce false negatives and break the
existing `should query an out-of-order transaction log` and
`keeps every committed entry findable … (#1148)` (HarperFast/harper#1148) tests.

## Proposed option — a suffix-minimum ("min remaining timestamp") structure

Maintain, alongside the maxima index, the minimum timestamp over the *remaining* (suffix)
entries. A forward scan can then **terminate as soon as `min(remaining) > start`** — at that
point no `=== start` entry can remain. (The same stop generalizes range queries: a `[start,
end)` scan can stop once `min(remaining) >= end`.)

- **(a) Correctness under arbitrary disorder:** safe. The stop is value-based, not
  position-based, so an out-of-order match always keeps `min(remaining) <= start` until it is
  read — no false negatives. Keeps the out-of-order and #1148 tests green (for a present `t`,
  at `t`'s position `min(remaining) <= t`, so it cannot stop early).
- **(b) Tightness:** tight for mostly-increasing logs (the min sits near the front, so
  `suffix-min` rises immediately past `start`). **Adversarial worst case:** a single small
[...]

PR #677, opened 2026-06-24
Correct findPositionByTimestamp docstring to match running-maxima semantics

kriszyp:
## Summary

Doc-only fix. Corrects the `findPositionByTimestamp` function docstring
(`src/binding/transaction_log/transaction_log_file.cpp`) to match what the code actually
computes. No behavior change.

## Why

The old docstring — *"the position where there are no transactions with an earlier timestamp"* —
reads like a suffix guarantee. The function actually builds a **running-maxima** index answered
with `lower_bound`: it returns the first entry with timestamp `>= t` such that no entry *before*
it is `>= t` — i.e. a tight **lower bound to START a forward scan**, with no upper bound. The
N-API wrapper docstring (`transaction_log.cpp`) and the inline comment in this same function
already state the correct semantics; only the function-level docstring was inverted.

## Notes

- The "no upper bound" property (an absent-timestamp `exactStart` query scans to end-of-log) is
  a real but separate, optional perf enhancement — tracked in #676, not addressed here.
- Trivial doc change; no tests/behavior affected.

_Generated with assistance from Claude (Opus 4.8)._

### DIFF
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -385,10 +385,16 @@ std::shared_ptr<MemoryMap> TransactionLogFile::getMemoryMap(uint32_t fileSize, b
 }
 
 /**
- * Find the position of a timestamp, specifically the position where there are no transactions with an earlier timestamp
- * @param timestamp - the timestamp to find the position of
+ * Find the start position for a forward scan of all transactions with timestamp >= the given one:
+ * the position of the first entry whose timestamp is >= `timestamp`, such that no entry BEFORE it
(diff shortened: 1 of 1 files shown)


kriszyp:
Superseded: folding this docstring correction into the actual suffix-minimum early-termination implementation (which closes #676). The implementation carries the corrected docstring by construction, so a standalone doc PR is redundant. Closing in favor of the implementation PR.

kriszyp:
Reopened: after evaluating the suffix-min implementation, we chose to ship this docstring correctness fix on its own and track the (now-deferred) early-termination optimization in #676. My earlier "superseded" note no longer applies — this is the deliverable. Scope is doc-only; no behavior change.

cb1kenobi (approved):

PR #687, opened 2026-07-06
Stream backups to a WritableStream as a tar archive

cb1kenobi:
## Summary
- Adds a streaming backup mode: `db.backup(stream)` writes a consistent snapshot of the whole database to a WHATWG `WritableStream` as an (optionally gzipped) tar archive, with no intermediate copy on disk. `db.backup()` still accepts a directory string for the existing directory-backup path.
- Replaces the backup-directory lock's pidfile scheme with a kernel advisory lock (`flock`/`LockFileEx`) held entirely in native code, fixing container pid-namespace staleness issues and a Windows fd/CRT mismatch found while implementing it.

## Purpose
Directory backups require a full on-disk copy before the caller can move the data elsewhere (e.g. upload to object storage). Streaming backups let a caller pipe a live snapshot directly to its destination. Building this surfaced a pre-existing gap — concurrent backup operations on the same directory could race and corrupt the backup set — addressed here with the directory lock.

## Where to focus review
- `src/binding/database/backup_stream.cpp` — the native streaming path: file enumeration under `DisableFileDeletions`, worker-thread reads, and the condvar-based backpressure handshake with the JS-side threadsafe function. The lifetime/refcount handling (async-work completion, tsfn thread-finalize, in-flight promise continuations) and the `isClosing()`/`isCancelled()` polling during backpressure waits are the trickiest part of this PR — worth a close look for teardown races.
- `src/binding/core/file_lock.cpp` — the advisory-lock implementation. Note the design constraint documented in AGENTS.md: no file descriptor/handle crosses the JS boundary (the addon statically links its own CRT, so a Node fd can't be resolved via `_get_osfhandle` on Windows); native owns open/lock/close and only returns an opaque token.
- `src/tar.ts` — dependency-free streaming USTAR encoder (GNU base-256 fallback for >8 GiB entries), validated against the real `tar` package in tests.
- Teardown-race paths were verified under Guard Malloc locally (see AGENTS.md's "Debugging native heap corruption" section) as well as the full test suite.

Generated by Claude (Sonnet 5).


cb1kenobi:
Addressed both findings from the review above in f714dbd7:

- **Windows path encoding** (`src/binding/core/file_lock.cpp`): now converts `backupDir` from UTF-8 to UTF-16 explicitly before constructing `std::filesystem::path`, avoiding ANSI code-page corruption of non-ASCII paths on Windows.
- **Unchecked `DisableFileDeletions` status** (`src/binding/database/backup_stream.cpp`): the stream now aborts immediately if disabling file deletions fails, instead of proceeding to enumerate/read live files that a concurrent compaction could drop.

Verified via `pnpm build:binding`, `pnpm test:native`, and the backup/backup-stream Vitest suites (all passing).

cb1kenobi:
Closing/reopening to re-trigger CI — pushes since the base-branch retarget to backup-concurrent-dir-lock stopped creating GitHub Actions check-suites (verified: zero check-suites for the last two commits after ~45 min, while pushes to backup-concurrent-dir-lock itself trigger fine).

cb1kenobi:
Update: after the base retarget to \`backup-concurrent-dir-lock\`, that branch moved forward with the \`tryFileLock\`/\`fileLockRelease\` API refactor (#684), which put this PR's history (still on the old \`backupDir\`-based API) into a real merge conflict — \`mergeable: CONFLICTING\`. That's also why GitHub Actions stopped scheduling runs entirely (no check-suite could be created without a valid merge ref); an empty-commit nudge and a close/reopen didn't help because the underlying cause wasn't a delivery glitch.

Merged \`backup-concurrent-dir-lock\` into this branch in b179f7dc: adopted the generic \`tryFileLock\`/\`fileLockRelease\` API throughout, and fixed two stale references the refactor missed along the way — a README section documenting a nonexistent \`fileLockTryAcquire\` with boolean/callback semantics (doesn't match the real token-based API), and an outdated function-name callout in AGENTS.md.

Verified post-merge: \`pnpm build:binding\`, \`pnpm test:native\` (50/50 non-skipped), \`pnpm type-check\`, and the backup/backup-stream/file-lock/tar Vitest suites (50/50) all pass. CI is now scheduling normally.

kriszyp (approved):
Reviewed via Claude review-queue — very cool! Real backpressure handshake (worker blocks on a condvar until the JS-side write() promise settles, so peak memory is one 1 MiB chunk regardless of database size), clean consumer-error abort path, and the teardown-race handling (polling isClosing()/isCancelled() instead of an unconditional wait, refcounted state so a late promise continuation lands on live memory) is exactly the kind of care this needs. Both Gemini-bot findings from earlier commits are confirmed fixed at HEAD. CI is fully green — the Windows/Bun failure noted in triage was stale, from a different branch (#684) during its merge-conflict window. LGTM, merge.

PR #693, opened 2026-07-08
VerificationTable invalidation hardening: clear() sweep, non-txn write settle, snapshot TOCTOU, backdated-version gate

kriszyp:
Four fixes to the VerificationTable invalidation contract, found during a multi-agent review of harper#410 (which layers a per-worker WeakLRUCache over the VT via `expectedVersion`/`FRESH_VERSION_FLAG`). The invariant being enforced: **every write path on a VT-enabled store must invalidate the affected slot(s)** — previously only transactional writes did.

## A — `clear()` sweeps the VT (`db_handle.cpp`, `verification_table.{h,cpp}`)

`DBHandle::clear()` (compactRange + DeleteFilesInRange) never touched the VT, so after a table clear every other worker kept verifying its cached pre-clear versions as FRESH indefinitely — deleted records served forever (also mid-HNSW-reindex `dbi.clear()`). New `VerificationTable::settleAllSlots()` advances every non-lock slot to a fresh settled generation after the delete completes. Lock slots are skipped (their write-intent lifecycle already advances them on release). The sweep is coarse — process-global, since non-lock slots don't track provenance — costing spurious cache misses on other stores; acceptable for an operation as rare as clear(). Ordering argument (delete-visible → sweep → fresh generation) closes the stale re-publication window: a populate that observed a pre-sweep slot value fails its CAS; one that observed a post-sweep value read post-delete data.

## B — non-transactional `putSync`/`removeSync` settle the slot (`database.cpp`)

The direct `db->Put`/`db->Delete` branches had zero VT interaction, so any bare write on a VT-enabled store left the old version verifying FRESH on every other worker (harper aliases `db.put = db.putSync`, making this a standing footgun). Now mirrors the transactional path exactly: `lockSlotForWrite` before the write, `releaseWriteIntent` (→ fresh settled generation) after.

## C — snapshot pinned before VT slot load (`transaction.cpp`, `database.cpp`)

In the transactional read paths the slot was loaded *before* `ensureSnapshot()`. A full write cycle (lock → commit → settle) landing in that window on a transaction's first read could return cached V_old while pinning a post-V_new snapshot — a torn view. The snapshot is now established before the slot load in both `Transaction::GetSync` and the `Database::GetSync` txnId path (which also consolidates the duplicated txn lookup). Non-transactional reads are unaffected. Not deterministically testable from JS; correctness is by ordering argument.

## D — sequence gate for backdated versions (`database.h` `vtPopulateIfSettled`)

The populate gate compared the oldest open snapshot's *wall-clock creation time* against the record's *Harper version timestamp*. Replicated/catch-up writes carry origin timestamps that can predate an open local snapshot, so the gate passed and a pinned snapshot that legitimately sees V_old could get FRESH V_new. Added a second gate, run **only when Gate 1 passes** (open snapshot + version older than it — exactly the backdated case): defer publication when `oldest-snapshot-sequence < GetLatestSequenceNumber()`.

**Known conservatism:** Gate 2 uses the DB-wide latest sequence as a proxy for the key's write sequence, so while a snapshot is open and *any* write has landed after it, population of Gate-1-passing (old-versioned) records is deferred until the snapshot drains. The no-snapshot hot path is untouched. If benchmarks show population starvation under mixed read/write load, the follow-up is a per-key sequence check (`GetLatestSequenceForKey` with memtable-window handling).

## Tests

6 new cases in `test/verification-table.test.ts` covering A (clear → stale expectedVersion falls through to DB, verifyVersion false), B (non-txn putSync/removeSync invalidate the slot), and D (backdated version + open snapshot defers population; publication resumes after snapshot release). Full suite: 45 files, 602 passed / 1 skipped.

Pairs with harper#410; harper should bump this dep once released so the clear()/cross-worker fixes land behind the record-caching feature.

### DIFF
--- a/src/binding/core/verification_table.cpp
+++ b/src/binding/core/verification_table.cpp
@@ -255,6 +255,34 @@ void VerificationTable::unrefTracker(LockTracker* t) {
 	if (t->refcount.fetch_sub(1, std::memory_order_acq_rel) == 1) delete t;
 }
 
+void VerificationTable::settleAllSlots() {
+	if (!slots_) return;
+	size_t n = mask_ + 1;
+	for (size_t i = 0; i < n; ++i) {
+		uint64_t v = slots_[i].load(std::memory_order_acquire);
+		// Skip lock slots — the write-intent lifecycle (releaseWriteIntent)
+		// advances them to a fresh settled-empty when the last holder releases.
+		// A concurrent lockSlotForWrite uses an unconditional store (under
+		// writerMutex_) that can race with our CAS; if it wins, the slot is a
+		// lock that will advance further when committed/aborted — still correct.
+		if (vtIsLock(v)) continue;
+		// CAS non-lock (version or settled-empty) to a fresh generation. On CAS
+		// failure the current value is loaded into v; if it's now a lock we
+		// stop (handled above). If it's a different non-lock value (concurrent
+		// populate or another sweep) we retry once more to ensure advancement.
+		while (!vtIsLock(v)) {
+			if (slots_[i].compare_exchange_strong(
+					v,
+					vtEncodeSettled(vtNextSettleGen()),
+					std::memory_order_release,
+					std::memory_order_acquire)) {
+				break;
+			}
+			// v has been reloaded by the failed CAS; if now a lock, exit.
+		}
+	}
+}
+
 void VerificationTable::cancelForDB(uintptr_t dbPtr) {
 	if (!slots_) return;
 	std::lock_guard<std::mutex> lock(writerMutex_);
--- a/src/binding/core/verification_table.cpp
+++ b/src/binding/core/verification_table.cpp
@@ -258,6 +258,11 @@ void VerificationTable::unrefTracker(LockTracker* t) {
 void VerificationTable::settleAllSlots() {
 	if (!slots_) return;
 	size_t n = mask_ + 1;
+	// One fresh generation shared by every slot in this sweep: the CAS is
+	// per-slot, so uniqueness only matters versus values a reader may have
+	// previously observed in that slot — a just-minted generation satisfies
+	// that for all slots, and hoisting avoids n atomic fetch-adds.
+	const uint64_t settledValue = vtEncodeSettled(vtNextSettleGen());
 	for (size_t i = 0; i < n; ++i) {
 		uint64_t v = slots_[i].load(std::memory_order_acquire);
 		// Skip lock slots — the write-intent lifecycle (releaseWriteIntent)
@@ -266,19 +271,17 @@ void VerificationTable::settleAllSlots() {
 		// writerMutex_) that can race with our CAS; if it wins, the slot is a
 		// lock that will advance further when committed/aborted — still correct.
 		if (vtIsLock(v)) continue;
-		// CAS non-lock (version or settled-empty) to a fresh generation. On CAS
-		// failure the current value is loaded into v; if it's now a lock we
-		// stop (handled above). If it's a different non-lock value (concurrent
-		// populate or another sweep) we retry once more to ensure advancement.
+		// CAS non-lock (version or settled-empty) to the sweep generation,
+		// retrying until the CAS lands or the slot turns into a lock (each
+		// failed CAS reloads v; a lock exits via the loop condition).
 		while (!vtIsLock(v)) {
 			if (slots_[i].compare_exchange_strong(
 					v,
-					vtEncodeSettled(vtNextSettleGen()),
+					settledValue,
 					std::memory_order_release,
 					std::memory_order_acquire)) {
 				break;
 			}
-			// v has been reloaded by the failed CAS; if now a lock, exit.
 		}
 	}
 }
--- a/src/binding/database/database.cpp
+++ b/src/binding/database/database.cpp
@@ -474,6 +474,14 @@ napi_value Database::Drop(napi_env env, napi_callback_info info) {
 		// owns the unregister; the name may now point to a freshly-created
 		// family, so unregistering here would corrupt the registry.
 		(*dbHandle)->descriptor->unregisterColumnFamily((*dbHandle)->getColumnFamilyName());
+		// Dropping a column family bulk-deletes its data exactly like clear();
+		// sweep the VT so pre-drop versions can no longer verify FRESH (see
+		// DBHandle::clear). Only on the ok path — on already-dropped, the
+		// handle that performed the drop owns the sweep.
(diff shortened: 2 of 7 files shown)


src/binding/database/db_handle.cpp
```diff
@@ -96,7 +98,24 @@ rocksdb::Status DBHandle::clear() {
...
+	// (covers all slots in the process-global table) since slot provenance is
+	// not tracked for non-lock slots; the cost is spurious cache misses for
+	// other stores, which is acceptable given clear() is already a rare,
+	// expensive operation. Lock slots are skipped — the write-intent lifecycle
+	// handles them independently. We sweep regardless of clearStatus: if the
+	// delete partially succeeded, any keys that were removed must not remain
+	// cacheable via stale VT entries.
+	if (this->enableVerificationTable) {
+		VerificationTable* vt = DBSettings::getInstance().getVerificationTableRaw();
+		if (vt) vt->settleAllSlots();
```
cb1kenobi, line 116: **Medium: Drop/DropSync (non-default column family) don't sweep the VT — the same stale-cache gap Fix A closes for `clear()`**

`Database::Drop` / `Database::DropSync` call `DropColumnFamily` directly (`database.cpp:459` and `:506`) with no VT interaction. A non-default-CF drop bulk-deletes the store's data exactly like `clear()`, but leaves every cached version slot for that `(descriptor, cfId, key)` untouched — so other workers keep verifying pre-drop versions as FRESH indefinitely, the identical cross-worker stale read this sweep fixes for `clear()`. (`cancelForDB` only clears lock slots on descriptor teardown, and default-CF drop routes to `clear()` so it's already covered — only the non-default-CF `DropColumnFamily` path is exposed.)

This directly contradicts the PR's stated invariant that every write path on a VT-enabled store must invalidate the affected slot(s). If a VT-enabled store maps to a non-default column family, dropping it is such a path.

Suggested fix: mirror Fix A — after a successful `DropColumnFamily` on a VT-enabled handle, sweep the VT (`getVerificationTableRaw()` → `settleAllSlots()`), the same block added here in `clear()`.

—
_Generated by Barber AI_

src/binding/core/verification_table.cpp
```diff
@@ -255,6 +255,34 @@ void VerificationTable::unrefTracker(LockTracker* t) {
...
+		// advances them to a fresh settled-empty when the last holder releases.
+		// A concurrent lockSlotForWrite uses an unconditional store (under
+		// writerMutex_) that can race with our CAS; if it wins, the slot is a
+		// lock that will advance further when committed/aborted — still correct.
+		if (vtIsLock(v)) continue;
+		// CAS non-lock (version or settled-empty) to a fresh generation. On CAS
+		// failure the current value is loaded into v; if it's now a lock we
+		// stop (handled above). If it's a different non-lock value (concurrent
+		// populate or another sweep) we retry once more to ensure advancement.
+		while (!vtIsLock(v)) {
```
cb1kenobi, line 277: **Nit: comment says "retry once more" but this is an unbounded retry loop**

The comment above ("we retry once more to ensure advancement") describes a single retry, but the `while (!vtIsLock(v))` loop retries until the CAS lands or the slot turns into a lock. The code is correct; the comment mischaracterizes the control flow — worth fixing since this is subtle concurrency code a future reader will trust the comment over.

—
_Generated by Barber AI_

src/binding/database/database.h
```diff
@@ -117,22 +117,49 @@ inline void vtPopulateIfSettled(
...
 		if (static_cast<double>(oldestSnapshotSec) * 1000.0 < versionMs) {
 			return;
 		}
+		// Gate 2: sequence-number gate for backdated replicated versions.
+		// Only run when Gate 1 passes (there IS an open snapshot, but the
+		// version's timestamp predates it — exactly the backdated-write case).
+		uint64_t oldestSnapshotSeq = 0;
+		uint64_t latestSeq = db->GetLatestSequenceNumber();
+		if (db->GetIntProperty(cf, "rocksdb.oldest-snapshot-sequence", &oldestSnapshotSeq) &&
+		    oldestSnapshotSeq != 0 && oldestSnapshotSeq < latestSeq) {
```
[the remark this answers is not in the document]
kriszyp: Fixed in 47fea429 — removed the `oldestSnapshotSeq != 0` guard and added a seq-0 snapshot test. — Claude (Opus 4.8)

src/binding/core/verification_table.cpp
```diff
@@ -255,6 +255,34 @@ void VerificationTable::unrefTracker(LockTracker* t) {
...
+					v,
+					vtEncodeSettled(vtNextSettleGen()),
+					std::memory_order_release,
+					std::memory_order_acquire)) {
+				break;
+			}
+			// v has been reloaded by the failed CAS; if now a lock, exit.
+		}
+	}
+}
```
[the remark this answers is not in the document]
kriszyp: Fixed in 47fea429 — one generation minted per sweep (per-slot CAS only needs a value unseen in that slot). — Claude (Opus 4.8)

src/binding/database/db_handle.cpp
kriszyp: Good catch — fixed in 47fea429: Drop/DropSync now sweep the VT after a successful DropColumnFamily (ok path only; the already-dropped path defers to the dropping handle), with dropSync/drop invalidation tests added. — Claude (Opus 4.8)

src/binding/core/verification_table.cpp
kriszyp: Fixed in 47fea429 — comment now describes the actual retry-until-landed-or-lock loop. — Claude (Opus 4.8)

src/binding/database/database.cpp
```diff
@@ -1002,33 +1019,41 @@ napi_value Database::GetSync(napi_env env, napi_callback_info info) {
 
 	bool wantsPopulate = (flags & POPULATE_VERSION_FLAG) != 0;
 
+	// FIX C: For transactional reads, establish the snapshot BEFORE loading
```
cb1kenobi: These "FIX C" labels are nice for code reviewing, but don't make sense post-merge. Nobody looking at this code is going to understand what "FIX C" means, so I suggest removing them.

cb1kenobi (approved):

src/binding/database/database.cpp
kriszyp: Good call — done in a7a0be4. Dropped the `FIX A/B/C/D` prefixes across all four files (database.cpp, database.h, db_handle.cpp, transaction.cpp); the explanatory comment text is unchanged. — Claude (Opus 4.8)

PR #701, opened 2026-07-10
ci: wire Claude + Gemini AI review (ai-review-prompts callers)

heskew:
## What

Wires **@harperfast/rocksdb-js** into the AI review pipeline — the Phase-1 expansion target from the 07-02 retro (52 PRs/30d, highest risk-per-line of the candidate repos). Three thin callers of the [ai-review-prompts](https://github.com/HarperFast/ai-review-prompts) reusables, pinned to main **`1255f2e`** ([#72](https://github.com/HarperFast/ai-review-prompts/pull/72) — the `repo-type/native-addon` layer written for this repo):

- **claude-review.yml** — layers: `universal` + `repo-type/native-addon`; repo-specific checks for the pnpm/oxlint/oxfmt toolchain, vendored-RocksDB bumps, the three-runtime-deps bar, and the harper-core consumer coupling (MaybePromise / key-encoding changes need a core reference).
- **gemini-review.yml** — mirrors the Claude caller; opt-in via `gemini-review` label unless `GEMINI_ALWAYS_ON=true`.
- **validate-caller-workflows.yml** — SHA-pin + shadow-job guard.

Already done outside this PR: `repo:rocksdb-js` label in ai-review-log; `claude-review` / `gemini-review` labels here.

## Setup needed on merge (repo settings — can't be done from a PR)

1. **Secrets**: add `ANTHROPIC_API_KEY` (required); `GEMINI_API_KEY` + `AI_REVIEW_LOG_TOKEN` (optional — missing keys skip cleanly; the log token reuses the existing fine-grained PAT). Add this repo to the visibility list of the org-level `HARPERFAST_AI_CLIENT_ID` / `HARPERFAST_AI_APP_PRIVATE_KEY` secrets.
2. **Variable**: set `CLAUDE_ALWAYS_ON=true` to auto-review trusted-author PRs (or leave unset for label-only opt-in while calibrating).
3. **Required check**: make `validate / validate` a required status check on `main`.
4. ai-review-log README "Reviewed repos" table — follow-up PR there once this lands.

Note: the Claude review check on *this* PR will fail with the expected "workflow file must match default branch" guard (it always does on PRs that introduce the caller workflows — claude-code-action's anti-tamper). Ignore it; it clears after merge.

---

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -34,6 +34,22 @@ jobs:
     # `_claude-review.yml`'s authorize `if:`, not in this caller.
     if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
     uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@1255f2e155e2202992508034d21b421fa99d14be # main 2026-07-10 (#72 native-addon layer; on 9cf49d2 = #70 calibration + #71 prompt-ref tracking)
+    # Caller-side permissions at the calling-job level (NOT workflow-
+    # level — that placement caps the reusable's per-job grants below
+    # what they need and breaks the workflow at startup; see
+    # ai-review-prompts#39/#40). Union of what the reusable's authorize
+    # (`contents: read`) and review (`contents: read` + `pull-requests:
+    # write` + `id-token: write`) jobs declare. Without this block the
+    # grants are inherited from the repo's default-workflow-permissions
+    # setting instead (GitHub intersects the reusable's requests with
+    # the caller's ceiling — un-grantable scopes are silently dropped,
+    # and `pull-requests: write` survives only while the repo default
+    # is "write"). Explicit grants keep the caller independent of that
+    # setting, and match gemini-review.yml in this repo.
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
     with:
       # Same SHA as the `uses:` ref above. The reusable uses this to
       # check out HarperFast/ai-review-prompts (layer files + bash
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -1,0 +1,78 @@
+name: Claude PR Review
+
+# Thin caller of the reusable in HarperFast/ai-review-prompts. The single
+# `uses:` ref pin below controls everything that moves together — workflow
+# logic, layer files, bash scripts, auth-gate behavior. Bumping the pin
+# is the entire upgrade motion.
+#
+# Pre-requisites (org-level secrets, configured once on HarperFast):
+#   - HARPERFAST_AI_CLIENT_ID       (the App's Client ID, like Iv23li…)
+#   - HARPERFAST_AI_APP_PRIVATE_KEY (.pem file contents)
+#
+# Plus the per-repo / inherited:
+#   - ANTHROPIC_API_KEY             (required)
+#   - AI_REVIEW_LOG_TOKEN           (optional — if set, threads each run
+#                                   into a per-PR issue in HarperFast/ai-review-log)
+
+on:
+  pull_request:
+    # `labeled` admits the `claude-review` label gesture for
+    # bot-authored PRs (renovate, dependabot). See ai-review-prompts#38.
+    types: [opened, synchronize, reopened, labeled]
+
+concurrency:
+  group: claude-review-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  review:
+    # Always-on toggle — see ai-review-prompts USAGE.md "Reviewers & the
+    # always-on toggle". CLAUDE_ALWAYS_ON=true (repo/org variable) → auto-
+    # review trusted-author PRs; unset → opt-in via the claude-review
+    # label. The reusable's authorize job still owns WHO is admitted.
+    # Note: the `claude-review` label name is matched there too —
+    # `_claude-review.yml`'s authorize `if:`, not in this caller.
+    if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
+    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@1255f2e155e2202992508034d21b421fa99d14be # main 2026-07-10 (#72 native-addon layer; on 9cf49d2 = #70 calibration + #71 prompt-ref tracking)
+    with:
+      # Same SHA as the `uses:` ref above. The reusable uses this to
+      # check out HarperFast/ai-review-prompts (layer files + bash
+      # scripts) at the same ref as the workflow logic itself — keeps
+      # the upgrade motion atomic.
+      #
+      # The duplication is unavoidable: reusable workflows can't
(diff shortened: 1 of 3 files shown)


.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,78 @@
...
+jobs:
+  review:
+    # Always-on toggle — see ai-review-prompts USAGE.md "Reviewers & the
+    # always-on toggle". CLAUDE_ALWAYS_ON=true (repo/org variable) → auto-
+    # review trusted-author PRs; unset → opt-in via the claude-review
+    # label. The reusable's authorize job still owns WHO is admitted.
+    # Note: the `claude-review` label name is matched there too —
+    # `_claude-review.yml`'s authorize `if:`, not in this caller.
+    if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
+    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@1255f2e155e2202992508034d21b421fa99d14be # main 2026-07-10 (#72 native-addon layer; on 9cf49d2 = #70 calibration + #71 prompt-ref tracking)
```
cb1kenobi, line 36: **High: `review` job grants the reusable no `permissions`; `id-token: write` will be missing and claude-code-action will fail**

This job delegates to `_claude-review.yml`, whose `review` job declares `contents: read` + `pull-requests: write` + `id-token: write` (the last commented "required by claude-code-action even with API-key auth"). A reusable workflow's per-job permissions are capped by the caller's grant, and with no `permissions:` block here the ceiling is the repo default `GITHUB_TOKEN`. `id-token: write` is never included in the default token — regardless of the org's read vs read/write default — so the OIDC token can't be minted and claude-code-action fails at startup. `gemini-review.yml` in this same PR declares this block (lines 62-65); this caller is missing it. Add the union to the `review` job:

```yaml
  review:
    if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
    permissions:
      contents: read
      pull-requests: write
      id-token: write
    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@1255f2e155e2202992508034d21b421fa99d14be
```

—
_Generated by Barber AI_

.github/workflows/gemini-review.yml
```diff
@@ -0,0 +1,104 @@
...
+    uses: HarperFast/ai-review-prompts/.github/workflows/_gemini-review.yml@1255f2e155e2202992508034d21b421fa99d14be # main 2026-07-10 (#72 native-addon layer; on 9cf49d2 = #70 calibration + #71 prompt-ref tracking)
+    # Caller-side permissions at the calling-job level (NOT workflow-
+    # level — that placement caps the reusable's per-job grants below
+    # what they need and breaks the workflow at startup; see
+    # ai-review-prompts#39/#40). Union of what the reusable's authorize
+    # (`contents: read`) and review (`contents: read` + `pull-requests:
+    # write` + `id-token: write`) jobs declare.
+    permissions:
+      contents: read
+      pull-requests: write
```
cb1kenobi, line 64: **Medium: this caller is outside the validator's scope — the shadow-job / mutable-ref guard does not cover it**

`validate-caller-workflow.sh` (ai-review-prompts@1255f2e) globs only `.github/workflows/claude-*.yml` (`files=(.github/workflows/claude-*.yml)` and bails with "No claude-*.yml caller workflows found" otherwise). This new caller grants `pull-requests: write` and delegates to a reusable, but it will never be checked by the fail-closed shadow-job / SHA-pin validator that `validate-caller-workflows.yml` runs as the required `validate` check. A later edit that adds a non-`uses:` shadow job (running with this `pull-requests: write` grant, bypassing the reusable's auth gate) or downgrades the `@<sha>` pin to a mutable ref would still pass `validate` green — exactly the escalation the validator exists to stop. Extend the validator glob upstream to cover the gemini caller (e.g. `{claude,gemini}-*.yml`), or track a follow-up before making `validate` the required guard.

—
_Generated by Barber AI_

kriszyp (approved):
LGTM — clean add-only AI-review CI wiring. Correct `pull_request` trigger (not `pull_request_target`), SHA-pinned action refs, and a fail-closed prompt validator, so no fork-PR token-exposure path. Reviewed via review-queue.

— Claude (Sonnet 5)

.github/workflows/claude-review.yml
heskew: Fixed in 2753ede1 — the caller now grants `contents: read` + `pull-requests: write` + `id-token: write` at the calling-job level, matching `gemini-review.yml`.

One correction on the mechanism, verified empirically: the missing block does **not** fail the workflow at startup. The fleet callers (harper/oauth/harper-pro) have shipped without it since day one and run successfully daily — GitHub silently **intersects** the reusable's per-job requests with the caller's ceiling rather than erroring. A successful harper run this morning shows the granted set as exactly `Contents: read, PullRequests: write` (id-token dropped), and claude-code-action completes fine without OIDC on our API-key path.

The real exposure is subtler and worth fixing anyway: with no explicit block, `pull-requests: write` survives the intersection only because the repo's default-workflow-permissions setting is "write". Flip that setting to "read" and reviews start failing to post with unhelpful 403s. Explicit grants make the caller independent of the repo setting — applied, and the fleet callers should get the same treatment upstream.

.github/workflows/gemini-review.yml
heskew: Confirmed real, and fleet-wide rather than specific to this PR: `validate-caller-workflow.sh` at this pin globs only `.github/workflows/claude-*.yml`, so the gemini callers in harper/oauth/harper-pro are equally outside the shadow-job / mutable-ref guard today. The fix belongs upstream in ai-review-prompts (extend the glob to `claude-*.yml` + `gemini-*.yml` and accept `_gemini-review.yml` in the allowed-reusable check); consumer repos then pick it up on the next pin bump with no caller changes. Tracking as an ai-review-prompts follow-up — not blocking this PR, since merging it neither widens nor narrows the existing gap.

heskew:
Follow-ups from the review feedback here, for the paper trail:

- **Caller permissions fixed in this PR** (2753ede1) and now being retrofitted fleet-wide: HarperFast/harper#1790 and a matching harper-pro PR (oauth already had the block). Pattern documented upstream in HarperFast/ai-review-prompts#74 so future wirings can't drop it.
- **Validator gemini-caller gap** fixed upstream in HarperFast/ai-review-prompts#74 (glob extended to `gemini-*.yml`); consumers pick it up on their next validate-pin bump.
- Both trace back to HarperFast/harper#568 (the standing token-permissions audit) — the AI-review-caller slice of that issue closes with these.

cb1kenobi (approved):

kriszyp (approved):
Thin AI-review callers look safe — `pull_request` (not `pull_request_target`), explicit per-secret passing (no blanket inherit), job-level least-privilege perms, and dual 40-char SHA pins. One thing to file as a follow-up (not blocking): the shadow-job/mutable-ref validator globs only `claude-*.yml`, so the `gemini-review.yml` caller with the same write + id-token grants isn't covered — an upstream ai-review-prompts limitation. Approving.

_🤖 Reviewed with KrAIs (Claude Opus 4.8)_

PR #706, opened 2026-07-14
Use latest pnpm in CI workflows

### LINKED ISSUE #703: Re-enable latest pnpm
pnpm@11.12.0 introduced a bug (https://github.com/pnpm/pnpm/issues/12959) that blocks our CI. The solution was to pin pnpm to 11.11.0.

Once the issue is fixed, we need to restore the latest version in the CI workflows.

cb1kenobi:
We had to disable the latest pnpm version in CI workflows due to this bug: https://github.com/pnpm/pnpm/issues/12959.

Fixes #703.


kriszyp (changes requested):
This reintroduces the outage it's meant to close. Flipping the pins back to `version: latest` currently resolves to pnpm **11.12.0** — the exact release that introduced [pnpm#12959](https://github.com/pnpm/pnpm/issues/12959), which is still open. All 21 test/build jobs fail identically at the pnpm-install step with `Cannot use 'in' operator to search for 'integrity' in undefined` — the same error the 11.11.0 pin was created to avoid. The CI red is 100% caused by this diff, not pre-existing.

Two asks:
1. **Don't merge until #12959 is fixed upstream and CI is green.**
2. When it does land, please **pin to a specific known-good version** (e.g. `11.13.x` once it ships the fix) rather than `latest`. Unpinned `latest` is non-reproducible and is exactly how #12959 bit the first time — a floating pnpm can break or subtly change resolution with no diff in this repo. That still satisfies #703's intent (stop being stuck on 11.11.0) without giving up reproducibility.

_🤖 Reviewed with KrAIs (Claude Opus 4.8)_

kriszyp (approved):
Approving — CI-only change to use latest pnpm; whatever works here. — Claude

kriszyp (approved):
Approving — CI-only change to use latest pnpm; whatever works here. — Claude

kriszyp (approved):
test approve check

kriszyp (approved):
Approving — CI-only pnpm bump; whatever works here. (Prior duplicate/test approvals on this PR were an automation slip — this is the intended one.) — Claude

Issue #668, opened 2026-06-21
Data loss: IsBusy commit retry orphans an uncommitted log position, pinning the committed watermark → committed reads truncate


kriszyp:
### Summary

When an optimistic transaction's `Commit()` returns **`IsBusy`** and is retried, the log position recorded by the *first* `writeBatch` is **orphaned** in `uncommittedTransactionPositions` and never erased. Because the committed-read watermark (`lastCommittedPosition`) is computed as `uncommittedTransactionPositions.front()` (the earliest in-flight position), that orphaned position becomes a permanent floor: **every subsequent committed read on that log is capped at the orphaned offset**, even though the log file physically contains all the later entries. The result is silent, permanent truncation of committed reads with no error.

### Impact

- **Silent permanent data loss** for any consumer of committed log reads after an IsBusy retry. We hit it as a **replication backlog never reconciling** (Harper [harper-pro#426](https://github.com/HarperFast/harper-pro/issues/426)): a follower resumes from a valid cursor `T`, the leader's resume iterator drains a short prefix and then returns `done` hundreds of records short of head; the follower sits permanently behind, `connected: true`, no error.
- Not limited to replication — **any** committed read bounded by `lastCommittedPosition` (audit-log reads, time-travel, etc.) truncates at the pinned offset.
- No self-heal: the orphaned position is the `front()` minimum forever, and no flush path advances past it.

### Root cause (verified on `main`)

`src/binding/transaction/transaction.cpp` (async path; sync path is the same shape at ~490/510):

1. `writeBatch(..., txnHandle->committedPosition)` (~:268) inserts position **A** into `uncommittedTransactionPositions` and durably writes A's bytes to the log file; `committedPosition` now references A.
2. `Commit()` returns **`IsBusy`**.
3. `commitFinished(...)` is **gated on `!IsBusy()`** (~:300) → **skipped**. A is not erased.
4. `resetTransaction()` (~:319) recreates the txn but **does not reset/erase `committedPosition`** (`src/binding/transaction/transaction_handle.cpp` — clears `txn`/`logEntryBatch`/`snapshotSet` only).
5. The retry's `writeBatch` overwrites `committedPosition` with a **new** position **B**. **A is now orphaned** — no handle references it, so neither `commitFinished` nor `commitAborted` can ever target it.
[...]

PR #710, opened 2026-07-15
fix(txn): align ERR_TRY_AGAIN with the IsBusy reset path — reset in place, publish only on real commit

kriszyp:
Pairs with the HarperFast/harper PR "retry ERR_TRY_AGAIN on the same transaction" (linked in first comment) — the two must land together; this side releases first as **2.5.0**, which is harper's new dependency floor.

## Problem

An optimistic commit returns `TryAgain` (not `Busy`) when the transaction's snapshot is stranded outside the memtable window — the sequence history the conflict check needs was flushed away (bulk-ingest bursts, compaction). Two defects followed:

1. **Premature transaction-log publish.** `commitFinished` — which advances the committed-read watermark, making log/change-feed entries visible — was gated on `!IsBusy()`. `TryAgain` is not `Busy`, so a **failed** TryAgain commit published its entries while the data was rolled back: a change-feed entry visible ahead of its data for every reader in that window.
2. **No native retry path.** `resetTransaction()` fired only on `IsBusy`, so recommitting after `TryAgain` re-checked the same stranded snapshot forever. harper#1695's source-apply wedge (uncapped retry spinning, replication apply loop frozen) is the field manifestation; harper#1696 worked around it in the JS layer by replaying onto a fresh transaction — which depended on the premature publish of defect 1.

## Fix

Treat `TryAgain` exactly like `Busy`:

- `commitFinished` is now gated on `status.ok()` — entries publish only when the data commit actually succeeds. On a hard error, `close()`'s `commitAborted` stops the position pinning the watermark (bytes cannot be unwritten; abandoning a logged transaction remains the loudly-flagged `ERR_TRANSACTION_ABANDONED` case, unchanged).
- `resetTransaction()` fires on `IsBusy() || IsTryAgain()` — preserves `committedPosition` (WAL stays write-once, #668) but takes a fresh snapshot, so the caller's re-run validates against current state and converges.
- TS adds a `TransactionRetryableError` base (shared by `TransactionIsBusyError` / new `TransactionTryAgainError`) expressing the contract: the native layer reset the handle; re-run the body and retry. Both `db.transaction()` retry loops now use one `instanceof` check. `TransactionAbandonedError` deliberately does not extend the base — retryability is the boundary.
- The coordinated-retry `RETRY_NOW` fast path stays `IsBusy`-only by design: it parks on the conflicting holder's VT-slot lock, and a stranded snapshot has no lock holder to park on (documented in-code).

## Validation

- New `test/transaction-log-tryagain-retry.test.ts`: converges on the **same** transaction id, WAL written exactly once, and — the discriminating assertion — the entry is **not visible mid-retry** (fails under the old `!IsBusy` gate with `expected 1 to be +0`).
- Forced-`TryAgain` is injected via a gated native test seam (`forceTryAgainForTesting`, inert unless called) because a real stranded snapshot can't be staged deterministically through the public API (OCC retains flushed-memtable history). The **real** repro lives in the paired harper PR: its `compact()`-induced `ERR_TRY_AGAIN` test converges on the same transaction with this build linked.
- Full suite: 636 passed / 1 skipped; #668 IsBusy write-once regression test unchanged and green; type-check/lint/fmt clean.

## Cross-model review (Codex ×2 + Harper-domain adjudication; Gemini leg failed — agy hang)

- **Blocker (addressed):** harper's dependency floor must encode this fix → released as **2.5.0**; harper PR bumps to `^2.5.0` and stays draft until publish.
- **Adjudicated not-new:** `transactionSync` fresh-`Transaction`-per-attempt lifecycle, worker-thread `resetTransaction` envelope, and operation-vs-state log semantics are all the pre-existing `IsBusy`/#668 design that `TryAgain` now joins — no new hazard; the publish-gate change is strictly safer.
- **Considered, declined:** debug-gating the test seam — CI exercises release builds, and the seam is inert unless explicitly armed (precedent: `transactionLogMapCount`).
- **Follow-up (pre-existing):** async/sync retry-bound off-by-one (`attempt < maxRetries` vs `<=`) — present for `IsBusy` before this change; aligning it changes user-visible retry counts, so it deserves its own decision.

*KrAIs, via Claude*


kriszyp:
Paired harper PR: HarperFast/harper#1823 (fix(txn): retry ERR_TRY_AGAIN on the same transaction) — draft until this releases as 2.5.0.

*KrAIs, via Claude*

kriszyp:
@heskew Could you help unblock CI here? The only failing check, `review / authorize`, isn't caused by this PR's diff — every real check (build, native tests, all Node/Bun/Deno matrices, benchmarks, stress) is green.

`review / authorize` fails in 4s with:

```
Error: The 'client-id' (or deprecated 'app-id') input must be set to a non-empty string.
```

`claude-review.yml`/`gemini-review.yml` pass `secrets.HARPERFAST_AI_CLIENT_ID` / `HARPERFAST_AI_APP_PRIVATE_KEY` into the `HarperFast/ai-review-prompts` reusable workflow, but `rocksdb-js` has no secret by that name — it has `GH_APP_CLIENT_ID` / `GH_APP_PRIVATE_KEY` instead. `harper` and `harper-pro` both have `HARPERFAST_AI_CLIENT_ID`/`HARPERFAST_AI_APP_PRIVATE_KEY` set at the repo level, so it looks like `rocksdb-js` was never onboarded with the correctly-named secret (or its App secret predates the naming standardization). It only shows up here because the authorize job only actually runs when `CLAUDE_ALWAYS_ON`/`GEMINI_ALWAYS_ON` is true or the PR is labeled — every other open PR in the repo shows this check as `skipping`.

Could you add `HARPERFAST_AI_CLIENT_ID`/`HARPERFAST_AI_APP_PRIVATE_KEY` repo secrets to `rocksdb-js` (copying from `GH_APP_CLIENT_ID`/`GH_APP_PRIVATE_KEY` if it's the same App, or from `harper`/`harper-pro` if not)? Happy to help further if useful.

— Claude (dev-agent)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@harperfast/rocksdb-js",
-  "version": "2.4.0",
+  "version": "2.5.0",
```
cb1kenobi: Don't bump the version. Let the release workflow do this.
```suggestion
  "version": "2.4.0",
```

Side note, this code suggests a minor bump, but this PR says it's a fix which would be a patch.

cb1kenobi (changes requested):

cb1kenobi (approved):
LGTM!

PR #716, opened 2026-07-21
Fix async transaction reads across column families

### LINKED ISSUE #1881: Secondary-index reads return partial/empty results for the second table accessed in a single request (5.1.19 and 5.1.22)
## Summary

In Harper **5.1.19 and 5.1.22**, within a **single resource request**, the first
access to a table returns an **incomplete** result set if a **different table**
was read earlier in that same request. Reading that table first — or as the only
read in the request — returns the full set.

Only the **secondary-index** read path is affected. An unconditioned full scan of
the same table, in the same position in the same request, returns every row. The
records are intact; the index scan comes back short.

The magnitude ranges from a partial count to **zero**. It silently
empties/undercounts any endpoint that aggregates across multiple tables in one
request.

## Environment

- Harper **5.1.19**, and reproduced identically on **5.1.22**.
- `@harperfast/rocksdb-js` **2.4.0**.
- Reproduced on a single node (macOS) and on a 2-node Fabric cluster.
- Component plugins: `rest` + `graphqlSchema` + `jsResource`; TypeScript resources.

## Reproduction — self-contained, no real data required

Two tables, one indexed attribute, seeded synthetically. **100% deterministic.**

The essential ingredient is the write pattern: every key gets rows on **every
wave**, so a single key's index entries are scattered across many sorted runs
rather than sitting contiguously in one. Seeding the same rows in a single pass
does **not** reproduce.

### `schemas/repro.graphql`

```graphql
type GenA @table(database: "metrics-repro") @export {
  id: ID @primaryKey
  repositoryId: String @indexed
  n: Int @indexed
  body: String
}

type GenB @table(database: "metrics-repro") @export {
  id: ID @primaryKey
  repositoryId: String @indexed
  n: Int @indexed
  body: String
}
```

### `resources/Gen.ts`

```ts
import { Resource, databases } from "harper";

const db: any = (databases as any)["metrics-repro"];
const PAD = "x".repeat(1024);

// Every key gets rows on EVERY wave, so one key's index entries end up
// scattered across many sorted runs rather than contiguous in one.
async function generate(table, keys, perKeyPerWave, waves, bodyKb) {
  const body = PAD.repeat(bodyKb);
  for (let w = 0; w < waves; w++)
    for (let r = 0; r < keys; r++)
      for (let i = 0; i < perKeyPerWave; i++) {
        const n = w * perKeyPerWave + i;
        await db[table].put({
          id: `${table}-r${r}-n${n}`,
          repositoryId: `repo-${r}`,
          n,
          body,
        });
      }
}

async function drain(table, key, scan) {
  const query =
    scan === "full"
      ? { conditions: [] }
      : {
          operator: "and",
          conditions: [
            { attribute: "repositoryId", comparator: "equals", value: key },
          ],
        };
  let n = 0;
  for await (const r of db[table].search(query)) {
    if (scan === "full" && r.repositoryId !== key) continue;
    n++;
  }
  return n;
}

export class Gen extends Resource {
  static loadAsInstance = false;

  async get(target: any) {
    const p = (k, d) => String(target?.get?.(k) ?? d);
    if (p("action", "probe") === "gen") {
      const table = p("table", "GenA");
      await generate(table, 50, 5, 20, 4); // 50 keys x 5 rows x 20 waves, 4KB body
      return { table, written: 50 * 5 * 20 };
    }
    const key = p("key", "repo-0");
    const scan = p("scan", "index");
    const order = p("tables", "GenB,GenA").split(",").filter(Boolean);
    const out = {};
    for (let i = 0; i < order.length; i++)
      out[`${i + 1}:${order[i]}`] = await drain(order[i], key, scan);
    return out;
  }
}
```

### Steps

```bash
# 1. Seed (once)
curl -k "$HOST/Gen?action=gen&table=GenA"
curl -k "$HOST/Gen?action=gen&table=GenB"

# 2. Probe — each request counts repo-0 in the given table order
curl -k "$HOST/Gen?tables=GenA"
curl -k "$HOST/Gen?tables=GenA,GenB"
curl -k "$HOST/Gen?tables=GenB,GenA"
curl -k "$HOST/Gen?tables=GenB,GenA&scan=full"
```

### Results — every count should be 100

| Request | Result |
|---|---|
| `?tables=GenA` | `GenA = 100` ✅ |
| `?tables=GenB` | `GenB = 100` ✅ |
| `?tables=GenA,GenB` | `100`, `100` ✅ |
| `?tables=GenB,GenA` | `100`, **`GenA = 0`** ❌ |
| `?tables=GenB,GenA&scan=full` | `100`, **`100`** ✅ (full scan is correct) |

Stable across 5/5 consecutive runs and across every key (`repo-0`, `repo-1`,
`repo-7`, `repo-23`, `repo-49` — all return `GenA = 0`).

## What this narrows it to

1. **It is the secondary-index path, not the data.** Same table, same request
   position: via the indexed attribute → `0`; via an unconditioned full scan
   filtered in memory → `100`. Every record is present and readable.

2. **The damage belongs to the table read *second*, not the one read first.**
   Crossing a long-lived production table with a freshly-created copy of it:

   | Read order | Result |
   |---|---|
   | **old** `PullRequest` → **new** `ReproIssue` | 962, **895** ✅ |
   | **new** `ReproPullRequest` → **old** `Issue` | 962, **547** ❌ |
   | **old** `Commit` → **new** `ReproIssue` | 2820, **895** ✅ |
   | **new** `ReproCommit` → **old** `Issue` | 2820, **547** ❌ |

   What precedes it is irrelevant; it only has to be a different table.

3. **It is latched per request.** `GenB → GenA → GenA → GenA` returns the same
   wrong count every time. Once a table's read view is established within a
   request, re-reading does not recover the missing entries.

4. **Row content is not involved.** Copying the rows of an affected production
   table into a brand-new table — identical fields, values and indexes — does
   **not** reproduce. Only the write *history* differs.

5. **The losses are contiguous key runs.** On the production `Issue` table
   (895 → 547), walking the result in primary-key order gives 74 kept-runs and
   74 missing-runs (kept runs 1–155 entries, missing runs 1–24). Whole
   contiguous stretches drop out — consistent with a sorted run missing from the
   merged iterator view rather than per-record filtering.

6. **It is not a stale-by-time snapshot.** Rows written in the *same* bulk write
   split 543 kept / 327 lost. Neither `syncedAt` nor `updatedAt` separates the
   kept set from the missing set.

7. **`@harperfast/rocksdb-js` 2.4.0 does not reproduce it standalone.** Driving
   the binding directly — two column families on one database, the same
   interleaved wave seeding, index-shaped keys, then `getRange` on B followed by
   A — returns the full count in every order. That test does not replicate
   Harper's per-request transaction/snapshot handling or its record cache, so it
   is suggestive rather than conclusive, but it points at the Harper layer.

## What does NOT fix it

- Running the reads sequentially (awaited one at a time) instead of `Promise.all`.
- Explicitly closing each async iterator (`iterator.return()`) after draining.
- Wrapping each read in its own `transaction()`.
- Re-reading the affected table later in the same request.
- Upgrading 5.1.19 → 5.1.22.

## Impact

Any endpoint aggregating across multiple tables in a single request silently
returns partial or zero data for every table except the first one read. In our
application this zeroed out issues, commits and releases on a dashboard that
reads `PullRequest` + `Issue` + `Commit` + `Release` in one `Promise.all`, while
each of those tables returns correct data when queried on its own.

The only reliable workaround we have found is **one table per request**, which is
a significant restructuring for aggregation endpoints.

## Questions

- Is this a known issue, and is a fix planned?
- Where is the per-request read view for a table established, and is there a
  supported way to force a fresh/isolated index read per table within one
  request?
- Does a full compaction of the affected column family clear it? If so we would
  take that as an interim mitigation — can you point us at the right operation?

Happy to provide access to a reproducing instance, or the full probe component.

---
Comment generated by kAIle (Claude Opus 4.8)

kriszyp:
Fixes HarperFast/harper#1881 by carrying the caller-resolved column family into deferred transaction reads, preserving database-wide transaction and snapshot behavior for cold cross-table reads. Adds a regression covering optimistic and pessimistic modes with and without snapshots.

Reviewer attention: the async state pins the target column-family descriptor only through the worker read, then releases it before signaling transaction completion so database teardown cannot outlive the RocksDB handle. Cross-model coverage: codex ✓ / gemini ✓; no unresolved blockers or significant concerns.

Generated by GPT-5 Codex.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -245,6 +245,13 @@ C++ code that needs to emit to JS without a database context should call
    unsupported/errored or reporting 0 (which also lets a genuinely-full local volume through — a real 0
    is indistinguishable from the spurious 0 some network filesystems report). The stream-target backup
    path never opens a `BackupEngine` against a volume and is not checked.
+9. **Transaction reads are database-wide, but column-family-specific**: one native transaction may
+   serve stores backed by any column family in its database. `TransactionHandle::get` must therefore
+   use the column family resolved from its `dbHandleOverride` in both the synchronous cache probe and
+   the queued disk read. The async state pins that `ColumnFamilyDescriptor` while the worker uses its
+   native handle, then releases the pin before `signalExecuteCompleted()` because that signal may let
+   transaction/database teardown proceed (HarperFast/harper#1881). Keep the cold-read cross-column-
+   family coverage in `test/transactions.test.ts`; cache hits alone do not exercise the async path.
 
 ## Debugging native heap corruption
 
--- a/src/binding/database/database.h
+++ b/src/binding/database/database.h
@@ -364,16 +364,22 @@ struct AsyncGetState final : BaseAsyncState<T> {
 		napi_env env,
 		T handle,
 		rocksdb::ReadOptions& readOptions,
-		std::string key
+		std::string key,
+		std::shared_ptr<ColumnFamilyDescriptor> targetColumnDescriptor = nullptr
 	) :
 		BaseAsyncState<T>(env, handle),
 		readOptions(readOptions),
-		key(std::move(key)) {}
+		key(std::move(key)),
+		targetColumnDescriptor(std::move(targetColumnDescriptor)) {}
 
 	rocksdb::ReadOptions readOptions;
 	// the data for key and value both need to be owned by AsyncGetState, so we need to use std::string (RocksDB Slice doesn't preserve ownership)
 	std::string key;
 	std::string value;
+	// Pins a transaction read's resolved column family until the native worker
+	// finishes. It must be reset before signalling transaction completion, which
+	// may allow the database to close.
+	std::shared_ptr<ColumnFamilyDescriptor> targetColumnDescriptor;
 
 	// Verification table state for post-read check and populate.
 	bool hasExpectedVersion = false;
--- a/src/binding/transaction/transaction_handle.cpp
+++ b/src/binding/transaction/transaction_handle.cpp
@@ -364,7 +364,13 @@ napi_value TransactionHandle::get(
 	));
 
 	readOptions.read_tier = rocksdb::kReadAllTier;
-	auto state = new AsyncGetState<TransactionHandle*>(env, this, readOptions, std::move(key));
+	auto state = new AsyncGetState<TransactionHandle*>(
+		env,
+		this,
+		readOptions,
+		std::move(key),
+		dbHandle->columnDescriptor
+	);
 	state->vtSlot = vtSlot;
 	state->vtObserved = observedSlot;
 	state->hasExpectedVersion = hasExpectedVersion;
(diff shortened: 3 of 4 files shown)


kriszyp:
Closing in favor of #717

PR #717, opened 2026-07-21
Resolve transaction reads against the caller's column family

kylebernhardy:
## Summary

A transaction belongs to a database, not one column family. Cold-cache asynchronous reads issued through another DBI were falling back to the transaction's original column family, so the same query could return complete results while warm and partial or empty results after restart.

This PR now:

- routes synchronous and asynchronous transaction reads through the caller's column family;
- routes transactional `getKeysCount()` through that same column family;
- applies the transaction snapshot to `getKeysCount()`, while `disableSnapshot` continues to observe the latest committed state;
- balances `Database::Get` async-work registration so close waits for in-flight ordinary reads.

Surfaced by [harper#1881](https://github.com/HarperFast/harper/issues/1881).

## Native lifetime handling

The async path pins the selected `ColumnFamilyDescriptor`, not a bare RocksDB column-family handle. Its ownership sequence is deliberate:

1. register transaction setup before inspecting `txn`/state so descriptor-wide close cannot tear down the transaction mid-call;
2. for a foreign DBI, register setup against that `DBHandle` before copying its column descriptor, closing the cross-environment open-check/capture race;
3. transfer the transaction registration to `AsyncGetState` only when queueing the worker;
4. reset the descriptor pin before `signalExecuteCompleted()`, because that signal can unblock teardown, which destroys column families before the RocksDB database;
5. clean up references, async work, state, registrations, and the descriptor pin on every N-API setup/queue failure path.

The worker no longer reads members from a concurrently closing target `DBHandle`.

## Testing

`test/transaction-cross-column-family.test.ts` covers:

- cold-cache async reads through a foreign column family;
- no fallback to the transaction's original column family;
- alternating reads across the origin and foreign column families;
- reads spanning three column families;
- queued reads surviving closure of the target DBI after issuance;
- cross-column-family counts, snapshot stability after an outside commit, and latest-state behavior with `disableSnapshot`.

Local final-artifact gates:

- Release and Debug native builds: passed
- `pnpm check`: passed
- focused regression: 6/6 passed
- native GoogleTests: 90/90 passed
- full Vitest suite: 644 passed, 2 skipped across 48 files
- cross-model coverage: Codex ✓ / Gemini CLI ✗ → direct API ✓; final Harper-domain adjudication found no blockers or significant concerns

## Deliberately out of scope

- The pre-existing closed-transaction guard for `TransactionHandle::getCount()` requires a separate N-API boundary change.
- General transactional range iterators (`getRange()` / `getKeys()`) still have their pre-existing snapshot-semantics question; this PR fixes `getKeysCount()`, the iterator path directly changed here.

### DIFF
--- a/src/binding/transaction/transaction_handle.cpp
+++ b/src/binding/transaction/transaction_handle.cpp
@@ -368,11 +368,10 @@ napi_value TransactionHandle::get(
 	// Read against the column family the caller issued the read on, which is not
 	// necessarily the one this transaction was created with. Resolve the column
 	// family here on the JS thread and hold it, so the worker never traverses a
-	// DBHandle that close() may concurrently tear down.
+	// DBHandle that close() may concurrently tear down. The block-cache attempt
+	// above already dereferenced columnDescriptor on this thread, so it is live.
 	state->readDbHandle = dbHandle;
-	if (dbHandle->columnDescriptor) {
-		state->readColumnFamily = dbHandle->columnDescriptor->column;
-	}
+	state->readColumnFamily = dbHandle->columnDescriptor->column;
 	state->vtSlot = vtSlot;
 	state->vtObserved = observedSlot;
 	state->hasExpectedVersion = hasExpectedVersion;
@@ -389,10 +388,11 @@ napi_value TransactionHandle::get(
 			auto state = reinterpret_cast<AsyncGetState<TransactionHandle*>*>(data);
 			// check if database is still open before proceeding
 			auto& readDbHandle = state->readDbHandle;
+			// Abort if either the transaction or the column family being read was
+			// torn down after this work was queued.
 			if (
 				!state->handle
-				|| !readDbHandle
-				|| !state->readColumnFamily
+				|| state->handle->isCancelled()
 				|| !readDbHandle->opened()
 				|| readDbHandle->isCancelled()
 			) {
--- a/src/binding/transaction/transaction_handle.cpp
+++ b/src/binding/transaction/transaction_handle.cpp
@@ -365,6 +365,14 @@ napi_value TransactionHandle::get(
 
 	readOptions.read_tier = rocksdb::kReadAllTier;
 	auto state = new AsyncGetState<TransactionHandle*>(env, this, readOptions, std::move(key));
+	// Read against the column family the caller issued the read on, which is not
+	// necessarily the one this transaction was created with. Resolve the column
+	// family here on the JS thread and hold it, so the worker never traverses a
+	// DBHandle that close() may concurrently tear down.
+	state->readDbHandle = dbHandle;
+	if (dbHandle->columnDescriptor) {
+		state->readColumnFamily = dbHandle->columnDescriptor->column;
+	}
 	state->vtSlot = vtSlot;
 	state->vtObserved = observedSlot;
 	state->hasExpectedVersion = hasExpectedVersion;
@@ -380,12 +388,19 @@ napi_value TransactionHandle::get(
 		[](napi_env doNotUse, void* data) { // execute
 			auto state = reinterpret_cast<AsyncGetState<TransactionHandle*>*>(data);
 			// check if database is still open before proceeding
-			if (!state->handle || !state->handle->dbHandle || !state->handle->dbHandle->opened() || state->handle->dbHandle->isCancelled()) {
+			auto& readDbHandle = state->readDbHandle;
+			if (
+				!state->handle
+				|| !readDbHandle
+				|| !state->readColumnFamily
+				|| !readDbHandle->opened()
+				|| readDbHandle->isCancelled()
+			) {
 				state->status = rocksdb::Status::Aborted("Database closed during transaction get operation");
 			} else {
 				state->status = state->handle->txn->Get(
 					state->readOptions,
-					state->handle->dbHandle->getColumnFamilyHandle(),
+					state->readColumnFamily.get(),
 					state->key,
 					&state->value
 				);
@@ -421,7 +436,8 @@ void TransactionHandle::getCount(
 	uint64_t& count,
 	std::shared_ptr<DBHandle> dbHandleOverride
 ) {
-	std::unique_ptr<DBIteratorHandle> itHandle = std::make_unique<DBIteratorHandle>(this, itOptions);
+	std::unique_ptr<DBIteratorHandle> itHandle =
+		std::make_unique<DBIteratorHandle>(this, itOptions, dbHandleOverride);
 	for (count = 0; itHandle->iterator->Valid(); ++count) {
 		itHandle->iterator->Next();
 	}
--- a/src/binding/transaction/transaction_handle.cpp
+++ b/src/binding/transaction/transaction_handle.cpp
@@ -11,6 +11,70 @@
 
 namespace rocksdb_js {
 
+namespace {
+
+/**
+ * Keeps a target DBHandle open while a transaction resolves and pins that
+ * handle's column descriptor. This bridges the gap between the caller's open
+ * check and the transaction's async-work registration without making the
+ * worker access a concurrently closing DBHandle.
+ */
+struct ScopedAsyncWorkRegistration {
+	AsyncWorkHandle* handle;
+
+	explicit ScopedAsyncWorkRegistration(AsyncWorkHandle* handle)
+		: handle(handle) {
+		if (this->handle) {
+			this->handle->registerAsyncWork();
+		}
+	}
+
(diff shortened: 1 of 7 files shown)


src/binding/transaction/transaction_handle.cpp
```diff
@@ -380,12 +388,19 @@ napi_value TransactionHandle::get(
...
 			// check if database is still open before proceeding
-			if (!state->handle || !state->handle->dbHandle || !state->handle->dbHandle->opened() || state->handle->dbHandle->isCancelled()) {
+			auto& readDbHandle = state->readDbHandle;
+			if (
+				!state->handle
+				|| !readDbHandle
+				|| !state->readColumnFamily
+				|| !readDbHandle->opened()
+				|| readDbHandle->isCancelled()
+			) {
```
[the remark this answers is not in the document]
kylebernhardy: Good catch on the asymmetry — applied in 63a5142d, and it also closes the open question I'd flagged in the PR description.

The worker now checks the transaction's own cancellation:

```cpp
if (
    !state->handle
    || state->handle->isCancelled()
    || !readDbHandle->opened()
    || readDbHandle->isCancelled()
) {
```

`isCancelled()` comes from `AsyncWorkHandle`, which `TransactionHandle` inherits — verified before applying.

I also took the second half of your point and removed the two null checks I'd added. You're right that they were defensive rather than load-bearing: the block-cache attempt earlier in `get()` already dereferences `dbHandle->getColumnFamilyHandle()` on the JS thread, so `columnDescriptor` is provably live by the time we capture `column` a few lines later, and `readDbHandle` is assigned from a non-null local. Both are gone; the capture is now unconditional.

I kept the pre-existing `!state->handle` check rather than removing it, since it predates this change.

Full suite green (642 passed, 1 skipped).

Comment generated by kAIle (Claude Opus 4.8)

src/binding/transaction/transaction_handle.cpp
```diff
@@ -421,7 +436,8 @@ void TransactionHandle::getCount(
 	uint64_t& count,
 	std::shared_ptr<DBHandle> dbHandleOverride
 ) {
-	std::unique_ptr<DBIteratorHandle> itHandle = std::make_unique<DBIteratorHandle>(this, itOptions);
+	std::unique_ptr<DBIteratorHandle> itHandle =
+		std::make_unique<DBIteratorHandle>(this, itOptions, dbHandleOverride);
```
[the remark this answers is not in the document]
kylebernhardy: I dug into this one and I'm going to leave it out of this PR — but the underlying gap you've spotted is real, so thank you for raising it.

Two things I verified before deciding:

**1. The suggested code would crash at one of the two call sites.** `Transaction::GetCount` (`transaction.cpp:598`) has no `try`/`catch`, so a `DBException` thrown out of `getCount` would escape the N-API callback uncaught rather than surfacing as a JS error. Several other methods in that file do wrap their bodies, but this one doesn't — so the fix needs to be paired with error handling at the boundary, not just a `throw`.

**2. The path this PR adds cannot reach a closed transaction.** `TransactionHandle::close()` calls `descriptor->transactionRemove(...)` (`transaction_handle.cpp:196`), so a closed transaction is gone from the registry. `Database::GetCount` — the caller that passes `dbHandleOverride`, i.e. the cross-column-family path this PR introduces — looks the transaction up with `transactionGet(txnId)` and already returns a clean error when that comes back null (`database.cpp:803-807`).

So the exposure is limited to `txn.getKeysCount()` called on an already-closed transaction via `Transaction::GetCount`, which is pre-existing and orthogonal to the column-family routing this PR is about. Fixing it properly means adding a state guard plus N-API error handling to that entry point, which I'd rather not fold into a targeted data-correctness fix that's waiting on review.

Happy to open a follow-up issue for it, or to include it here if @kriszyp would prefer it in one change.

Comment generated by kAIle (Claude Opus 4.8)

cb1kenobi, line 520: **Low: `getCount` doesn't apply the transaction snapshot, so a transactional count can diverge from snapshotted `get` reads**

Pre-existing and separate from the closed-transaction guard already discussed on this line — flagging for awareness, not blocking. `get`/`getSync` set `readOptions.snapshot = txn->GetSnapshot()` (and `get` calls `ensureSnapshot`), but `getCount` builds the iterator with `itOptions.readOptions` that never has the snapshot set, and it never calls `ensureSnapshot`. `Transaction::GetIterator` uses `read_options.snapshot` as-is and does not auto-apply the snapshot from `SetSnapshot()`, so `getKeysCount` reads latest-committed state while `get` in the same transaction reads the snapshot. With a concurrent commit between snapshot establishment and the count, `getKeysCount()` can disagree with the rows a `get`/iterator sees under the same `txn` — which is close to the cross-table counting use case (harper#1881) this PR fixes.

Now that cross-CF counts resolve the correct column family, this latent gap is actually reachable. Worth a follow-up (set `readOptions.snapshot` in `getCount` / route through `ensureSnapshot`), likely alongside the closed-transaction guard already deferred to its own PR.

Moderate confidence it matters in practice (depends on whether callers expect count to be snapshot-consistent); high confidence the divergence exists.

—
_Generated by Barber AI_

kylebernhardy: Confirmed, and it's broader than `getCount`. Thanks — this is a good catch and the reachability argument is right.

I grepped every assignment of `readOptions.snapshot` in the binding. There are exactly two, both in `transaction_handle.cpp`:

- `get` — line 321
- `getSync` — line 467 (via `ensureSnapshot()` at line 464)

**No iterator path sets it at all.** `DBIteratorOptions::setReadOptionDefaults()` doesn't, `initReadOptionsFromObject()` doesn't, and neither `DBIteratorHandle` constructor does. So this isn't specific to `getCount` — every transactional iterator (`txn.getRange()`, `txn.getKeys()`, `getKeysCount`) reads latest-committed while `get`/`getSync` in the same transaction read the snapshot.

On reachability, the JS layer splits interestingly:

- `Store.getCount` **does** forward the transaction — `context.getCount(options, this.getTxnId(options))` (`store.ts:607`) — so `db.getKeysCount({ transaction })` reaches `Database::GetCount` with a txnId. That is exactly the cross-column-family path this PR fixes, so you're right that the gap becomes meaningful here: before, that call counted the wrong column family outright; now it counts the right one, just not at the snapshot.
- `Store.getRange` never calls `getTxnId`, so `db.getRange({ transaction })` silently drops the transaction and never reaches the transactional iterator.

I'm leaving it out of this PR. Setting `readOptions.snapshot` on transactional iterators is a semantic change to *all* transactional iteration, not a localized fix, and whether iterators should be snapshot-consistent is a design call rather than an obvious bug — the `disableSnapshot` option means some callers deliberately opt out of snapshot pinning, and the JS layer's decision to drop the transaction in `getRange` hints the intent may be that range reads are non-transactional. That wants @kriszyp's read before anyone changes it.

Flagging it in the PR description alongside the closed-transaction guard so it stays visible rather than living only in a resolved thread.

Comment generated by kAIle (Claude Opus 4.8)

kriszyp: Addressed in 3cb37605. `getCount()` now establishes the transaction snapshot and applies `txn->GetSnapshot()` to the iterator read options; `disableSnapshot` leaves the snapshot null. The regression establishes a 25-key snapshot, commits a 26th key outside the transaction, verifies the transactional count stays 25, and separately verifies a snapshot-disabled transaction advances from 26 to 27 after an outside commit.

kriszyp:
Review follow-up pushed in 3cb37605. The async read now pins `ColumnFamilyDescriptor` only through worker execution and releases it before completion can unblock database teardown. Setup is registered against both the transaction and foreign DBHandle to close cross-environment close races, with exact-once RAII cleanup for N-API setup/queue failures. I also applied the transaction snapshot to `getKeysCount()` and added snapshot-enabled, snapshot-disabled, and target-DBI-close regressions. Final local gates: Release + Debug builds, `pnpm check`, 6/6 focused, 90/90 native, 644 passed / 2 skipped full suite. Cross-model coverage: Codex ✓ / Gemini CLI ✗→API ✓; final artifact adjudication clean.

kriszyp (approved):
Reviewed the final 3cb37605 artifact, including the cross-column-family routing, descriptor/transaction lifetime ordering, N-API failure cleanup, and snapshot-consistent counts. Local Release build, type/lint checks, focused regressions, native tests, and the full Vitest suite passed; the complete GitHub native/runtime/stress/benchmark matrix is green. No remaining blocking issues found.

PR #721, opened 2026-07-25
fix(database): apply the database's column family options to late-created column families

Minh-Ng:
## Summary

A named column family does not exist when a database is first created, so it is added immediately
after `DB::Open`. That path built its own options instead of reusing the ones the database was
opened with, so RocksDB's defaults applied. Values passed to `open()` were silently dropped for
those column families, as were this library's own tuned defaults:

- write buffer: 64MB, against a library default of 16MB or whatever was passed
- max write buffer number: 2 against 16
- max write buffer size to maintain: 0 against the derived value, leaving no memtable history for
  optimistic transaction conflict detection
- `noBlockCache`: ignored

Databases using only the default column family are unaffected. Stored data is unaffected either
way; these govern how memtables are sized and flushed, not what gets written.

The descriptor now saves the column family options used by `DB::Open`. When a handle creates a
column family later, it starts with those saved options, then applies `writeBufferSize`,
`maxWriteBufferNumber`, and `maxWriteBufferSizeToMaintain` from that handle.

Named column families now use the configured write buffer, 16MB by default rather than 64MB, so
they will flush more frequently than they do today.

## Test Plan

Added a test that opens two named column families, identical apart from write buffer size (64KB and
64MB), and writes 512KB to each. The smaller one flushes to disk unprompted; the larger one does
not. Both measured zero before this change, so the test fails without the fix.

Added a test that gives the default column family a 64MB write buffer and a named column family a
64KB write buffer. After writing 512KB to the named column family, the test confirms that it flushes
instead of inheriting the default column family's 64MB setting.

`noBlockCache` was confirmed manually: a named column family now reports no block cache, against
557KB of usage when caching is enabled. The remaining two options are corrected by the same change
but have no direct assertion, as I could not find a way to observe them through the JavaScript API.

### DIFF
--- a/src/binding/database/db_registry.cpp
+++ b/src/binding/database/db_registry.cpp
@@ -334,6 +334,8 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 				throw rocksdb_js::DBException("Column family \"" + name + "\" not found: cannot create column family in read-only mode");
 			}
 			DEBUG_LOG("%p DBRegistry::OpenDB Creating column family \"%s\"\n", instance.get(), name.c_str());
+			// Preserve retained settings while applying the current handle's
+			// per-CF compression and memory options.
 			auto cfOptions = entry.descriptor->cfOptions;
 			if (options.compression) {
 				cfOptions.compression = *options.compression;
@@ -342,6 +344,9 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 					? *options.compressionLevel
 					: rocksdb::CompressionOptions::kDefaultCompressionLevel;
 			}
+			cfOptions.write_buffer_size = static_cast<size_t>(options.writeBufferSize);
+			cfOptions.max_write_buffer_number = options.maxWriteBufferNumber;
+			cfOptions.max_write_buffer_size_to_maintain = options.maxWriteBufferSizeToMaintain;
 			auto column = rocksdb_js::createRocksDBColumnFamily(
 				entry.descriptor->db, name, cfOptions
 			);
--- a/src/binding/database/db_registry.cpp
+++ b/src/binding/database/db_registry.cpp
@@ -334,9 +334,9 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 				throw rocksdb_js::DBException("Column family \"" + name + "\" not found: cannot create column family in read-only mode");
 			}
 			DEBUG_LOG("%p DBRegistry::OpenDB Creating column family \"%s\"\n", instance.get(), name.c_str());
-			// Preserve retained settings while applying the current handle's
-			// per-CF compression and memory options.
-			auto cfOptions = entry.descriptor->cfOptions;
+			// Preserve retained settings while applying every per-CF option from
+			// the handle creating this family.
+			auto cfOptions = buildColumnFamilyOptions(options, entry.descriptor->cfOptions);
 			if (options.compression) {
 				cfOptions.compression = *options.compression;
 				cfOptions.blob_compression_type = *options.compression;
@@ -344,9 +344,6 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 					? *options.compressionLevel
 					: rocksdb::CompressionOptions::kDefaultCompressionLevel;
 			}
-			cfOptions.write_buffer_size = static_cast<size_t>(options.writeBufferSize);
-			cfOptions.max_write_buffer_number = options.maxWriteBufferNumber;
-			cfOptions.max_write_buffer_size_to_maintain = options.maxWriteBufferSizeToMaintain;
 			auto column = rocksdb_js::createRocksDBColumnFamily(
 				entry.descriptor->db, name, cfOptions
 			);
--- a/src/binding/database/db_registry.cpp
+++ b/src/binding/database/db_registry.cpp
@@ -334,7 +334,17 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 				throw rocksdb_js::DBException("Column family \"" + name + "\" not found: cannot create column family in read-only mode");
 			}
 			DEBUG_LOG("%p DBRegistry::OpenDB Creating column family \"%s\"\n", instance.get(), name.c_str());
-			auto column = rocksdb_js::createRocksDBColumnFamily(entry.descriptor->db, name, options.compression, options.compressionLevel);
+			auto cfOptions = entry.descriptor->cfOptions;
+			if (options.compression) {
+				cfOptions.compression = *options.compression;
+				cfOptions.blob_compression_type = *options.compression;
+				cfOptions.compression_opts.level = options.compressionLevel
+					? *options.compressionLevel
+					: rocksdb::CompressionOptions::kDefaultCompressionLevel;
+			}
+			auto column = rocksdb_js::createRocksDBColumnFamily(
+				entry.descriptor->db, name, cfOptions
+			);
 			auto columnDescriptor = std::make_shared<ColumnFamilyDescriptor>(column);
 			columns[name] = columnDescriptor;
 			entry.descriptor->columns[name] = columnDescriptor;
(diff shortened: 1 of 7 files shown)


test/db-options.test.ts
```diff
@@ -75,6 +75,40 @@ describe('Database write buffer options', () => {
 			expect(() => db.open()).toThrow('maxOpenFiles must be');
 		}));
 
+	// A named CF is created after DB::Open, so it misses the open options.
+	// Observed via flush behavior: RocksDB exposes no property for the size.
+	it('should apply writeBufferSize to a named column family', () =>
```
kriszyp, line 80: The production change also routes a missing CF through the already-open descriptor in `DBRegistry::OpenDB`, but these two fresh paths only exercise creation inside `DBDescriptor::open`. Could we add a same-path case that opens the default CF first with a 64 KiB buffer, then opens a missing named CF and verifies that it flushes? That would protect the retained `cfOptions` member and the registry wiring, which this test can currently pass without.

kriszyp (approved):
Nice, good catch.

---
Additional inline comments:
- `test/db-options.test.ts:80`: The production change also routes a missing CF through the already-open descriptor in `DBRegistry::OpenDB`, but these two fresh paths only exercise creation inside `DBDescriptor::open`. Could we add a same-path case that opens the default CF first with a 64 KiB buffer, then opens a missing named CF and verifies that it flushes? That would protect the retained `cfOptions` member and the registry wiring, which this test can currently pass without.

src/binding/database/db_registry.cpp
```diff
@@ -333,7 +333,9 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 				throw rocksdb_js::DBException("Column family \"" + name + "\" not found: cannot create column family in read-only mode");
 			}
 			DEBUG_LOG("%p DBRegistry::OpenDB Creating column family \"%s\"\n", instance.get(), name.c_str());
-			auto column = rocksdb_js::createRocksDBColumnFamily(entry.descriptor->db, name);
+			auto column = rocksdb_js::createRocksDBColumnFamily(
+				entry.descriptor->db, name, entry.descriptor->cfOptions
```
cb1kenobi: **Low: the already-open-DB late-CF path is untested**

The new test opens two *separate* databases (each via `generateDBPath()`), so it only exercises `DBDescriptor::open`'s late-CF branch. This branch — adding a named CF to an already-open DB using the freshly-stored `descriptor->cfOptions` — is the round-trip unique to this fix (options captured at first `DB::Open`, reused on a later `open()` for a new CF name) and has no coverage.

Worth a targeted test: open a DB with `writeBufferSize: 64 * 1024`, then open a second named CF on the *same* path, and assert it flushes (`rocksdb.total-sst-files-size > 0`). Note this path also means a `writeBufferSize` / `noBlockCache` passed on the *later* open call is intentionally dropped in favor of the first open's options — consistent with the PR's stated design, but a test would pin that behavior down.

—
_Generated by Barber AI_

test/db-options.test.ts
```diff
@@ -75,6 +75,40 @@ describe('Database write buffer options', () => {
...
+			async ({ db: smallBuffer }, { db: largeBuffer }) => {
+				const value = 'x'.repeat(1024);
+				for (const db of [smallBuffer, largeBuffer]) {
+					for (let i = 0; i < 512; i++) {
+						await db.put(`key-${i.toString().padStart(6, '0')}`, value);
+					}
+				}
+
+				// Flushes run in the background.
+				const deadline = Date.now() + 5000;
```
cb1kenobi: Would it be possible to force a `await db.flush()` and reduce the amount of time we need to wait? 5 seconds seems arbitrary, so I wonder if we could get this down to 1 second?

cb1kenobi (changes requested):

src/binding/database/db_registry.cpp
MinhNguyen-at: Covered by the new test discussed above

test/db-options.test.ts
MinhNguyen-at: Added test coverage. Used a 64 MiB buffer for the first handle and 64 KiB for the late named CF because these memory settings are per-CF and the late handle’s value should win. The test verifies that the late CF automatically flushes through the DBRegistry::OpenDB path.

test/db-options.test.ts
MinhNguyen-at: Yes 5 seconds is arbitrary, agreed polling is not ideal. Calling flush() would remove the behavior the test is checking. Instead, we limit the CF to two write buffers. When both are full, RocksDB pauses later writes until it automatically flushes one. Since the test awaits every put(), it cannot reach the assertion until that flush completes, so no separate timeout or polling is needed.

src/binding/database/db_registry.cpp
```diff
@@ -333,7 +333,15 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 				throw rocksdb_js::DBException("Column family \"" + name + "\" not found: cannot create column family in read-only mode");
 			}
 			DEBUG_LOG("%p DBRegistry::OpenDB Creating column family \"%s\"\n", instance.get(), name.c_str());
-			auto column = rocksdb_js::createRocksDBColumnFamily(entry.descriptor->db, name);
+			// Preserve the retained table/blob settings, but apply the current
+			// handle's per-CF memory options to the family it is creating.
+			auto cfOptions = entry.descriptor->cfOptions;
```
kriszyp: The follow-up correctly takes the three memtable fields from the handle creating this CF, but `noBlockCache` is also a per-CF setting: it is encoded in `ColumnFamilyOptions::table_factory` from that handle on the first-open path. Copying the descriptor's factory here makes the cache policy depend on whichever handle opened the path first. For example, opening the default family normally and then opening `late` with `noBlockCache: true` still lets a post-flush read populate the cache; reversing those options leaves `late` uncached even though it requested the default cache. Could we centralize construction of the complete `ColumnFamilyOptions` from the current `DBOptions` for both creation paths, or at least rebuild the table factory from `options.noBlockCache` here, and add a same-path regression with opposing cache policies?

kriszyp (commented):
Maybe an improvement could be made to centralizing construction of ColumnFamilyOptions? Otherwise LGTM.

src/binding/database/db_registry.cpp
MinhNguyen-at: Will centralize construction in buildColumnFamilyOptions() and use it from both creation path, thanks

kriszyp (commented):
Looks good, but a few failing tests, I wonder if they are failing on main?

Minh-Ng:
Looks like it is because the PR comes from my fork. Fork workflows do not receive secrets.GH_TOKEN, so .github/workflows/pr.yml:75 passes an empty token. scripts/init-rocksdb/get-prebuild.ts:15 then queries GitHub anonymously. Concurrent cache-miss jobs exhaust the anonymous API limit, explaining why only some matrix jobs fail. These will probably pass if re-run.

The benchmark workflow requests write access, but GitHub downgrades its token to read-only for the fork. Its attempt to create the benchmark comment at .github/workflows/benchmark.yml:189 then fails

cb1kenobi:
Re-reviewed `362c2171` with a focus on the compression options — no issues found.

The refactor to `buildColumnFamilyOptions()` does **not** affect compression. That helper only sets the table factory (block cache), blob settings, and the three write-buffer tunables; it never touches `compression`, `blob_compression_type`, or `compression_opts`. Compression is still applied on top afterward in every column-family creation path — the existing-CF and fresh-default paths in `DBDescriptor::open` (via `applyCompression`, unchanged), the late-add path in `open`, and the already-open path in `DBRegistry::OpenDB` — so a late-created CF gets the same algorithm/level it did before, and opening one family still can't restamp another's.

Also confirmed the previously-requested registry-path coverage is now in place: `should apply the current handle memory options to a late column family` shares one DB path across both handles, so it exercises `DBRegistry::OpenDB` (not just `DBDescriptor::open`).

The failing Node 24 / Deno-macos jobs are not code failures — they die at native-build setup with `Failed to fetch latest RocksDB release (403 rate limit exceeded)` in `get-prebuild.ts`, the fork-token rate limit @Minh-Ng described. Native C++ tests pass on all three OSes and the other JS matrices pass; these should clear on re-run.

This PR looks good, nice job!

—
_Generated by Barber AI_

cb1kenobi (approved):
Looks good, thanks!

Issue #201, opened 2025-10-28
Compression configuration


cb1kenobi:
Add configuration for compression. This includes the database option which can be as simple as enabled or more customizable as an object containing various settings.

RocksDB supports several compression algorithms and supports different compression settings at the SST file, data blocks and index blocks levels. Research how these apply to the data model.

PR #722, opened 2026-07-25
feat(database): add opt-in SST block compression

Minh-Ng:
## Summary

Adds a `compression` option accepting `true` (zlib), `false` (none), or an explicit algorithm name.
Omitting it leaves RocksDB's own default untouched, so existing callers are unaffected. That
default is snappy on builds that link it and none otherwise; the published prebuilt has no snappy,
so omitting the option today means no compression. Values at or above 2,048 bytes are stored in
blob files, which remain uncompressed and are unaffected by this option.

Requesting an algorithm the RocksDB build does not include causes `open()` to fail. This is
deliberate: RocksDB reports no error in that case and silently writes uncompressed files. The error
message lists what the running build supports.

The name resolver is in a translation unit free of Node dependencies so the GoogleTest suite can
link against it directly. Validation runs on every `open()` rather than only the first, so a call
that reuses an already-open path is checked as well.

This also rejects an `open()` requesting different compression on a path that is already
open. All handles on a path share one RocksDB instance, so a later open cannot change how table
files are written; previously the request was ignored and the caller would believe compression was
active. The check compares resolved `CompressionType` values rather than the requested names, since
the default varies by build as noted above. The error is uncoded, consistent with the existing
"already open in X mode" error rather than `ERR_DATABASE_READONLY`.

Depends on #721. Compression is a column family option and cannot reach named column families
without that fix; this diff includes that commit until #721 merges.

## Test Plan

Compression is asserted against on-disk SST size rather than a round trip, since an option that has
no effect still reads and writes correctly. The same payload produces substantially smaller SST
files with zlib than without compression. Verified on the default column family, a named column
family, and one added to an already-open database, which exercise three separate code paths.

Tests adapt to the compression algorithms the running build supports. Also verified with
`ROCKSDB_PATH` pointing to a build with all optional compressors disabled. The packaging smoke test
checks that the published prebuilt includes zlib.

Local runs are macOS arm64 only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,13 @@ Creates a new database instance.
 - `path: string` The path to write the database files to. This path does not need to exist, but the
   parent directories do.
 - `options: object` [optional]
+  - `compression: boolean | string` Enables SST block compression for newly written table files.
+    `true` selects `'zlib'`, the only algorithm the published prebuilt links (and slower per
+    write than `lz4`/`snappy`, so benchmark write-heavy databases); `false` means none. Omitting
+    it leaves whatever your RocksDB build defaults to, which is none in the published prebuilt but
+    is snappy in a build that links it. Naming another algorithm (`'snappy'`, `'lz4'`, `'lz4hc'`,
+    `'zstd'`, `'bzip2'`, `'none'`) throws unless your build links it. Applies database-wide and is fixed by the first `open()` of a
+    path, since every handle on a path shares one RocksDB instance.
   - `disableWAL: boolean` Whether to disable the RocksDB write ahead log. Defaults to `false`.
   - `enableStats: boolean` When `true` and the database is open, RocksDB will captures stats that
     are retrieved by calling `db.getStats()`. Enabling statistics imposes 5-10% in overhead.
--- a/README.md
+++ b/README.md
@@ -44,12 +44,14 @@ Creates a new database instance.
   parent directories do.
 - `options: object` [optional]
   - `compression: boolean | string` Enables SST block compression for newly written table files.
-    `true` selects `'zlib'`, the only algorithm the published prebuilt links (and slower per
-    write than `lz4`/`snappy`, so benchmark write-heavy databases); `false` means none. Omitting
-    it leaves whatever your RocksDB build defaults to, which is none in the published prebuilt but
-    is snappy in a build that links it. Naming another algorithm (`'snappy'`, `'lz4'`, `'lz4hc'`,
-    `'zstd'`, `'bzip2'`, `'none'`) throws unless your build links it. Applies database-wide and is fixed by the first `open()` of a
-    path; a later handle may omit it, but a conflicting value throws.
+    Values whose uncompressed size is 2,048 bytes or greater are stored in blob files, which remain
+    uncompressed; this option does not affect them. `true` selects `'zlib'`, the only algorithm the
+    published prebuilt links (and slower per write than `lz4`/`snappy`, so benchmark write-heavy
+    databases); `false` means none. Omitting it leaves whatever your RocksDB build defaults to, which
+    is none in the published prebuilt but is snappy in a build that links it. Naming another
+    algorithm (`'snappy'`, `'lz4'`, `'lz4hc'`, `'zstd'`, `'bzip2'`, `'none'`) throws unless your build
+    links it. Applies database-wide and is fixed by the first `open()` of a path; a later handle may
+    omit it, but a conflicting value throws.
   - `disableWAL: boolean` Whether to disable the RocksDB write ahead log. Defaults to `false`.
   - `enableStats: boolean` When `true` and the database is open, RocksDB will captures stats that
     are retrieved by calling `db.getStats()`. Enabling statistics imposes 5-10% in overhead.
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Creates a new database instance.
     it leaves whatever your RocksDB build defaults to, which is none in the published prebuilt but
     is snappy in a build that links it. Naming another algorithm (`'snappy'`, `'lz4'`, `'lz4hc'`,
     `'zstd'`, `'bzip2'`, `'none'`) throws unless your build links it. Applies database-wide and is fixed by the first `open()` of a
-    path, since every handle on a path shares one RocksDB instance.
+    path; a later handle may omit it, but a conflicting value throws.
   - `disableWAL: boolean` Whether to disable the RocksDB write ahead log. Defaults to `false`.
   - `enableStats: boolean` When `true` and the database is open, RocksDB will captures stats that
     are retrieved by calling `db.getStats()`. Enabling statistics imposes 5-10% in overhead.
--- a/src/binding/database/database.cpp
+++ b/src/binding/database/database.cpp
@@ -1,6 +1,7 @@
 #include <node_api.h>
 #include <algorithm>
 #include <cmath>
+#include <optional>
 #include <sstream>
 #include "database/database.h"
 #include "database/db_handle.h"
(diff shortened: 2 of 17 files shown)


src/binding/database/db_registry.cpp
```diff
@@ -333,7 +350,9 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 				throw rocksdb_js::DBException("Column family \"" + name + "\" not found: cannot create column family in read-only mode");
 			}
 			DEBUG_LOG("%p DBRegistry::OpenDB Creating column family \"%s\"\n", instance.get(), name.c_str());
-			auto column = rocksdb_js::createRocksDBColumnFamily(entry.descriptor->db, name);
+			auto column = rocksdb_js::createRocksDBColumnFamily(
```
kriszyp, line 359: Could we overlay this handle's per-CF memory options before creating the family? Reusing the first opener's entire `cfOptions` means a late CF silently ignores its own `writeBufferSize`, `maxWriteBufferNumber`, and `maxWriteBufferSizeToMaintain`. For example, default opened at 64 MiB followed by a new named CF requesting 64 KiB still gives the named CF 64 MiB. The public docs describe these knobs as per-CF. I suggest copying the descriptor base, overlaying the current handle's per-CF values, and retaining descriptor compression because that setting is intentionally database-wide; the test at `test/db-options.test.ts:81` should also exercise this already-open registry path.

src/binding/database/database.cpp
```diff
@@ -1402,6 +1403,19 @@ napi_value Database::Open(napi_env env, napi_callback_info info) {
 	}
 
 	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "name", dbHandleOptions.name));
+	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "compression", dbHandleOptions.compression));
+	// Validated here, not in DBDescriptor::open: an open that reuses an
+	// already-open path never reaches it.
+	if (!dbHandleOptions.compression.empty()) {
```
kriszyp: An explicit `compression: ''` reaches this as an empty string, so it is indistinguishable from omission and skips validation; the registry conflict guard skips it for the same reason. Untyped JS then silently gets RocksDB's default even though `tryResolveCompressionType("")` is explicitly invalid. Could we track property presence separately (for example with `std::optional`) or reject an explicitly empty value, with an integration test beside the `gzip` case?

test/native/compression_test.cc
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+// zlib backs the `compression: true` shorthand, and the prebuilt links libz
+// explicitly for this reason. If this fails on a platform, that shorthand is
+// broken there and needs a different default.
+TEST(CompressionTest, ZlibIsAvailable) {
+	rocksdb::CompressionType type = rocksdb::kNoCompression;
+	std::string error;
+	EXPECT_TRUE(tryResolveCompressionType("zlib", type, error)) << error;
```
kriszyp: This is a published-prebuild guarantee, not a property of every supported RocksDB build. With `ROCKSDB_PATH` pointing to a build without ZLIB, the resolver correctly rejects zlib but `pnpm test:native` fails. The integration assertion at `test/db-options.test.ts:284` has the same portability issue in the other direction: a Snappy-enabled default correctly conflicts with explicit `none`. Could these checks be gated by the build profile/effective support, or moved to packaging CI?

kriszyp (approved):
This looks good (not sure if the comment on db_registry.cpp is already covered by the other PR). And FYI, I believe only zlib is currently linked (I might add more).

src/binding/database/db_registry.cpp
```diff
@@ -312,6 +313,22 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
...
 		}
 
+		// Resolved types, not names: an omitted option leaves RocksDB's default.
+		if (!options.compression.empty()) {
+			rocksdb::CompressionType requested;
+			std::string error;
+			if (!tryResolveCompressionType(options.compression, requested, error)) {
+				throw rocksdb_js::DBException(error);
+			}
+			if (requested != entry.descriptor->cfOptions.compression) {
```
cb1kenobi, line 323: **Low: conflict check compares against the build-dependent default compression**

When the first `open()` of a path omits `compression`, `entry.descriptor->cfOptions.compression` holds RocksDB's default, which its `ColumnFamilyOptions` constructor sets to `Snappy_Supported() ? kSnappyCompression : kNoCompression`. On a build that links Snappy, opening a path first with omitted compression and then with explicit `'none'` makes `requested (kNoCompression) != kSnappyCompression` and throws a spurious `already open with compression 'default', requested 'none'`. The published prebuilt links no Snappy, so this is benign there, but it means the `should not treat an explicit none after an omitted compression as a conflict` test only passes on builds without Snappy — the behavior is build-dependent.

Suggested fix: only enforce the mismatch when the first open was itself explicit — e.g. skip the throw when `entry.descriptor->compression.empty()` (opened as default), matching the stated intent that a conflict is between two explicit values. Alternatively gate the no-conflict tests on Snappy support.

—
_Generated by Barber AI_

README.md
```diff
@@ -43,6 +43,13 @@ Creates a new database instance.
 - `path: string` The path to write the database files to. This path does not need to exist, but the
   parent directories do.
 - `options: object` [optional]
+  - `compression: boolean | string` Enables SST block compression for newly written table files.
```
cb1kenobi, line 46: **Low: doc omits that blob files (large values) are not compressed**

This option maps to `ColumnFamilyOptions::compression`, which compresses only SST blocks. Values at or above `min_blob_size` (2048 bytes, hardcoded) are stored in blob files, whose `blob_compression_type` is left at its default (`kNoCompression`) and is unaffected here. Given Harper routinely stores large values as blobs, enabling `compression` may yield little space savings on real payloads, which is surprising from the current wording.

Suggested fix: add a sentence noting the option applies to SST blocks only and that values >= 2048 bytes (stored as blob files) are not compressed by it.

—
_Generated by Barber AI_

src/binding/database/db_registry.cpp
Minh-Ng: Done. Late CF creation now copies the descriptor’s base options, overlays the current handle’s writeBufferSize, maxWriteBufferNumber, and maxWriteBufferSizeToMaintain, and retains descriptor compression/table settings. I also added a regression test covering the already-open registry path.

src/binding/database/database.cpp
Minh-Ng: Done. Compression is now parsed through std::optional, so an explicit '' is preserved and rejected by the resolver. Added an integration test beside the gzip case.

README.md
Minh-Ng: Added to README.md:47 and src/store.ts:87: values ≥2,048 bytes go into blob files, remain uncompressed, and are unaffected by this SST option.

test/native/compression_test.cc
Minh-Ng: Done. The tests now adapt to what the build supports, while release builds still verify that zlib is included. I also tested against a build with compression support disabled.

src/binding/database/db_registry.cpp
Minh-Ng: Fixed using second suggestion. The test now handles both types of build instead of assuming the default is always none. It expects a conflict when Snappy is available and no conflict when it isn’t.

Minh-Ng:
Thanks for the quick review. Addressed comments. Let me know if anything needs a second pass.

cb1kenobi:
@Minh-Ng Thank you for the pull request. We have been wanting to tackle https://github.com/HarperFast/rocksdb-js/issues/201 for quite some time now. Turns out we needed to put a bit of effort into getting the RocksDB prebuilds linked with the various compression libraries and to get `rocksdb-js`'s build system to support both RocksDB releases with and without compression linked.

Additionally, we want to also apply the compression settings to blobs and add some quality of life features like exposing the supported compression algorithms.

In the end, we landed https://github.com/HarperFast/rocksdb-js/issues/201 in favor of this pull request. Appreciate you help, thank you!

PR #723, opened 2026-07-27
fix(txnlog): recover the committed watermark from the log tail on load

### LINKED ISSUE #668: Data loss: IsBusy commit retry orphans an uncommitted log position, pinning the committed watermark → committed reads truncate

### LINKED ISSUE #1949: Crash recovery restores records but not the transaction log's committed watermark, silently stranding them from replication
## Summary

On an **unclean exit** (crash, OOM kill, container kill, power loss), records written in the last unflushed window before the crash are recovered into the primary store by boot replay, but the transaction log's *last-committed watermark* is not advanced to match — because that watermark is only persisted by the `shutdown()` hook, which a crash never runs. Every normal reader that is bounded by the watermark (replication's resume replay, `read_transaction_log`) stays blind to those records indefinitely. The data is durable and correct in the primary store; it is just invisible to replication and to the transaction log API, silently, forever (until some unrelated later write on that log happens to advance the watermark past the whole stale tail).

This is a genuine replication-durability defect: a peer that resumes replication after the crash never receives writes that were already acknowledged and durable before the crash.

## Mechanism (confirmed, not hypothesized)

- `core/resources/replayLogs.ts:95` replays the transaction log with `readUncommitted: true` on boot, so the **records** come back into the primary store correctly.
- `core/resources/RocksTransactionLogStore.ts:12` only persists the log's last-committed position in the `shutdown()` hook — skipped on any unclean exit.
- Result: replayed records exist in the store, but the transaction log's committed watermark still points before them. Replication's resume-read and `read_transaction_log` are both bounded by that watermark, so they never surface the recovered entries to a resuming peer.

## Evidence

Reproduced and measured 3 independent ways in `integrationTests/cluster/crashWindowReplication.test.mjs` (harper-pro, see PR link below):

1. 4–5 of 17–19 window writes durable on node A, never replicated to node B, across 3 runs.
2. Immediately after restart, A's own `read_transaction_log` returns **zero** of the stranded records. One fresh unrelated write to A advances the watermark past the entire recovered tail, and all of them appear at once and converge to B.
3. With `shutdown()` allowed to run (graceful restart instead of a crash), the same scenario is clean: 18/18 accepted, 18/18 in the log, 18/18 replicated.

## Why this isn't a quick JS-side fix

I prototyped advancing `TransactionLog._lastCommittedPosition` to the true end-of-log from JS after a replay that wrote anything. It did not work — the committed-position watermark is native, cross-thread state behind `_`-prefixed rocksdb-js internals, and `read_transaction_log` still returned nothing after the poke. A durable fix needs either:

- a rocksdb-js API to recover/advance the committed-position watermark after a `readUncommitted` replay, or
- committed-position recovery built into rocksdb-js's own boot/replay path.

rocksdb-js's `TransactionLog.getStats()` already computes `replayGapBytes` and `uncommittedTransactions`, so the gap is measurable today — a boot-time warning when replay detects an unresolved gap would be a cheap interim mitigation even before the real recovery fix lands.

## Related, not root cause

A resumed bulk copy plus this lost audit tail is a double miss: `replication/replicationConnection.ts:3467` resumes bulk copy at `afterKey` in key order, so any key sorting before that point is only ever delivered by the post-copy audit replay from `copyStartTime` — exactly the path this defect breaks. The code comment at that line already documents the copy-resume/audit-replay dependency; it has no defense today against a truncated committed view.

## Repro

`integrationTests/cluster/crashWindowReplication.test.mjs` (harper-pro), opt-in via `HARPER_TEST_CRASH_WINDOW=1` (skipped by default so CI stays green — this is a pin of a known-open defect, not a regression guard). Fails 3/3 opt-in runs with an explicit list of the stranded record IDs.

## Origin

Found while investigating an intermittent nightly failure in `blockCacheEviction.test.mjs` (harper-pro). That test itself had an unrelated race (restart-then-health-poll raced the still-running pre-restart process) which is fixed in HarperFast/harper-pro#618; fixing the race exposed this real defect underneath. Dispatch: `hp-blockcache-eviction-node22-flake`.

core submodule pointer at time of investigation: `cda8d63f` (2026-07-22). Confirmed `replayLogs.ts`, `RocksTransactionLogStore.ts`, and `auditStore.ts` are byte-identical between `cda8d63f` and `origin/main` — not already fixed by any change currently on harper main, including the resume-cursor fixes in Sync Core PR #610.

kriszyp:
Fixes [HarperFast/harper#1949 — Crash recovery restores records but not the transaction log's committed watermark, silently stranding them from replication](https://github.com/HarperFast/harper/issues/1949).

## The defect

After an unclean exit (crash, OOM kill, container kill), transaction-log entries that are durable on disk are invisible to every **committed** read — replication's resume replay and `read_transaction_log` — until some unrelated later commit happens to advance the watermark past the whole stale tail. A peer that resumes replication after the crash never receives writes that were already acknowledged and durable before it.

The two positions involved are easy to conflate, so, precisely:

- **`lastCommittedPosition`** is in-memory only and is never persisted. `commitFinished()`/`commitAborted()` advance it to "front of `uncommittedTransactionPositions`, else `nextLogPosition`". It is what bounds committed reads (`loadLastPosition()` in `transaction-log-reader.ts`).
- **`lastFlushedPosition`** lives in `txn.state` and is written by `databaseFlushed()` — the RocksDB `OnFlushCompleted` callback, so on *every* memtable flush. It means "the log up to here is already in SSTs", and it bounds replay (`startFromLastFlushed`) and purge eligibility.

The link is the bug: `getLastCommittedPosition()` lazily seeded the in-memory watermark **from `txn.state`** on first use. A graceful shutdown looks clean only because `shutdown()` forces a final flush, which writes `txn.state` at the true end of log — nothing ever persists the committed watermark itself. After a crash, `txn.state` sits at the last periodic flush, so the watermark is seeded behind the recovered tail and stays there: the consumer's boot replay re-applies the tail with `readUncommitted`, and those replayed writes deliberately don't re-append to the log (that would prevent replay convergence), so `commitFinished()` never runs and the watermark never moves.

## The fix

`load()` now seeds the watermark from the last entry that closed a transaction, using the active recovery scan and walking backward through rotated files until it finds a boundary or reaches the `txn.state` floor. The persisted `txn.state` flushed position is a safe floor, so recovery never hides entries already absorbed by RocksDB. Older-file probing is bounded, exception-safe, and closes files opened only for the scan. Active-file repair reads `txn.state` first and refuses to truncate below its same-file flushed offset, so a missing flag cannot erase history RocksDB already absorbed.

A store with no log files still leaves the position at `{0, 0}` for the existing lazy path to resolve.

## Tests

`test/transaction-log-crash-recovery.test.ts` (+ a SIGKILL fixture). Both fail on `main` and pass here:

- reopen after entries that no flush ever recorded: committed read returned **0 of 3**
- real SIGKILL, with a flush partway through: committed read returned **3 of 7** — the four post-flush entries stranded, the same shape as the harper-pro repro in the issue
- a transaction spanning four rotated files: the old one-predecessor scan seeded **0 of 1** committed entries; the backward scan recovers the prior boundary
- a flushed, uniform-timestamp tail with missing flags: the old repair shrank the log from **124 to 50 bytes**; the flushed floor now preserves all three entries

Also strengthened the two existing tail-recovery tests in `transaction-log.test.ts`. One carried the comment *"the committed position isn't persisted without a RocksDB flush, so read uncommitted to verify the entries survived on disk"* — this bug written down as an accepted workaround.

Local: 764 vitest passed / 56 files (2 skipped), 117 native gtests passed, production build and `pnpm check` passed.

## Worth a reviewer's attention

**Complete-but-uncommitted tail entries are now visible to committed readers.** The log entry is appended before the RocksDB commit, and `recoverTail()` only strips *torn* tails, so a crash can leave a structurally complete entry whose data commit never landed — the seed exposes it. Two things make that the right behavior rather than a new hazard:

1. It is the same exposure `commitAborted()` already produces at runtime: when a batch was written but the commit failed, `TransactionHandle::close()` calls it and the watermark advances past those bytes.
2. The consumer reconciles the tail by replaying it. In Harper, `replayLogs()` runs synchronously inside `readRocksMetaDb()` immediately after the database is opened and before the root store is returned, so no committed reader can observe the tail before replay has applied it.

The residual case is a replay that aborts early — Harper bounds replay on a no-progress counter (harper#1266) and a wall-clock limit (harper#1316). There the watermark advertises a tail the database doesn't fully have. Both abort paths already log `fatal` with "re-clone this node", so this seemed acceptable rather than worth gating the advance on replay completion, but it is a deliberate call and worth a second opinion.

## Recovery ends the log on a transaction boundary

Only a batch's final entry carries `TRANSACTION_LOG_ENTRY_LAST_FLAG`, so a crash partway through a multi-entry transaction leaves whole, well-framed entries that are merely a *prefix* of it. Left alone, the next transaction's flag closes the phantom group, merging two source transactions into one over replication. The PR closes that two ways, because either alone is incomplete:

1. **The committed watermark is seeded at the last complete transaction**, not at `nextLogPosition`. `scanTransactionLogForRecovery()` reports that boundary alongside `validEnd` from the same walk, so open-time recovery gets it without a second pass.
2. **`recoverTail()` discards the leftover entries** (`discardUnclosedTransaction`), so the file itself ends on a boundary. Seeding alone only holds until the next commit: the bytes stay on disk, the first `commitFinished()` after boot advances the watermark past them, and the merge happens then — one commit later rather than never.

Discarding is safe because nothing durable depends on those bytes. `writeBatch()` completes before `Transaction::Commit()` in every commit path (`transaction.cpp:236`/`286` async, `701`/`710` sync) and both commit-thread lanes preserve dispatch order, so an interrupted log write is always the newest thing in the log and its RocksDB commit never ran.

**The discard is gated, deliberately.** The last-entry flag only landed in `a69528c1` (2025-12-02); a log written before that has no flagged entries at all, and a naive "truncate to the last flagged entry" would wipe the whole file. So it requires both:

- a complete-transaction boundary **earlier in the same file** — the proof that this writer sets the flag; and
- **a single timestamp across the trailing run** — `writeEntriesV1()` stamps every entry of a batch with the batch timestamp while a prior flagged boundary prevents conflating separate transactions even if a caller reuses a timestamp, so more than one timestamp means more than one transaction went unflagged, which a flag-setting writer cannot produce.

Without that proof the bytes are kept and a warning is emitted. That is also what covers a batch `writeBatch()` split across a rotation (no flagged entry in the active file), which is why (1) is still needed: there the seed walks backward across every unflagged rotation until it finds a boundary or reaches the `txn.state` floor.

Both platforms physically truncate every torn tail the framing scan can identify. Windows drops its cached mapping before `SetEndOfFile`, uses the same path for torn and unclosed tails, and resets the timestamp index after either recovery shrink. A durable header followed by a partial payload remains indistinguishable from a valid zero-ending payload on a pre-extended Windows file; detecting that case requires an entry checksum.

**Behavior change worth noting:** a `readUncommitted` replay no longer sees the interrupted batch's prefix. That is intentional and arguably a second fix — Harper's `replayLogs()` currently applies those entries, i.e. writes a partial transaction into the database.

Covered by 13 GoogleTests on the pure scan (framing boundary + the unclosed-tail description that gates the discard) and two end-to-end tests: one truncates a real three-entry transaction to two and asserts the file is back at the boundary, `readUncommitted` sees 1 entry, and the following transaction's entries parse as flags `[1, 1]`; the other clears the flag on the last two entries of a three-transaction log (a pre-flag log's shape) and asserts nothing is dropped. Reverting just the source changes fails the first with `expected 124 to be 50`.

## Relationship to rocksdb-js #668

[#668 — Data loss: IsBusy commit retry orphans an uncommitted log position, pinning the committed watermark → committed reads truncate](https://github.com/HarperFast/rocksdb-js/issues/668) is the *runtime* sibling of this defect: there an orphaned entry in `uncommittedTransactionPositions` pins `lastCommittedPosition` as a permanent floor; here a stale `txn.state` seed sets that floor too low at boot. Same watermark, same silent-truncation symptom, different path — this PR does not fix #668.

Its "Suggested fix → Subtlety" section is worth reading alongside the design question above: it already names the same tension (`writeBatch` durably writes a transaction's bytes before the commit outcome is known, so advancing the watermark past them exposes bytes whose transaction never committed). Recovery now answers that structurally — the log cannot retain an unclosed transaction — while #668's runtime path can still strand one via an orphaned `IsBusy` position, and there the pin is permanent rather than cleared by the next commit. Its fix should probably adopt the same boundary rule rather than inventing a second one.

## Follow-up

Harper needs a `@harperfast/rocksdb-js` bump once this releases; `crashWindowReplication.test.mjs` in harper-pro can then drop its `HARPER_TEST_CRASH_WINDOW` opt-in and become a real regression guard.

---

Current-head review at `2a1833cb`: a full Gemini + Harper storage-domain review found that destructive tail repair ran before reading `txn.state` and could erase a committed, flushed tail. That major is fixed and regression-tested. A delta pass confirmed the fix; its remaining Human-Review-Need 4 items are the deliberate log-authoritative recovery decision Kris approved, inferred interrupted-tail deletion, and the documented Windows checksum gap. Cursor refused the earlier `AGENTS.md` diff and was policy-pruned from the low-risk delta; no Claude leg was used.

Cross-model review coverage at `ba0a59c7`: full passes by Claude, Gemini, and the Harper storage adjudicator, followed by a Claude + Gemini delta pass. Two review-found majors were fixed; the optional Cursor/Grok leg refused the `AGENTS.md` diff. Gemini's sole blocker — that `lastCommittedPosition` is a `std::optional` and the assignment is UB — did not survive a source check: it is a `std::shared_ptr<LogPosition>` always constructed via `make_shared`. Its two test findings (async `console.log` handshake lost to the SIGKILL; Windows self-kill surfacing as exit code 1) were real and are fixed.

Codex additionally caught that the seed's `fullPosition > 0` guard is unsound — that union member aliases the two uint32s as a `double`, so a `logSequenceNumber >= 0x7ff00000` reads back as NaN or negative and would have silently skipped the seed. The repo's own CI review bot independently flagged the same line. Now tested on `logSequenceNumber`. Its remaining finding was the transaction-boundary issue, now fixed here.

The Windows CI failure this PR initially hit was a genuine latent bug it exposed, not a test artifact: `findPositionByTimestamp()` checked for the zero-timestamp end-of-data marker *before* special-casing the header's own timestamp slot, so a log file with a zero header timestamp had its `size` corrected down below the 13-byte header and read as empty. Only Windows reached it at open (its `openFile()` indexes to undo memory-map zero-padding), but the indexing path is cross-platform and the added regression test reproduces it everywhere.

Generated with Claude Opus 5.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 027f0017399f</sub>

<sub>Human-Review-Need: 4 @ 027f0017399f</sub>


kriszyp:
I will note that https://github.com/HarperFast/rocksdb-js/pull/723/changes/24b12b943fd48b7b6d55ae4dfabeea3ac3d2e39b significantly expands this PR to more carefully avoid a partial transaction, but if you prefer it to be separated into a different PR, happy to do so.

kriszyp:
Pushed `1b0630bb` — recovery now **discards** the leftover entries of a transaction that never closed, instead of only holding the committed watermark short of them.

The reason for the follow-up: seeding the watermark at the last complete transaction is only true until the next commit. The bytes stayed in the file, the first `commitFinished()` after boot advanced the watermark past them, and the next batch's `TRANSACTION_LOG_ENTRY_LAST_FLAG` closed the phantom group — the same merge, one commit later. `recoverTail()` now ends the file on a boundary, which makes the invariant enforceable rather than merely observed.

Dropping the bytes is safe because `writeBatch()` completes before `Transaction::Commit()` in every commit path and both commit-thread lanes preserve dispatch order, so an interrupted log write is always the newest thing in the log and its RocksDB commit never ran.

The gating is the part most worth your eye: the last-entry flag only landed in `a69528c1` (2025-12-02), so a naive truncate-to-last-flagged-entry would wipe a pre-flag log wholesale. The discard therefore requires a boundary earlier in the same file *and* a single timestamp across the trailing run; anything else is kept and warned about, with the watermark seed still covering committed readers. Details and the platform split (POSIX truncates, Windows zero-fills) are in the updated description.

One intentional behavior change: a `readUncommitted` replay no longer sees the interrupted batch's prefix. In Harper that prefix was being applied by `replayLogs()`, i.e. a partial transaction written into the database, so this looks like a fix in its own right — but flagging it explicitly in case you read it the other way.

660 vitest / 51 files and 110 native gtests locally; the new end-to-end test fails without the source change (`expected 124 to be 50`).

— Claude Opus 5

src/binding/transaction_log/transaction_log_file_windows.cpp
```diff
@@ -458,6 +460,41 @@ bool TransactionLogFile::truncateFile(uint32_t newSize) {
...
+	// append (which writes at `size`, not at EOF) reuses the range.
+	constexpr uint32_t CHUNK = 64 * 1024;
+	std::vector<char> zeros(std::min(CHUNK, entriesEnd - newSize), 0);
+	for (uint32_t offset = newSize; offset < entriesEnd;) {
+		uint32_t chunk = std::min(static_cast<uint32_t>(zeros.size()), entriesEnd - offset);
+		int64_t written = this->writeToFile(zeros.data(), chunk, offset);
+		if (written <= 0) {
+			DEBUG_LOG("%p TransactionLogFile::eraseTail Failed to zero %u bytes at offset %u in %s (error=%lu)\n",
+				this, chunk, offset, this->path.string().c_str(), ::GetLastError());
+			return false;
```
cb1kenobi: **Medium: partial Windows `eraseTail` leaves `size` past the on-disk end-of-entries marker**

Zeros are written forward from `newSize` (the boundary). If a chunk write fails *after* at least one successful chunk — only reachable when the unclosed tail exceeds `CHUNK` (64 KiB) — the bytes at `[boundary, offset)` are already zeroed (so the first zero timestamp, i.e. the end-of-entries marker, now sits at `boundary`), but the function returns `false`. `discardUnclosedTransaction()` then bails without lowering `size`, so `size`/`nextLogPosition` stay at `entriesEnd`. The next append lands at `entriesEnd`, yet any reader scanning from the header stops at the zero at `boundary` — so a subsequently committed entry is silently invisible, the exact symptom class this PR fixes. (A first-chunk failure zeros nothing and degrades safely to the prior keep-the-tail behavior; only the mid-loop failure is inconsistent.)

Suggested fix: make the partial-failure path leave a consistent state — establish and flush the terminator at `boundary` first, then lower the logical `size` to `boundary` whenever any zeroing has occurred, so `size` can never exceed the first zero timestamp even when later chunks fail. Confidence is moderate: this requires a genuine mid-write I/O error during recovery on Windows with a >64 KiB unclosed tail, so it is narrow — but the impact if hit matches the bug being fixed.

—
_Generated by Barber AI_

kriszyp:
Re cb1kenobi's Windows `eraseTail` partial-failure finding: fixed in 52653446.

`eraseTail` now writes the 8-byte end-of-entries marker at `newSize` as its own write before the bulk zeroing loop. A reader stops at the first zero timestamp regardless of what follows it, so once that write lands, `newSize` is safe to use unconditionally as the new `size` — even if a later chunk in the bulk zero fails partway through the rest of the range. That closes the gap where `size` stayed at `entriesEnd` while the on-disk marker had already moved to `newSize`, which would have put the next append past where readers stop.

— KrAIs (Claude Sonnet 5)

kriszyp:
Fixed the Windows recovery-test failure in 05e20019.

The test now validates the platform-specific recovery contract: POSIX physically truncates to the last complete transaction, while Windows preserves the crash-time file length and zeroes the discarded range. Both paths also assert the parser's logical end is the completed-transaction boundary.

Local verification: `pnpm check`, focused crash-recovery tests (4/4), full JS suite (660 passed, 2 skipped), and native suite (110 passed). Final outside-model review found no blockers.

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -151,6 +151,31 @@ void TransactionLogFile::open(const double latestTimestamp) {
...
+		return 0;
+	}
+
+	uint32_t fileSize = this->size.load(std::memory_order_relaxed);
+	if (fileSize <= TRANSACTION_LOG_FILE_HEADER_SIZE) {
+		return 0; // header-only or empty: no transactions at all
+	}
+
+	std::vector<char> buffer(fileSize);
+	int64_t bytesRead = this->readFromFile(buffer.data(), fileSize, 0);
```
cb1kenobi, line 167: **Avoiding the whole-file read here (efficiency, non-blocking)**

Really nice, self-contained addition — the boundary-seed logic reads cleanly. One note on the read itself: the scan this feeds (`findLastCompleteTransactionEnd` → `scanTransactionLogForRecovery`) only ever reads the **13-byte entry header** (timestamp / length / flags) and then skips the body with `pos += HEADER + length`. It never dereferences a payload byte. So `std::vector<char> buffer(fileSize)` + a full `readFromFile` pulls the entire file — every value — into memory purely to jump over most of it.

A bounded header-walk would avoid that:

- `pread` the 13-byte header at `pos`, read `length`, advance `pos += 13 + length`, repeat — bounded by `this->size`.
- Memory drops from O(fileSize) to O(1) (one small header buffer, or a ~64 KB refill window to amortize syscalls on logs with many tiny entries).
- On logs with large (blob) values, you also skip faulting in the body pages entirely — real I/O savings, not just the allocation.

The fd is already in hand (`readFromFile` is a `pread` on `this->fd`), so no new handle is needed — this is purely swapping the byte source under the same framing logic. That's the "pass the fd instead of the buffer" idea in its lightest form.

Two honest caveats so it's easy to weigh:

- This is a **cold path** (older files, only the rare mid-transaction fallback), so it's an efficiency nicety, not a correctness issue — perfectly reasonable to defer.
- The only part that wants more than forward hops is `validFramingResumes` (the corruption resync), which needs a forward range scan — but that's corruption-only and can `pread` on demand.

`recoverTail` just below shares this exact pattern, so a small bounded-reader helper could serve both. More context in the summary comment.

*— Generated by Claude Opus 4.8*

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -174,8 +199,12 @@ void TransactionLogFile::recoverTail() {
 	}
 
 	RecoveryScan scan = scanTransactionLogForRecovery(buffer.data(), fileSize);
```
cb1kenobi, line 201: Same pattern as the new `scanForLastCompleteTransactionEnd`: the `buffer` here is a whole-file read (a couple lines up) feeding a header-only scan that never touches payload bytes. This read is **pre-existing** — nothing to change for this PR — I'm just flagging it as the natural sharing point: if the scan moves to a bounded header-walk / streaming reader, `recoverTail` and `scanForLastCompleteTransactionEnd` could use the same helper (both only need entry headers + `this->size`). Full write-up in the summary comment.

*— Claude Opus 4.8*

cb1kenobi:
## Follow-up thought: the whole-file-read-then-scan pattern

Really clean recovery work here — the committed-watermark seed walk and the `discardUnclosedTransaction` reasoning (flag proof + single-timestamp run) are easy to follow, and the union-member `logSequenceNumber` gotcha comment is a great catch to leave for the next reader.

While reading through it, one efficiency pattern stood out that's worth capturing. It's mostly **not blocking this PR** — only the first row below is new here — but the new code follows an existing shape that's worth a second look.

Several spots allocate the **entire file** and read every byte, then hand the buffer to a scan that only ever touches the **13-byte entry headers** and skips the bodies (`pos += HEADER + length`). None of these scans read payload bytes, so the full-file read pulls in all the values just to jump over them:

| Site | Reads | Header-only scan it feeds | In this PR? |
|---|---|---|---|
| [`scanForLastCompleteTransactionEnd`](https://github.com/HarperFast/rocksdb-js/blob/fe88e6b22a16ee3cd9dae96153e94a9ac534b363/src/binding/transaction_log/transaction_log_file.cpp#L166) | whole file → `std::vector<char>` | `findLastCompleteTransactionEnd` → `scanTransactionLogForRecovery` | **new here** |
| [`recoverTail`](https://github.com/HarperFast/rocksdb-js/blob/fe88e6b22a16ee3cd9dae96153e94a9ac534b363/src/binding/transaction_log/transaction_log_file.cpp#L193) | whole file → `std::vector<char>` | `scanTransactionLogForRecovery` | read pre-existing; fn modified here |
| [`countEntries`](https://github.com/HarperFast/rocksdb-js/blob/fe88e6b22a16ee3cd9dae96153e94a9ac534b363/src/binding/transaction_log/transaction_log_file.cpp#L340) | whole file → `std::vector<char>` | `countTransactionLogEntries` | pre-existing |
| [`validateTransactionLogImage` / `…Store`](https://github.com/HarperFast/rocksdb-js/blob/fe88e6b22a16ee3cd9dae96153e94a9ac534b363/src/binding/transaction_log/transaction_log_validation.cpp#L255) | whole file → `new char[]` | `scanTransactionLogForRecovery` + its own header walk | pre-existing (separate file) |

### Suggestion

Swap the byte source, keep the framing logic byte-identical: a bounded reader that `pread`s a 13-byte header, advances `pos += 13 + length`, and refills a ~64 KB window as needed — bounded by `this->size`. That gives O(buffer) memory instead of O(fileSize), and on logs with large blob values it skips faulting in the body pages entirely.

Since the scanners already take `(const char* data, uint32_t fileSize)`, the cleanest shape is a small reader abstraction (or the fd) threaded in behind that interface — i.e. "pass the fd instead of the buffer." For the recovery sites this is low-friction because `TransactionLogFile` already owns an open fd and `readFromFile` is already a `pread` on it.

### Wrinkles worth knowing before committing to it

- **The `const char*` interface is load-bearing for the native tests.** The GoogleTests call these scanners ~30 times with in-memory `LogImage` buffers, so a reader/fd signature change ripples through them — you'd likely want a buffer-backed reader shim so the pure functions stay testable without a real file on disk.
- **`validateTransactionLogImage` isn't a pure forward pass.** Its all-zero-tail check reads `[validEnd, fileSize)` *after* the framing walk, and it walks the headers a second time for per-entry anomalies. A `pread`/seek-capable reader handles both fine; a pure sequential stream would need to buffer or re-seek, and you'd probably want to fuse the two walks.
- **`this->size` is trustworthy as the bound at the recovery sites**, which is what makes dropping the `bytesRead != fileSize` full-read check safe: both scans run single-threaded at load before the store is published (no concurrent appends), and on Windows `openFile()` corrects the pre-extended/zero-padded size down before `recoverTail()` runs. A short header read or a `length` past `this->size` is exactly the torn-tail case the scanner already handles.

### Priority

Genuinely low for the recovery sites — they're cold, once-at-open, bounded by rotation size. The strongest case is `validateTransactionLogStore`, which reads *every* `.txnlog` in the directory on each `verify-logs` / `backups.verify()` call — but that's pre-existing and in a separate file, so it's best as its own PR.

Net: the whole-file read in the new `scanForLastCompleteTransactionEnd` (inline comment above) is the one worth a second look for this PR; the rest is an optional cleanup PR. Happy to prototype the bounded-reader helper if it'd be useful.

*— Generated by Claude Opus 4.8*

cb1kenobi (changes requested):

src/binding/transaction_log/transaction_log_file_windows.cpp
```diff
@@ -458,6 +460,58 @@ bool TransactionLogFile::truncateFile(uint32_t newSize) {
...
+	std::vector<char> zeros(std::min(CHUNK, entriesEnd - newSize), 0);
+	for (uint32_t offset = newSize; offset < entriesEnd;) {
+		uint32_t chunk = std::min(static_cast<uint32_t>(zeros.size()), entriesEnd - offset);
+		int64_t written = this->writeToFile(zeros.data(), chunk, offset);
+		if (written <= 0) {
+			DEBUG_LOG("%p TransactionLogFile::eraseTail Failed to zero %u bytes at offset %u in %s (error=%lu); "
+				"end-of-entries marker at %u is already committed, so the unzeroed remainder is unreachable dead "
+				"space that a future append will overwrite\n",
+				this, chunk, offset, this->path.string().c_str(), ::GetLastError(), newSize);
+			break;
```
cb1kenobi: **Low: a partially zeroed range is not "unreachable dead space" once appends grow past the failure offset**

The terminator-first ordering fixes the `size`/marker inconsistency from the previous round — `size` can no longer exceed the first zero timestamp. The residual is what the `break` leaves behind on disk. If zeroing fails at offset `F` (`newSize < F < entriesEnd`), the layout is: valid entries `[0, newSize)`, zeros `[newSize, F)`, **stale entry bytes `[F, entriesEnd)`**, pre-extension zeros after that. Appends resume at `newSize` and grow forward, so as soon as the write frontier passes `F` the bytes immediately after the newest entry are the middle of an old entry rather than zeros — a reader scanning for the end-of-entries marker parses them as a frame header and gets a bogus length or a corrupt frame, in the active log file. The DEBUG_LOG text asserts a future append will overwrite the remainder, but an append only overwrites the prefix of it; the part beyond the frontier stays.

Narrow: Windows, an unclosed tail larger than the 64 KiB chunk, and a `WriteFile` failure on an already-allocated (pre-extended) range. But it re-creates the same silent-truncation/corruption class this PR closes.

Suggested fix: on a chunk failure, fall back to physically truncating at `newSize` (`SetFilePointerEx` + `SetEndOfFile`) instead of `break`ing — `getMemoryMapLocked()` already re-extends lazily when a map request exceeds `this->size`, so the pre-extension model survives and no stale bytes can outlive the erase. Failing that, at least mark the file non-appendable so the store rotates rather than writing into the un-zeroed range, and correct the log text.

—
_Generated by Barber AI_

kriszyp: Fixed in ba0a59c7. Windows recovery no longer zero-fills this range: both unclosed-tail and torn-tail repair now drop the cached mapping and physically truncate with SetEndOfFile, so no stale suffix can become visible as appends advance. Recovery also resets the timestamp index after either shrink, and the Windows native tests cover physical truncation with a mapped file plus torn-tail index repair.\n\n— GPT-5 Codex

src/binding/transaction_log/transaction_log_file.cpp
kriszyp: Agreed that this is the natural sharing point for a future bounded reader. This pre-existing recovery read is unchanged here; combining both scans safely requires retaining the mid-file corruption resynchronization behavior, so I have recorded it as a separately scoped performance follow-up.\n\n— GPT-5 Codex

src/binding/transaction_log/transaction_log_file.cpp
kriszyp: Agreed on the cost and the bounded-header reader direction. I kept that refactor out of this correctness PR because it would also need to preserve corruption resynchronization semantics. The new fallback is now bounded to one predecessor, exception-safe, and closes a file opened only for the scan; the remaining whole-file read is recorded as a follow-up performance item.\n\n— GPT-5 Codex

src/binding/transaction_log/transaction_log_store.cpp
```diff
@@ -1094,6 +1091,55 @@ std::shared_ptr<TransactionLogStore> TransactionLogStore::load(
...
+	// pre-flag stores cannot turn startup into an unbounded full-history scan.
+	LogPosition recoveredPosition = { 0, 0 };
+	bool scannedOlderFile = false;
+	for (auto it = store->sequenceFiles.rbegin(); it != store->sequenceFiles.rend(); ++it) {
+		if (it->first > storeCurrentSeq) {
+			continue;
+		}
+		const bool isCurrent = it->first == storeCurrentSeq;
+		if (!isCurrent && scannedOlderFile) {
+			break;
```
cb1kenobi, line 1106: **Medium: limiting recovery to one predecessor can miss the real committed boundary**

`writeBatch()` can split one transaction across multiple rotations. If a crash leaves the active file and its predecessor with only unclosed entries, this `break` stops before reaching the earlier file that contains the last flagged boundary, so `recoveredPosition` stays `{0,0}` (or only `txn.state`). With no recent flush, committed entries before that large interrupted batch are again hidden from committed reads/replication after restart; a later commit can then advance past the retained unclosed tail.

Suggested fix: keep walking back until recovery either finds the first flagged boundary needed to seed the watermark or reaches a real bound such as the flushed `txn.state` floor; if that bound is unavailable, surface the multi-file unclosed tail as unrecoverable instead of silently seeding from `{0,0}`.

—
_Generated by Barber AI_

src/binding/transaction_log/transaction_log_file_windows.cpp
```diff
@@ -449,13 +449,37 @@ int64_t TransactionLogFile::writeToFile(const void* buffer, uint32_t size, int64
...
+		DEBUG_LOG("%p TransactionLogFile::truncateFile Refusing to truncate %s with an outstanding memory map\n",
+			this, this->path.string().c_str());
+		return false;
+	}
+
+	this->memoryMap.reset();
+	LARGE_INTEGER boundary;
+	boundary.QuadPart = newSize;
+	if (!::SetFilePointerEx(this->fileHandle, boundary, nullptr, FILE_BEGIN) ||
+		!::SetEndOfFile(this->fileHandle)) {
```
cb1kenobi, line 465: **Low: Windows still cannot *detect* the torn tail this now truncates — a torn entry with a durable header parses as complete**

The physical truncation itself is the right fix and it removes the partial-zeroing hazard from the last round entirely — no comment on that. What the header doc and AGENTS.md now assert, though, is that Windows and POSIX handle torn tails the same way, and the detection half of that isn't true.

A crash that leaves an entry's 13-byte header durable but its payload only partly durable (machine crash / power loss — a process kill leaves the page cache intact, so this is not the SIGKILL shape the tests cover) produces `header(ts, length=L, flags)` followed by pre-extension zeros. Then:

1. `findPositionByTimestamp()` reads a non-zero `ts`, advances `lastIndexedPosition += 13 + L`, finds the zero timestamp *past* the torn entry, and sets `this->size = entryStart + 13 + L`.
2. `recoverTail()` scans `[0, size)`. At `entryStart` the frame now fits **exactly** to EOF, so `scanTransactionLogForRecovery()` accepts it and returns `Clean` — `TruncateTail` is never reached.
3. If that entry was the batch's last and carries `TRANSACTION_LOG_ENTRY_LAST_FLAG`, `lastCompleteTransactionEnd` is set past it, so `load()` seeds the committed watermark past an entry whose payload tail is zeros. A committed reader (replication resume, `read_transaction_log`) then serves it as a valid entry.

On POSIX the same crash is caught: the file ends at the bytes actually written, so `pos + 13 + L > fileSize` and the frame is dropped as a torn tail. The divergence is the pre-extension padding, not the erase path — it is pre-existing, but the watermark seed added by this PR is what makes it observable to committed readers rather than hidden behind a stale `txn.state`.

Not suggesting this PR fix it. The concrete ask is to stop the docs from claiming parity — the `truncateFile`/`eraseTail` comments in `transaction_log_file.h` and AGENTS.md invariant 12 should say Windows truncates a torn tail *when the scan can see one*, and that a torn entry with a durable header is still absorbed as complete there. If it is worth closing later, a length/CRC check over the entry payload is the only thing that distinguishes "payload is zeros" from "payload happens to end in zeros" under a zero-padded file model.

—
_Generated by Barber AI_

src/binding/transaction_log/transaction_log_file.cpp
cb1kenobi: Follow-up in https://github.com/HarperFast/rocksdb-js/pull/792.

`recoverTail` and `scanForLastCompleteTransactionEnd` no longer allocate `vector(fileSize)` or `readFromFile` the whole image. `scanTransactionLogForRecovery` now walks through a positional reader (`pread` / `SetFilePointerEx`+`ReadFile`). The success path only pulls 13-byte entry headers and advances `pos += HEADER + length`, so payload bytes are never requested. Sequential headers within 64 KiB share a refill window; a jump past that (a large value) reads exactly one header. `findLastCompleteTransactionEnd()` is gone — last-complete comes off the same scan. `validFramingResumes` still does the corruption-only byte search, also through that reader rather than a buffer.

cb1kenobi (approved):
Awesome!

PR #725, opened 2026-07-27
fix(database): stop a dropped-column-family write from poisoning the environment

kriszyp:
## What

Sets `WriteOptions::ignore_missing_column_families` on the binding's **non-transactional**
writes, so a `putSync`/`removeSync` that names a dropped column family is discarded instead
of taking the whole RocksDB environment down with it. Transactions deliberately keep their
existing options — see [Why transactions are excluded](#why-transactions-are-excluded).

## Why

A single write naming a dropped column family fails inside RocksDB *after* the batch has
reached the WAL but *before* it reaches the memtables. RocksDB treats that gap as
unrecoverable inconsistency and latches a background error on the whole environment. From
that instant, **every write to every column family on that path — including column families
created afterwards — fails with the same, unattributable error**:

```
Invalid argument: Invalid column family specified in write batch
```

Reads keep working, so the database looks alive while silently accepting no writes. The latch
clears only when the environment is fully closed and reopened, which in production terms
means a node restart.

So one writer that loses a race against a drop takes down every unrelated writer sharing the
path. Harper hit exactly this: a test suite dropped a table while background writes to it
were still in flight, the stale write poisoned the environment, and 44 *unrelated* tests in
the same shard failed on a shared catalog store with an error naming nothing
(`HarperFast/harper` CI run 30237981148). Harper issues
[#1381](https://github.com/HarperFast/harper/issues/1381) and
[#843](https://github.com/HarperFast/harper/issues/843) both look like this same latch reached
by different paths.

This repo already knew about the behaviour — `test/drop.test.ts` pinned one assertion to last
position with a comment explaining that the failed write "contaminates the env's shared write
path". It was documented as test-ordering folklore rather than fixed.

### The fix

RocksDB has a first-class, documented option for this case:

> `ignore_missing_column_families` — If true and if user is trying to write to column families
> that don't exist (they were dropped), ignore the write (don't return an error). If there are
> multiple writes in a WriteBatch, other writes will succeed.

The binding simply wasn't setting it. So **no caller-side guard is needed** — the premise that
RocksDB requires callers to prevent writes to dropped column families turns out to be false.

`DBHandle` now builds its write options once in `open()` and exposes them as two accessors:

- `writeOptions()` — used by `Database::PutSync` and `Database::RemoveSync`. Sets the flag.
- `transactionWriteOptions()` — used to begin transactions. Does **not** set the flag.

That is the complete write surface. The binding issues exactly four RocksDB write calls
(`db->Put`, `db->Delete`, `txn->Put`, `txn->Delete`); the transactional pair inherits the
options `BeginTransaction` was given, and `resetTransaction()` — the commit-retry path — is
where those are built, so retries are covered. There is no `WriteBatch` and no separate async
write path (`DBI.put`/`remove` are `putSync`/`removeSync` wrapped in a promise). `clear()` uses
compaction plus `DeleteFilesInRange`; the custom transaction log writes its own files; backups
and checkpoints only flush and copy.

### Why transactions are excluded

Cross-model review flagged this and a direct experiment confirmed it, so it is worth stating
plainly: applying the flag to transactions **trades a contained failure for a silent one**.

A pessimistic transaction sends its staged batch straight through RocksDB's write path under
the transaction's own write options. Stage writes to a live column family and a dropped one,
and with the flag set the commit applies the live half, discards the dropped half, and
returns OK — a partial commit reported as success, which `Transaction::Commit` promises will
not happen, and which the custom transaction log would then mark committed. Measured on a
pessimistic database: without the flag the commit is rejected and nothing is applied; with it,
the commit resolves and the live key is readable.

Losing the whole transaction is the right outcome, so the transactional path is left alone.
Consequences, stated honestly:

- **Optimistic mode** (the default, and the only mode Harper uses) already fails safely:
  conflict validation rejects a commit naming a dropped family early, with
  `Invalid argument: Could not access column family <id>` — an error that *names* the family —
  and nothing is applied.
- **Pessimistic mode** still reaches the fatal path and poisons the environment on a commit
  naming a dropped family. That is a **pre-existing bug, unchanged by this PR**, tracked as
  [#726](https://github.com/HarperFast/rocksdb-js/issues/726). Fixing it properly means
  interlocking a drop against in-flight transactions that touch the family (real work, and
  its own design call) — not silently discarding half a transaction.

## Behaviour change

Confined to `putSync`/`removeSync`: a write through a handle whose column family has been
dropped is now silently discarded rather than throwing *and* poisoning the environment.
Nothing that worked before stops working, but a caller relying on the throw to detect a stale
handle no longer gets one. These writes carry a single key, so "discard the entries naming a
dropped family" discards the whole write — there is no partial state to leave behind.

A corollary for Harper: since optimistic transaction commits fail safely, the write that
actually poisoned the environment in the CI cascade was a **non-transactional**
`putSync`/`removeSync`, not the background cache *commit*. The
`Transaction commit failed: ... Invalid column family specified in write batch` lines in that
log are commits *reporting the already-latched* error, not causing it. Worth knowing when the
Harper-side trigger is scoped.

## Test notes

`pnpm test` — 660 passing, 2 skipped, 50 files. `pnpm check` clean. Only `test/drop.test.ts`
changed.

- **New:** *"should not poison the database when a stale handle writes to a dropped column
  family"*. Three handles on one path: an unrelated `victim` family, a `doomed` family, and a
  stale second handle to `doomed`. After the drop and the stale write, asserts `victim` still
  works via `putSync`, via `removeSync`, and via a transaction, and that a family opened
  *after* the poisoning write is writable. Covers a stale `removeSync` too, so the `RemoveSync`
  options are exercised independently of `PutSync`.
- **New:** *"should not partially apply a pessimistic transaction spanning a dropped column
  family"* — the guard that keeps the flag off the transactional path. Stages writes to a live
  and a doomed family, drops mid-transaction, and asserts the commit rejects **and** the live
  half was not applied.
- **New:** *"should reject only the dropped-family commit when transactions race a drop"* —
  characterization of optimistic mode: of three concurrent commits only the dropped-family one
  is rejected, its error matches `/Could not access column family \d+/`, the other two apply.
  Passes before and after; it pins behaviour the change must not break.
- **Changed:** the ghost-table test's last assertion
  (`expect(() => db2.putSync(...)).toThrow()`) becomes an assertion that the write is
  discarded, and the "must stay LAST" ordering constraint is gone — the reopen-by-name block
  now runs *after* the discarded write, additionally pinning that a recreated family stays
  writable.
- **Negative controls, both directions.** With `ignore_missing_column_families = false` on the
  non-transactional path, the poison tests fail. With it set to `true` on the *transactional*
  path, the pessimistic atomicity test fails. Neither test is vacuous.

## Review

Reviewed cross-model before opening (Gemini + Codex, adjudicated inline — the `reviewer`
subagent is not installed in this profile). Codex's blocker on pessimistic partial commits was
confirmed by experiment and is the reason for the two-accessor split; Gemini's build-the-options-
once point is applied. Findings not adopted, with reasons, are in the review comment below.

## Open question for reviewers

Should a non-transactional write through a dropped handle also *report* something rather than
vanishing? Adding a `dropped` flag to `ColumnFamilyDescriptor` (set in
`Database::Drop`/`DropSync`, checked in `PutSync`/`RemoveSync`) would close that, with the
RocksDB option still underneath as the guarantee that a missed path can never poison the
environment. Left out deliberately: it converts benign drop-vs-write races into errors callers
must handle, a separate design call from "stop the database dying". Happy to add it here.

## Follow-up in Harper

Harper picks this up on its next `@harperfast/rocksdb-js` bump — no Harper change is required.
Separately, Harper's `dropTable` does not quiesce in-flight writes before dropping a table's
column families, which is what triggers the race; that is tracked as its own task.

Refs [HarperFast/harper#1381](https://github.com/HarperFast/harper/issues/1381),
[HarperFast/harper#843](https://github.com/HarperFast/harper/issues/843)

Authored by Claude Sonnet 5.


kriszyp:
## Cross-model review — thorough

Legs: **Gemini** (`agy`, perf/maintainability lens) and **Codex** (`codex exec`, correctness/
concurrency lens, primed with a repo invariant checklist). Adjudicated inline.

*Caveat: adjudicated inline — the `reviewer` subagent is not installed in this profile, so
the isolated Opus domain pass did not run. Codex additionally reports that its own attempt to
reach an outside model from its read-only sandbox failed (`ENOTIMP`), so its findings are its
own analysis. Codex also observed the worktree changing mid-review and re-read live HEAD; its
findings were re-verified against the final state.*

### Blockers

- **[Codex] Pessimistic transactions could silently partially commit** —
  `src/binding/transaction/transaction_handle.cpp:114`. **Confirmed by experiment and fixed.**
  Applying `ignore_missing_column_families` to the options a transaction is begun with makes a
  commit spanning a live and a dropped column family apply the live half, discard the dropped
  half, and return OK. Measured on a pessimistic database: without the flag the commit is
  rejected and nothing is applied; with it, the commit resolves and the live key is readable.
  Fixed by splitting the accessors — `writeOptions()` (flag set) for `putSync`/`removeSync`,
  `transactionWriteOptions()` (flag unset) for `BeginTransaction` — with a regression test that
  fails if the flag is put back.

  Gemini independently raised the same hazard as "silent data loss and broken batch
  atomicity". Two models converging on it is why I ran the experiment rather than reasoning
  about it.

### Significant

- **[Codex] Tests were vacuous for two of the three changed paths** —
  `test/drop.test.ts`. Fixed. The `RemoveSync` option change is now exercised independently
  (a stale `removeSync` through the dropped handle), and the pessimistic
  stage → drop → commit sequence is now covered by the atomicity test above. Codex's point
  that the optimistic transaction test would still pass with the transaction change reverted
  is moot now that the transaction change is deliberately gone.

- **[Codex] The "write group" test could not form a write group** — `test/drop.test.ts`.
  Fixed by deleting the claim. Concurrent commits serialize through the database's single
  `commitWorker` lane, so `Promise.allSettled` proves nothing about leader-option inheritance.
  The test is kept as a characterization of optimistic-mode rejection, with an honest comment.

- **[Gemini] `writeOptions()` default-constructed and copied a struct per write** —
  `src/binding/database/db_handle.cpp`. Applied, with a caveat on the reasoning:
  `rocksdb::WriteOptions` is a trivial ~32-byte aggregate (6 bools, 2 enums, a `size_t`) with
  no allocations, so the cost is immaterial and was equally present before this PR — this is
  not the hot-path regression the finding describes. "Build it once when the handle opens" is
  the right shape regardless, so the options are now members initialised in `open()` and
  returned by const reference.

### Not adopted

- **[Gemini] "Incomplete write-option surface" (raised as a blocker)** — refuted. Gemini was
  reasoning from the diff alone and could not check. The binding issues exactly four RocksDB
  write calls (`db->Put`, `db->Delete`, `txn->Put`, `txn->Delete`); there is no `WriteBatch`,
  no separate async writer (`DBI.put`/`remove` wrap `putSync`/`removeSync` in a promise),
  `clear()` uses compaction plus `DeleteFilesInRange`, the custom transaction log writes its
  own files, and backups/checkpoints only flush and copy. Codex's independent cross-file audit
  reached the same conclusion and additionally confirmed `resetTransaction()` — the commit-retry
  path — is where the transaction's options are built, so retries are covered. The only
  remaining bare `WriteOptions()` is a native test fixture that cannot join the binding's write
  groups.

  The underlying concern is still right in principle, which is why the accessors are the only
  place write options are constructed.

### Coverage gaps

- Pessimistic mode still poisons the environment on a commit naming a dropped family. That is
  pre-existing and unchanged here, called out in the PR description, and needs a drop/in-flight-
  transaction interlock to fix properly. No test asserts the poisoning, deliberately — asserting
  a latched environment would make the test's own database unusable for anything after it.
- The genuine group-commit case (a merged write group with a differently-configured leader) is
  not testable from this API surface, since all commits on a database serialize through one
  commit worker. It is argued rather than tested: every binding writer reads from the same two
  accessors.

Reviewed by Claude Sonnet 5.

cb1kenobi:
Re-reviewed `b6a9f5fe` — no issues found. The only change since the prior review (`6b2b5616`) is CI/dependency config carried in from a `main` merge (Dependabot cooldown + pnpm `minimumReleaseAge`); the native fix and `test/drop.test.ts` are byte-identical, all four column-family write paths remain guarded, and there is no new use-after-free or dangling-handle risk. This PR looks good, nice job!

—
_Generated by Barber AI_

src/binding/database/database.cpp
```diff
@@ -1616,8 +1616,7 @@ napi_value Database::PutSync(napi_env env, napi_callback_info info) {
 			vtSlot = vt->slotFor(dbId, cfId, keySlice);
 			vtTracker = vt->lockSlotForWrite(vtSlot, dbId);
 		}
-		rocksdb::WriteOptions writeOptions;
-		writeOptions.disableWAL = (*dbHandle)->disableWAL;
+		const rocksdb::WriteOptions& writeOptions = (*dbHandle)->writeOptions();
```
cb1kenobi: Persisting both db-level and transaction-level write options in the `DBHandle` is unnecessary. If we add `writeOptions.ignore_missing_column_families = true;` in `PutSync()` and `RemoveSync()`, it solves the problem without adding complexity.

```suggestion
		rocksdb::WriteOptions writeOptions;
		writeOptions.disableWAL = (*dbHandle)->disableWAL;
		writeOptions.ignore_missing_column_families = true;
```

src/binding/database/database.cpp
```diff
@@ -1681,8 +1680,7 @@ napi_value Database::RemoveSync(napi_env env, napi_callback_info info) {
...
```
cb1kenobi:
```suggestion
		rocksdb::WriteOptions writeOptions;
		writeOptions.disableWAL = (*dbHandle)->disableWAL;
		writeOptions.ignore_missing_column_families = true;
```

cb1kenobi (changes requested):
Claude over-complicated the solution, but the test is great.

cb1kenobi (approved):

PR #732, opened 2026-07-28
Backport #717 to v2.4.x: resolve transaction reads against the caller's column family

cb1kenobi:
Backports [Resolve transaction reads against the caller's column family (#717)](https://github.com/HarperFast/rocksdb-js/pull/717) onto the `v2.4.x` release branch via cherry-pick.

## Commits
- `fix(transaction): resolve reads against the caller's column family`
- `fix(transaction): abort async get when the transaction is cancelled`
- `test(transaction): cover interleaved and three-way cross-CF reads`
- `fix(database): register async work for the async get`
- `Harden cross-column-family transaction reads`

## Notes
- The only conflict was in `AGENTS.md`: the merge context pulled in unrelated notes (the backup restore/shared-lock rewrite and the disk-space preflight item) that live on `main` but are not part of #717. Resolved by keeping the v2.4.0 text and adding only #717's new note, renumbered to item 8 to fit the v2.4.x list.
- No `package.json` version bump — the release script handles that.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -207,6 +207,21 @@ C++ code that needs to emit to JS without a database context should call
    (`list`, `verify`, a restore's source read) are not locked since concurrent readers are safe; a
    reader racing a `delete`/`purge` is a caller-managed hazard. Different directories are independent
    (separate lock files) and run fully in parallel.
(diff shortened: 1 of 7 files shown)


kriszyp (approved):
Thank you!

PR #733, opened 2026-07-29
fix(ci): don't tag older releases as latest on publish

cb1kenobi:
## Problem

The publish workflow tagged **every** non-prerelease as `latest`:

```yaml
${{ github.event.release.prerelease && 'next' || 'latest' }}
```

So publishing a backport/patch to an older line moved npm's `latest` tag **backward**. This happened for real with 2.4.1 — published while 2.5.0 was already the current `latest`, it stole the `latest` tag.

## Fix

Compute the dist-tag once in the `preflight` job and expose it as a job output that both publish steps (main package + platform bindings) consume:

- **prerelease** → `next`
- **newer than the current npm `latest`** (or genuine first publish) → `latest`
- **older stable release** → `release-<major>.<minor>` (e.g. `release-2.4`), so it stays installable via `pkg@release-2.4` while `latest` does not regress

### Reading the current latest

- Uses `npm view "$PACKAGE_NAME" dist-tags.latest`. `npm view` writes errors (incl. 404s) to **stderr**, so a failed lookup or a first publish leaves stdout empty.
- The result is validated as semver. If it's unreadable (transient registry error, or a first publish), the step **hard-fails** rather than silently defaulting to `latest` — guessing `latest` when the current latest is unknown is exactly the regression this change prevents. A genuine first publish is tagged manually.

## Note

This only fixes the automated path going forward. The already-published 2.4.1 → `latest` regression on npm is not corrected here; if `latest` should point back at the newest version, that's a one-time manual `npm dist-tag add @harperfast/rocksdb-js@<newest> latest`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,6 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.version.outputs.version }}
+      tag: ${{ steps.dist-tag.outputs.tag }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -57,6 +58,42 @@ jobs:
...
+          VERSION: ${{ steps.version.outputs.version }}
+          IS_PRERELEASE: ${{ github.event.release.prerelease }}
+        run: |
+          set -e
+          set -o pipefail
+          PACKAGE_NAME=$(node -p "require('./package.json').name")
+          if [ "$IS_PRERELEASE" = "true" ]; then
+            TAG=next
+          else
+            CURRENT_LATEST=$(npm view "$PACKAGE_NAME" dist-tags.latest 2>/dev/null || true)
```
kriszyp, line 74: This decision can be stale by the time `pnpm publish` runs. The workflow's concurrency key includes `github.ref`, so releases for different tags may run concurrently, and each then spends substantial time in the platform build matrix. For example, with `latest=2.5.0`, concurrent 2.6.0 and 3.0.0 runs both select `latest`; if 3.0.0 publishes first and 2.6.0 finishes last, npm ends at `latest=2.6.0`, reproducing the regression this change is meant to prevent. Please make promotion race-safe—for example, publish stable versions under a non-`latest` tag, then after publication recompute the highest published stable version and promote that version (including the binding packages), or otherwise serialize the compare-and-publish operation without dropping queued releases.

kriszyp (approved):
A minor note about timing, but this LGTM.

PR #734, opened 2026-07-29
Fix init-rocksdb skipping updates for pinned prerelease/revision versions

cb1kenobi:
## Summary

Pinning a downstream revision such as `11.1.2-1` in `package.json` (`rocksdb.version`) while `11.1.2` was already installed left the old version in place instead of downloading the pinned one.

**Root cause:** semver sorts a prerelease/revision suffix *below* the bare release — `11.1.2-1 < 11.1.2`. The second check in `scripts/init-rocksdb/main.ts` is a "don't downgrade" guard (so `ROCKSDB_VERSION=latest` won't clobber a newer manually-pinned build). With a pinned `11.1.2-1`, that guard saw `semver.lte('11.1.2-1', '11.1.2') === true` and skipped the download, treating the requested version as a downgrade. Note the *first* check (`semver.eq`) was already correct — it's not where the skip happened.

## What changed

Split the version decision into two intent-named pure predicates in a new `scripts/init-rocksdb/version-check.ts`:

- `installedSatisfiesPin` — the exact pin is already installed (full version match including the suffix, compatible runtime).
- `prebuildIsRedundant` — the downgrade guard, now applied **only** when resolving `latest` (or an unset version). An explicit pin always installs exactly what was requested.

`main.ts` now uses these instead of inline `semver` calls.

## Where to look

- `scripts/init-rocksdb/version-check.ts` — the new logic and the reasoning (see the doc comments on why a pin bypasses the downgrade guard).
- The behavior change: for an **exact pin**, the downgrade guard no longer applies, so a "sideways"/lower-sorting move like `11.1.2` → `11.1.2-1` now downloads. The `latest` path is unchanged.

## Also fixed (latent)

The old first check ran `semver.eq(currentVersion.version, 'latest')`, which **throws** `Invalid Version` when `ROCKSDB_VERSION=latest` and a prebuild is already installed. Gating it on "is an exact pin" avoids that.

## Tests

`test/init-rocksdb-version-check.test.ts` — 12 cases covering pin detection, exact-match/mismatch (incl. the `11.1.2` vs `11.1.2-1` regression), runtime matching, the `latest` downgrade guard, and an end-to-end assertion that pinning `11.1.2-1` over an installed `11.1.2` proceeds to download. Extracted as pure functions specifically so this is testable without network or `process.exit`.

## Notes for the reviewer

- Two related, **unaddressed** items I left out of scope (flag if you want them folded in): (1) `getPrebuild` sorts release tags with `semver.rcompare`, so if both `v11.1.2` and `v11.1.2-1` exist, the `latest` default still picks `11.1.2` over the "prerelease" `-1`; (2) the `version.h` fallback in `get-current-version.ts` can't represent a `-N` suffix (only MAJOR.MINOR.PATCH), so an old prebuild without `rocksdb.json` reports the bare version.
- Cross-model review was not run (the cross-model tooling wasn't available in this environment); worth a second set of eyes on the `latest`-vs-pin semantics.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: Claude Opus 4.8)

### DIFF
--- a/scripts/init-rocksdb/main.ts
+++ b/scripts/init-rocksdb/main.ts
@@ -13,11 +13,11 @@ import { buildRocksDBFromSource } from './build-rocksdb-from-source';
 import { downloadRocksDB } from './download-rocksdb';
 import { getCurrentVersion } from './get-current-version';
 import { getPrebuild } from './get-prebuild';
+import { installedSatisfiesPin, prebuildIsRedundant } from './version-check';
 import { config } from 'dotenv';
 import { readFileSync } from 'node:fs';
 import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import semver from 'semver';
 
 const __dirname = fileURLToPath(dirname(import.meta.url));
 
@@ -40,23 +40,21 @@ try {
 	const pkgJson = JSON.parse(readFileSync(resolve(__dirname, '../../package.json'), 'utf8'));
 	const desiredVersion = process.env.ROCKSDB_VERSION || pkgJson.rocksdb?.version || undefined;
 
-	if (
-		currentVersion &&
-		desiredVersion &&
-		semver.eq(currentVersion.version, desiredVersion) &&
-		(!currentVersion.runtime || currentVersion.runtime === runtime)
-	) {
-		console.log(`No update needed, RocksDB ${currentVersion.version} is already installed.`);
+	// Exact version pin already installed: nothing to do. This must match the
+	// full version including any downstream revision suffix (e.g. `11.1.2-1`),
+	// which semver treats as a prerelease of `11.1.2`.
+	if (installedSatisfiesPin(currentVersion, desiredVersion, runtime)) {
+		console.log(`No update needed, RocksDB ${currentVersion!.version} is already installed.`);
 		process.exit(0);
 	}
 
 	const prebuild = await getPrebuild(desiredVersion);
 
-	if (
-		currentVersion &&
-		semver.lte(prebuild.version, currentVersion.version) &&
-		(!currentVersion.runtime || currentVersion.runtime === runtime)
-	) {
+	// Downgrade guard, applied only when resolving `latest`: skip a redundant
+	// re-install or a downgrade. An explicit pin never reaches this — it must
+	// install exactly what was requested, even a revision suffix that semver
+	// ranks below the bare release.
+	if (prebuildIsRedundant(currentVersion, desiredVersion, prebuild.version, runtime)) {
 		console.log(`No update needed, latest version ${prebuild.version} is active.`);
 		process.exit(0);
 	}
--- a/scripts/init-rocksdb/version-check.ts
+++ b/scripts/init-rocksdb/version-check.ts
@@ -33,6 +33,11 @@ export function installedSatisfiesPin(
 	return (
 		isExactVersionPin(desiredVersion) &&
 		!!installed?.version &&
+		// Guard against a corrupted rocksdb.json or a non-semver ROCKSDB_VERSION:
+		// semver.eq throws on invalid input. Treating them as "not satisfied"
+		// falls through to getPrebuild, which fails with a clear "not found".
+		!!semver.valid(installed.version) &&
+		!!semver.valid(desiredVersion) &&
 		semver.eq(installed.version, desiredVersion) &&
 		runtimeMatches(installed, runtime)
 	);
@@ -55,6 +60,10 @@ export function prebuildIsRedundant(
 	return (
 		!isExactVersionPin(desiredVersion) &&
 		!!installed?.version &&
+		// See installedSatisfiesPin: guard invalid input so a corrupted version
+		// never throws. "Not redundant" falls through to a (re)download.
+		!!semver.valid(installed.version) &&
+		!!semver.valid(prebuildVersion) &&
 		semver.lte(prebuildVersion, installed.version) &&
 		runtimeMatches(installed, runtime)
 	);
--- a/scripts/init-rocksdb/version-check.ts
+++ b/scripts/init-rocksdb/version-check.ts
@@ -1,0 +1,69 @@
+import semver from 'semver';
+
+/** The version/runtime of the RocksDB prebuild currently installed in `deps/rocksdb`. */
+export type InstalledVersion = { version?: string; runtime?: string };
+
+/**
+ * Whether `desiredVersion` is an exact version pin (a concrete version from
+ * `package.json`'s `rocksdb.version` or the `ROCKSDB_VERSION` env var) rather
+ * than `latest` or unset.
+ *
+ * The distinction matters because a downstream revision suffix such as
+ * `11.1.2-1` is a *prerelease* to semver, which sorts it **below** the bare
+ * `11.1.2`. So a pinned `11.1.2-1` must be installed exactly as requested — the
+ * "don't downgrade" guard used for `latest` would otherwise refuse it, thinking
+ * the installed `11.1.2` is newer.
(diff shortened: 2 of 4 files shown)


scripts/init-rocksdb/version-check.ts
```diff
@@ -0,0 +1,69 @@
...
+	desiredVersion: string | undefined,
+	runtime: string | undefined
+): boolean {
+	return (
+		isExactVersionPin(desiredVersion) &&
+		!!installed?.version &&
+		semver.eq(installed.version, desiredVersion) &&
+		runtimeMatches(installed, runtime)
+	);
+}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in 2f80721f: guarded with `semver.valid` before comparing, so invalid input returns false instead of throwing.

scripts/init-rocksdb/version-check.ts
```diff
@@ -0,0 +1,69 @@
...
+	prebuildVersion: string,
+	runtime: string | undefined
+): boolean {
+	return (
+		!isExactVersionPin(desiredVersion) &&
+		!!installed?.version &&
+		semver.lte(prebuildVersion, installed.version) &&
+		runtimeMatches(installed, runtime)
+	);
+}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in 2f80721f: guarded with `semver.valid` before comparing, so invalid input returns false instead of throwing.

test/init-rocksdb-version-check.test.ts
```diff
@@ -0,0 +1,100 @@
...
+			// mismatched runtime → must re-download
+			expect(
+				installedSatisfiesPin({ version: '11.1.2-1', runtime: '-glibc' }, '11.1.2-1', '-musl')
+			).toBe(false);
+		});
+
+		it('is false when nothing is installed', () => {
+			expect(installedSatisfiesPin(undefined, '11.1.2-1', undefined)).toBe(false);
+			expect(installedSatisfiesPin({ version: undefined }, '11.1.2-1', undefined)).toBe(false);
+		});
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in 2f80721f: added invalid-semver test cases for this function.

test/init-rocksdb-version-check.test.ts
```diff
@@ -0,0 +1,100 @@
...
+		it('allows a newer "latest" resolve to download', () => {
+			expect(prebuildIsRedundant({ version: '11.1.2' }, 'latest', '11.1.3', undefined)).toBe(false);
+		});
+
+		it('does not block when nothing is installed or the runtime differs', () => {
+			expect(prebuildIsRedundant(undefined, 'latest', '11.1.2', undefined)).toBe(false);
+			expect(
+				prebuildIsRedundant({ version: '11.1.2', runtime: '-glibc' }, 'latest', '11.1.2', '-musl')
+			).toBe(false);
+		});
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in 2f80721f: added invalid-semver test cases for this function.

scripts/init-rocksdb/version-check.ts
```diff
@@ -0,0 +1,78 @@
...
+): boolean {
+	return (
+		isExactVersionPin(desiredVersion) &&
+		!!installed?.version &&
+		// Guard against a corrupted rocksdb.json or a non-semver ROCKSDB_VERSION:
+		// semver.eq throws on invalid input. Treating them as "not satisfied"
+		// falls through to getPrebuild, which fails with a clear "not found".
+		!!semver.valid(installed.version) &&
+		!!semver.valid(desiredVersion) &&
+		semver.eq(installed.version, desiredVersion) &&
```
kriszyp, line 41: `installedSatisfiesPin` promises exact pin identity, but `semver.eq()` deliberately ignores build metadata. With `rocksdb.json` recording `11.1.2+rev1` and an explicit `ROCKSDB_VERSION=11.1.2+rev2`, both inputs validate and this returns true, so initialization skips the requested release. Could we compare the validated recorded/tag identities (or use `semver.compareBuild`) and add a differing-`+metadata` test?

scripts/init-rocksdb/main.ts
```diff
@@ -13,11 +13,11 @@ import { buildRocksDBFromSource } from './build-rocksdb-from-source';
 import { downloadRocksDB } from './download-rocksdb';
 import { getCurrentVersion } from './get-current-version';
 import { getPrebuild } from './get-prebuild';
+import { installedSatisfiesPin, prebuildIsRedundant } from './version-check';
```
kriszyp, line 16: This extraction adds a build input that `prepare-rocksdb` does not track: `binding.gyp` currently lists only `main.ts`. The generated make rule therefore sees no changed prerequisite after a future helper-only edit, so `pnpm build:binding` skips initialization and continues with stale version-selection behavior. Please add `version-check.ts` to the action inputs; ideally the inputs should cover all imported init scripts plus `package.json`.

test/native/rocksdb_version_test.cc
```diff
@@ -24,7 +24,13 @@ TEST(RocksDBVersion, MatchesPackagePin) {
 
 	std::string expected = expectedVersion();
 	if (!expected.empty()) {
-		EXPECT_EQ(version, expected) << "Linked librocksdb version should match package.json rocksdb.version";
+		// package.json may pin a downstream prebuild revision suffix (e.g.
+		// "11.1.2-1") that RocksDB's own version.h does not encode. Compare
+		// against the base MAJOR.MINOR.PATCH, ignoring any "-N" suffix.
+		std::string expectedBase = expected.substr(0, expected.find('-'));
```
kriszyp, line 30: This avoids the false failure for a `-N` pin, but it also removes the only assertion of the prebuild revision. If a binary was built with `11.1.2-1`, the package pin changes to `11.1.2-2`, and `pnpm test:native` reuses the existing binary (as `run.mjs` currently does), both expectations reduce to `11.1.2` and the stale binary passes. Could the runner verify the exact installed/compiled prebuild revision before reusing the binary, while retaining this base `GetRocksVersionAsString()` check?

kriszyp (approved):
I've gotten nothing (human) to offer on trying battle cache poisoning and GH version pinning. Good luck.

scripts/init-rocksdb/version-check.ts
cb1kenobi: Addressed in #735 (commit 213c95a7): `installedSatisfiesPin` now uses `semver.compareBuild` instead of `eq`, so build metadata is part of the pin identity — an installed `11.1.2+rev1` no longer satisfies a pinned `11.1.2+rev2`. Added a differing-`+metadata` test. Landing via #735 rather than a new commit here (see PR note).

scripts/init-rocksdb/main.ts
cb1kenobi: Resolved in #735 by restructuring rather than expanding this list: that branch moves RocksDB provisioning to gyp **configure** time (`configure-rocksdb.mjs`, invoked via `<!@()`) and **removes the `prepare-rocksdb` build action entirely**, so there is no stale-input list to complete. (On this branch as-is the action still lists only `main.ts`; it goes away in #735.)

test/native/rocksdb_version_test.cc
cb1kenobi: Addressed in #735 (commit 213c95a7): `scripts/native-test/run.mjs` now records the pinned version alongside the built binary and rebuilds when it changes, so a stale binary across e.g. `11.1.2-1`→`11.1.2-2` is not reused. The base `GetRocksVersionAsString()` assertion is retained.

PR #735, opened 2026-07-30
Configurable RocksDB compression

cb1kenobi:
## Summary

Exposes RocksDB block/blob compression as a per-column-family option ([issue #201](https://github.com/HarperFast/rocksdb-js/issues/201)).

- **`compression` open option** — an algorithm name (`'lz4'`, `'zstd'`, `'zlib'`, `'none'`, …) or `{ algorithm, level }`. Applies to **both** SST block compression **and** blob-file compression (large values ≥ 2 KB become blobs, whose compression otherwise defaults to none).
- **Default: LZ4** (applied natively in `Database::Open`) when the build supports it, else RocksDB's own default. Confirmed & documented that RocksDB's stock default is Snappy; we override to LZ4.
- **`supportedCompression`** — module constant listing algorithms compiled into the build (from `GetSupportedCompressions()`). Opening with an unavailable algorithm throws.
- **`db.compression`** getter — reads the live value via `DB::GetOptions`.
- Native name↔enum mapping + supported list in the Node-free `core/compression.{h,cpp}` (GoogleTest-covered).

## Build robustness (dynamic lib linking)

`librocksdb.a` references, but doesn't bundle, the compression libraries it was built with. `binding.gyp` now links **exactly** the compression static libs the resolved prebuild ships, enumerated at configure time by `scripts/configure-rocksdb.mjs` (which also provisions the pinned prebuild). This means the build works against both an older `none+zlib` prebuild and a compression-enabled one — no version-bump coupling. To actually activate LZ4, a compression-enabled prebuild must be pinned in `package.json` (currently `11.1.2`); the feature is forward-compatible until then.

## Where to look / open items for the reviewer

- **Cold open applies the requested `compression` to every existing column family**, not just the one being opened (`DBDescriptor::open`). This is intended per the author, but worth a look — it means opening a multi-CF DB with a compression option changes all CFs' algorithm for subsequently-written files. A second *in-process* open that explicitly requests a **different** algorithm for an already-open CF now **throws** (`DBRegistry::OpenDB`); a plain reopen inherits the live setting (`DBOptions::compressionExplicit` gates this).
- **Block vs blob compression are unified** (same algorithm for both). Splitting them (e.g. lz4 blocks + zstd blobs) was deliberately deferred (YAGNI); the object form extends cleanly later.
- `scripts/configure-rocksdb.mjs` runs `init-rocksdb` at gyp **configure** time (before the `prepare-rocksdb` build action) so the enumerated lib set matches what will link. The **Windows** path (spawn + `AdditionalDependencies` splice) is unverified locally — worth watching on CI.

## Testing

- `test/compression.test.ts` — supported list, getter/defaults, validation, on-disk size comparison across multiple databases, already-open conflict cases, and a `normalizeCompression` guard suite.
- `test/native/compression_test.cc` — name↔type mapping + supported-list GoogleTests.
- `benchmark/compression.bench.ts` — RocksDB-only write/read throughput per available algorithm.

## Notes

- **Stacked on [#734 (init-rocksdb prerelease/revision version)](https://github.com/HarperFast/rocksdb-js/pull/734)** — based on that branch; retarget to `main` once #734 merges. The bottom three commits belong to #734.
- User-facing surface is documented in the repo `README.md` (this library's docs); no separate documentation-repo change needed.
- Generated by an LLM (Claude Opus 4.8). The code-review findings surfaced during development were addressed in the latest commit; the remaining open items are called out above.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -123,11 +123,6 @@
 					'link_settings': {
 						'libraries': [
 							'<(module_root_dir)/deps/rocksdb/lib/librocksdb.a',
-							# Compression libs the prebuild actually ships (zlib only on older
-							# prebuilds; snappy/lz4/zstd/bzip2 too on compression-enabled ones),
-							# enumerated by scripts/rocksdb-link-libs.mjs. Spliced after
-							# librocksdb.a because GNU ld is order-sensitive (it references their
-							# symbols).
 							'<@(rocksdb_compression_libs)'
 						]
 					},
@@ -172,7 +167,6 @@
 							],
 							'AdditionalDependencies': [
 								'rocksdb.lib',
-								# Compression libs the prebuild ships (see rocksdb-link-libs.mjs).
 								'<@(rocksdb_compression_libs)'
 							]
 						}
@@ -199,7 +193,6 @@
 							'AdditionalDependencies': [
 								# 'rocksdbd.lib',
 								'rocksdb.lib',
-								# Compression libs the prebuild ships (see rocksdb-link-libs.mjs).
 								'<@(rocksdb_compression_libs)'
 							]
 						}
@@ -279,7 +272,6 @@
 							],
 							'AdditionalDependencies': [
 								'rocksdb.lib',
-								# Compression libs the prebuild ships (see rocksdb-link-libs.mjs).
 								'<@(rocksdb_compression_libs)'
 							]
 						}
@@ -291,11 +283,6 @@
 					'link_settings': {
 						'libraries': [
 							'<(module_root_dir)/deps/rocksdb/lib/librocksdb.a',
-							# Compression libs the prebuild actually ships (zlib only on older
-							# prebuilds; snappy/lz4/zstd/bzip2 too on compression-enabled ones),
-							# enumerated by scripts/rocksdb-link-libs.mjs. Spliced after
-							# librocksdb.a because GNU ld is order-sensitive (it references their
-							# symbols).
 							'<@(rocksdb_compression_libs)'
 						]
 					},
@@ -334,7 +321,6 @@
 							],
 							'AdditionalDependencies': [
 								'rocksdb.lib',
-								# Compression libs the prebuild ships (see rocksdb-link-libs.mjs).
 								'<@(rocksdb_compression_libs)'
 							]
 						}
@@ -355,7 +341,6 @@
 							],
 							'AdditionalDependencies': [
 								'rocksdb.lib',
-								# Compression libs the prebuild ships (see rocksdb-link-libs.mjs).
 								'<@(rocksdb_compression_libs)'
 							]
 						}
--- a/binding.gyp
+++ b/binding.gyp
@@ -6,16 +6,17 @@
 	#   ROCKSDB_ASAN=1 node-gyp rebuild
 	'variables': {
 		"rocksdb_asan%": "<!(node -p \"process.env.ROCKSDB_ASAN==='1'?1:0\")",
-		# Compression libraries the resolved RocksDB prebuild actually ships (as
-		# `-l` flags / `.lib` names, not absolute paths — see configure-rocksdb.mjs
-		# for why), resolved against the library search dir configured in each
-		# target's link settings. Older prebuilds ship only zlib; a compression-
-		# enabled prebuild adds snappy/lz4/zstd/bzip2. Linking exactly what is
-		# present lets both build (a missing lib errors "no such file"; an unlinked
-		# one the prebuild needs errors "undefined symbols"). The script also
-		# provisions the pinned prebuild first. The script path is quoted so a
-		# repo checkout under a path with spaces still launches it.
-		'rocksdb_compression_libs': ['<!@(node "<(module_root_dir)/scripts/configure-rocksdb.mjs")'],
+		# RocksDB link libraries — the core `librocksdb` archive plus the compression
+		# libs the resolved prebuild actually ships — emitted as `-l` flags / `.lib`
+		# names (not absolute paths — see configure-rocksdb.mjs for why), resolved
+		# against the library search dir configured in each target's link settings.
+		# Older prebuilds ship only zlib; a compression-enabled prebuild adds
+		# snappy/lz4/zstd/bzip2. Linking exactly what is present lets both build (a
+		# missing lib errors "no such file"; an unlinked one the prebuild needs
+		# errors "undefined symbols"). The script also provisions the pinned prebuild
+		# first. The script path is quoted so a repo checkout under a path with
+		# spaces still launches it.
+		'rocksdb_link_libs': ['<!@(node "<(module_root_dir)/scripts/configure-rocksdb.mjs")'],
 	},
 	# Node 26's Windows headers (common.gypi) inject Clang ThinLTO options
 	# (-flto=thin and /opt:lldltojobs=N) into every Release target. The official
@@ -123,12 +124,19 @@
 					'cflags+': ['-fexceptions'],
 					'cflags_cc+': ['-fexceptions'],
 					'link_settings': {
+						# All link libraries (core librocksdb + the compression libs) come
+						# from configure-rocksdb.mjs as `-l` tokens, not absolute paths,
+						# resolved against this search dir. The search dir is module-relative
+						# (not <(module_root_dir)/...) so GYP rebases it to a ../-only path
+						# under the build dir: an absolute path carrying a spaced
+						# module_root_dir is emitted unquoted into LDFLAGS/LIBS and the link
+						# shell splits it, while a ../-relative path has no space to split
+						# (fixes a checkout under e.g. /tmp/rocks db).
 						'library_dirs': [
-							'<(module_root_dir)/deps/rocksdb/lib'
+							'deps/rocksdb/lib'
 						],
 						'libraries': [
-							'<(module_root_dir)/deps/rocksdb/lib/librocksdb.a',
-							'<@(rocksdb_compression_libs)'
+							'<@(rocksdb_link_libs)'
 						]
 					},
 					'xcode_settings': {
@@ -171,8 +179,7 @@
 								'<(module_root_dir)/deps/rocksdb/lib'
 							],
 							'AdditionalDependencies': [
-								'rocksdb.lib',
-								'<@(rocksdb_compression_libs)'
+								'<@(rocksdb_link_libs)'
 							]
 						}
 					}
@@ -196,9 +203,7 @@
 								'<(module_root_dir)/deps/rocksdb/lib'
 							],
 							'AdditionalDependencies': [
-								# 'rocksdbd.lib',
-								'rocksdb.lib',
-								'<@(rocksdb_compression_libs)'
+								'<@(rocksdb_link_libs)'
 							]
 						}
 					},
@@ -276,8 +281,7 @@
 								'<(module_root_dir)/deps/rocksdb/lib'
 							],
 							'AdditionalDependencies': [
-								'rocksdb.lib',
-								'<@(rocksdb_compression_libs)'
+								'<@(rocksdb_link_libs)'
 							]
 						}
 					}
@@ -286,12 +290,19 @@
 					'cflags+': ['-fexceptions'],
 					'cflags_cc+': ['-fexceptions'],
 					'link_settings': {
+						# All link libraries (core librocksdb + the compression libs) come
+						# from configure-rocksdb.mjs as `-l` tokens, not absolute paths,
+						# resolved against this search dir. The search dir is module-relative
+						# (not <(module_root_dir)/...) so GYP rebases it to a ../-only path
+						# under the build dir: an absolute path carrying a spaced
+						# module_root_dir is emitted unquoted into LDFLAGS/LIBS and the link
+						# shell splits it, while a ../-relative path has no space to split
+						# (fixes a checkout under e.g. /tmp/rocks db).
 						'library_dirs': [
-							'<(module_root_dir)/deps/rocksdb/lib'
+							'deps/rocksdb/lib'
 						],
 						'libraries': [
-							'<(module_root_dir)/deps/rocksdb/lib/librocksdb.a',
-							'<@(rocksdb_compression_libs)'
+							'<@(rocksdb_link_libs)'
 						]
 					},
 					'xcode_settings': {
@@ -328,8 +339,7 @@
 								'<(module_root_dir)/deps/rocksdb/lib'
 							],
 							'AdditionalDependencies': [
-								'rocksdb.lib',
-								'<@(rocksdb_compression_libs)'
+								'<@(rocksdb_link_libs)'
 							]
 						}
 					}
@@ -348,8 +358,7 @@
 								'<(module_root_dir)/deps/rocksdb/lib'
 							],
 							'AdditionalDependencies': [
-								'rocksdb.lib',
-								'<@(rocksdb_compression_libs)'
+								'<@(rocksdb_link_libs)'
 							]
 						}
 					},
--- a/binding.gyp
+++ b/binding.gyp
@@ -6,13 +6,16 @@
 	#   ROCKSDB_ASAN=1 node-gyp rebuild
 	'variables': {
 		"rocksdb_asan%": "<!(node -p \"process.env.ROCKSDB_ASAN==='1'?1:0\")",
-		# Compression static libs the resolved RocksDB prebuild actually ships,
-		# enumerated at configure time. Older prebuilds ship only zlib; a
-		# compression-enabled prebuild adds snappy/lz4/zstd/bzip2. Linking exactly
-		# what is present lets both build (a missing lib errors "no such file"; an
-		# unlinked one the prebuild needs errors "undefined symbols"). The script
-		# also provisions the pinned prebuild first (see scripts/configure-rocksdb.mjs).
-		'rocksdb_compression_libs': ['<!@(node <(module_root_dir)/scripts/configure-rocksdb.mjs)'],
+		# Compression libraries the resolved RocksDB prebuild actually ships (as
+		# `-l` flags / `.lib` names, not absolute paths — see configure-rocksdb.mjs
+		# for why), resolved against the library search dir configured in each
+		# target's link settings. Older prebuilds ship only zlib; a compression-
+		# enabled prebuild adds snappy/lz4/zstd/bzip2. Linking exactly what is
+		# present lets both build (a missing lib errors "no such file"; an unlinked
+		# one the prebuild needs errors "undefined symbols"). The script also
+		# provisions the pinned prebuild first. The script path is quoted so a
+		# repo checkout under a path with spaces still launches it.
+		'rocksdb_compression_libs': ['<!@(node "<(module_root_dir)/scripts/configure-rocksdb.mjs")'],
 	},
 	# Node 26's Windows headers (common.gypi) inject Clang ThinLTO options
 	# (-flto=thin and /opt:lldltojobs=N) into every Release target. The official
@@ -120,6 +123,9 @@
 					'cflags+': ['-fexceptions'],
 					'cflags_cc+': ['-fexceptions'],
 					'link_settings': {
+						'library_dirs': [
+							'<(module_root_dir)/deps/rocksdb/lib'
+						],
 						'libraries': [
 							'<(module_root_dir)/deps/rocksdb/lib/librocksdb.a',
 							'<@(rocksdb_compression_libs)'
@@ -280,6 +286,9 @@
 					'cflags+': ['-fexceptions'],
 					'cflags_cc+': ['-fexceptions'],
 					'link_settings': {
+						'library_dirs': [
+							'<(module_root_dir)/deps/rocksdb/lib'
+						],
 						'libraries': [
 							'<(module_root_dir)/deps/rocksdb/lib/librocksdb.a',
 							'<@(rocksdb_compression_libs)'
--- a/binding.gyp
+++ b/binding.gyp
@@ -40,6 +40,7 @@
 			],
 			'sources': [
 				'src/binding/binding.cpp',
+				'src/binding/core/compression.cpp',
 				'src/binding/core/debug.cpp',
 				'src/binding/core/platform.cpp',
 				'src/binding/core/file_lock.cpp',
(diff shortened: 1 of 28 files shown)


src/binding/database/database.cpp
```diff
@@ -1384,6 +1408,40 @@ napi_value Database::Open(napi_env env, napi_callback_info info) {
...
+	// compression: a friendly algorithm name (e.g. "lz4", "zstd", "none"). The
+	// TypeScript layer normalizes the string|object public option and validates
+	// against the supported list before we get here; this is the defensive
+	// backstop for a name that isn't recognized or isn't compiled in. When the
+	// caller does not specify one, default to LZ4 where the build supports it
+	// (otherwise leave RocksDB's own default). The default is not marked
+	// explicit, so a plain reopen of an already-open column family does not
+	// conflict with its live algorithm (see DBRegistry::OpenDB).
+	std::string compressionName;
+	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, options, "compression", compressionName));
```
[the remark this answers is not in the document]
cb1kenobi: `getProperty`'s 5th argument is `required` (default `false`), not `optional` — see `napi/helpers.h`. With the default, an omitted/`undefined` `compression` returns `napi_ok` and leaves the name empty, so the database opens fine (covered by the default-compression tests). Passing `true` would make it *required* and break opening without a compression option. No change needed.

src/binding/database/database.cpp
```diff
@@ -1384,6 +1408,40 @@ napi_value Database::Open(napi_env env, napi_callback_info info) {
...
+		double compressionLevel = 0;
+		if (rocksdb_js::getProperty(env, options, "compressionLevel", compressionLevel, true) == napi_ok) {
+			if (std::isnan(compressionLevel) || compressionLevel != std::trunc(compressionLevel) ||
+				compressionLevel < -2147483648.0 || compressionLevel > 2147483647.0
+			) {
+				::napi_throw_error(env, nullptr, "compressionLevel must be a 32-bit integer");
+				return nullptr;
+			}
+			dbHandleOptions.compressionLevel = static_cast<int>(compressionLevel);
+		}
```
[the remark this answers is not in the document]
cb1kenobi: Same flag semantics: the `true` here is `required=true` (`napi/helpers.h`), so an omitted `compressionLevel` returns `napi_invalid_arg`, the `== napi_ok` check is false, and the block is skipped — `compressionLevel` stays `std::nullopt` and RocksDB keeps its per-algorithm default. It is not treated as level 0. No change needed.

scripts/configure-rocksdb.mjs
```diff
@@ -0,0 +1,78 @@
...
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const isWin = process.platform === 'win32';
+const root = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+const libDir = join(root, 'deps', 'rocksdb', 'lib');
+
+const candidates = isWin
+	? ['snappy.lib', 'lz4.lib', 'zstd.lib', 'bz2.lib', 'zs.lib']
+	: ['libsnappy.a', 'liblz4.a', 'libzstd.a', 'libbz2.a', 'libz.a'];
```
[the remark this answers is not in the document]
cb1kenobi: `zs.lib` is the actual zlib static library shipped in the HarperFast Windows prebuild (verified — the archive contains zlib 1.3.2 `deflate`/`inflate`). The candidate list is existence-checked (`files.has(name)`), so a name that isn't present is simply skipped. No change needed.

scripts/configure-rocksdb.mjs
```diff
@@ -0,0 +1,78 @@
...
+
+const result = spawnSync(
+	join(root, 'node_modules', '.bin', 'tsx'),
+	[join(root, 'scripts', 'init-rocksdb', 'main.ts')],
+	{
+		cwd: root,
+		stdio: ['ignore', 'ignore', 'inherit'],
+		shell: isWin,
+	}
+);
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in 5f50b248: now spawns the current node executable against tsx's CLI entry (`tsx/dist/cli.mjs`) directly, dropping `shell` entirely — no `.cmd` wrapper, identical across platforms.

src/binding/database/db_descriptor.cpp
```diff
@@ -835,6 +836,17 @@ std::shared_ptr<DBDescriptor> DBDescriptor::open(const std::string& path, const
 	cfOptions.write_buffer_size = static_cast<size_t>(options.writeBufferSize);
 	cfOptions.max_write_buffer_number = options.maxWriteBufferNumber;
 	cfOptions.max_write_buffer_size_to_maintain = options.maxWriteBufferSizeToMaintain;
+	if (options.compression) {
+		cfOptions.compression = *options.compression;
```
kriszyp: This does not remain per-column-family: the resulting `cfOptions` is reused for every existing CF below. For example, create CF `a` with `none` and CF `b` with `zstd`, close all handles, then open `a` first with `none`; `DBDescriptor::open` opens both CFs with `none`. Opening `b` with `zstd` then hits the registry conflict, while a plain open silently inherits `none`, so first-open order controls every CF. Please preserve/build options per CF and apply the request only to `options.name` (for example, load per-CF options and override the target, or use `SetOptions` for RocksDB's dynamic compression/blob fields). Add a same-path, two-CF close/reopen regression. The same path should also apply or explicitly reject a changed level; the current registry check compares only the algorithm and silently accepts a different level.

binding.gyp
```diff
@@ -4,7 +4,16 @@
...
-	'variables': { "rocksdb_asan%": "<!(node -p \"process.env.ROCKSDB_ASAN==='1'?1:0\")" },
+	'variables': {
+		"rocksdb_asan%": "<!(node -p \"process.env.ROCKSDB_ASAN==='1'?1:0\")",
+		# Compression static libs the resolved RocksDB prebuild actually ships,
+		# enumerated at configure time. Older prebuilds ship only zlib; a
+		# compression-enabled prebuild adds snappy/lz4/zstd/bzip2. Linking exactly
+		# what is present lets both build (a missing lib errors "no such file"; an
+		# unlinked one the prebuild needs errors "undefined symbols"). The script
+		# also provisions the pinned prebuild first (see scripts/configure-rocksdb.mjs).
+		'rocksdb_compression_libs': ['<!@(node <(module_root_dir)/scripts/configure-rocksdb.mjs)'],
```
kriszyp: This still breaks the repo-path-with-spaces case the helper is trying to support. GYP runs this command through a shell, so `node /path with spaces/scripts/configure-rocksdb.mjs` is split before the script starts. On POSIX the script also returns absolute library paths, and `<!@()` splits those with `shlex.split`, turning `/path with spaces/deps/.../liblz4.a` into multiple library entries. A checkout under `C:\Users\Jane Doe\...` or `/tmp/rocks db` therefore fails during configure/link. Please preserve argument boundaries (for example, use GYP's array command form or quote the script path) and emit module-relative library paths rather than whitespace-bearing absolute paths, with a configure smoke test under a spaced temp path.

src/store.ts
```diff
@@ -74,13 +75,94 @@ export type CompactOptions = {
...
+	}
+
+	let algorithm: string;
+	let level: number | undefined;
+
+	if (typeof option === 'string') {
+		algorithm = option;
+	} else if (typeof option === 'object' && option.algorithm !== undefined) {
+		algorithm = option.algorithm;
+		level = option.level;
```
kriszyp: `level` is copied without runtime validation. A JavaScript caller can pass `{ algorithm: 'zstd', level: '9' as any }` or `null`; native `getProperty(..., "compressionLevel", ...)` then returns a non-OK status, but `Database::Open` only compares that status and silently proceeds with the algorithm's default level. The explicit option appears accepted while being ignored. Please validate that a provided level is a finite int32 integer in `normalizeCompression`, retain a throwing native backstop, and add malformed-level tests.

kriszyp (changes requested):
Awesome, this will be great!
One question I had: it looks like `compressionLevel` is partly wired up, but I don't see it all the way through and documented? Or am I missing something?
Anyway, GPT has comments too; I'd think the column family separation is most important.

src/binding/database/db_descriptor.cpp
cb1kenobi: Fixed in b1de03ae. `DBDescriptor::open` now reads each existing CF's persisted compression via `LoadLatestOptions` and opens each CF with its own value, applying the request only to `options.name` and only when explicit (the LZ4 default never overrides an existing CF, so a plain reopen inherits the live setting). The `DBRegistry::OpenDB` already-open check now compares **algorithm and level**. Added a two-CF close/reopen regression test (`per-column-family` in test/compression.test.ts).

binding.gyp
cb1kenobi: Fixed in b1de03ae. `configure-rocksdb.mjs` now emits whitespace-free tokens — `-l:libX.a` (Linux) / `-lX` (macOS) / `X.lib` (Windows), never absolute paths — resolved via a single `library_dirs` entry (one gyp-quoted token, space-safe), and the script path in binding.gyp is quoted. Added a smoke test asserting every emitted token is whitespace-free and well-formed.

src/store.ts
cb1kenobi: Fixed in b1de03ae. `normalizeCompression` now coerces a numeric-string level and throws for anything that isn't a finite int32; `Database::Open` gained a throwing native backstop (a present-but-wrong-type level is rejected, not silently dropped). Added malformed-level tests.

cb1kenobi:
Thanks @kriszyp — all three inline comments addressed in b1de03ae; you were right on the column-family separation.

On your `compressionLevel` question — it's now wired end-to-end and documented:
- **Input**: `{ algorithm, level }` → `normalizeCompression` coerces/validates it as a finite int32 (throws otherwise) → native applies it to `compression_opts.level`, with a throwing backstop in `Database::Open` for a present-but-wrong-type value.
- **Read-back**: `db.compression` now returns `{ algorithm, level? }` (the `level` is included only when a non-default level is set), so a configured level round-trips.
- **Conflict**: the already-open-CF reject now compares level too, not just algorithm.
- **Docs**: README (`compression` option, `db.compression` getter, the Compression section's per-CF/level notes) and the AGENTS.md design note updated.

Per-CF summary: `DBDescriptor::open` preserves each CF's persisted compression (`LoadLatestOptions`) and applies the request only to the target CF — a cold open of one CF no longer restamps the others, and a two-CF close/reopen regression test guards it. Full suite green (697 passed); native GoogleTests green.

<sub>Generated by Claude Opus 4.8</sub>

binding.gyp
```diff
@@ -111,13 +123,12 @@
 					'cflags+': ['-fexceptions'],
 					'cflags_cc+': ['-fexceptions'],
 					'link_settings': {
+						'library_dirs': [
```
kriszyp, line 135: The configure helper now launches correctly and emits whitespace-free compression tokens, but a spaced POSIX checkout still does not build. With `module_root_dir=/tmp/rocksdb-js gyp-space-check`, node-gyp 13 generates raw `-L/tmp/rocksdb-js gyp-space-check/deps/rocksdb/lib` and `/tmp/rocksdb-js gyp-space-check/deps/rocksdb/lib/librocksdb.a` entries, which the link shell splits. Configure also fails earlier at `deps/gtest.gyp:8`, whose two `module_root_dir` paths are still unquoted (`/bin/sh: /tmp/rocksdb-js: not found`). The new unit test invokes only the helper from the existing unspaced checkout, so it cannot catch either failure. Please keep these link inputs module-relative (or otherwise make GYP emit shell-safe arguments), quote/refactor the gtest initializer, and add a real configure/link smoke test from a spaced temporary checkout.

src/binding/database/db_descriptor.cpp
```diff
@@ -20,6 +24,61 @@ static uint64_t nextVtEpoch() {
...
+	std::unordered_map<std::string, PersistedCompression> result;
+	rocksdb::ConfigOptions configOptions;
+	// Be permissive: we only read compression fields, so unknown/unsupported
+	// options in the persisted file must not fail the load.
+	configOptions.ignore_unknown_options = true;
+	configOptions.ignore_unsupported_options = true;
+	rocksdb::DBOptions loadedDbOptions;
+	std::vector<rocksdb::ColumnFamilyDescriptor> loadedCfDescriptors;
+	rocksdb::Status status =
+		rocksdb::LoadLatestOptions(configOptions, path, &loadedDbOptions, &loadedCfDescriptors);
```
kriszyp, line 77: `LoadLatestOptions` is the only authoritative source used here for each CF's persisted compression, but every non-OK status is discarded as an empty map. For example, create `plain=none` and `compressed=zstd`, close them, remove the latest OPTIONS file (the RocksDB database itself is still openable), then reopen `plain`: `ListColumnFamilies` succeeds, this load returns `NotFound`, and `compressed` is opened with the base default instead of zstd. A corrupt or unparseable OPTIONS file takes the same path. That reintroduces silent cross-CF restamping on future files. Please fail the open with the RocksDB status (or recover another validated authoritative options source) and add missing/corrupt-OPTIONS regressions rather than falling back after an existing DB was already detected.

src/binding/database/db_descriptor.cpp
```diff
@@ -20,6 +24,61 @@ static uint64_t nextVtEpoch() {
...
+	rocksdb::ColumnFamilyOptions& cfOptions,
+	rocksdb::CompressionType compression,
+	const std::optional<int>& level
+) {
+	cfOptions.compression = compression;
+	// Large values are stored in blob files (enable_blob_files), which have their
+	// own compression that defaults to none; apply the same algorithm so the
+	// whole dataset is compressed, not just the inline (< min_blob_size) portion.
+	cfOptions.blob_compression_type = compression;
+	if (level) {
```
kriszyp: `cfOptions` can already contain a persisted algorithm-specific level, but this helper leaves it untouched when the caller explicitly selects an algorithm without a level. A CF persisted as zstd level 19 and cold-reopened with `{ compression: 'zlib' }` therefore becomes zlib level 19 instead of zlib's default. The warm reused-handle check has the same semantic gap because it compares a level only when one was provided. Since the API/native comments say omission uses the selected algorithm's default, please set `compression_opts.level` to `kDefaultCompressionLevel` when explicit compression omits a level, compare that effective request in the registry, and add cold/warm cross-algorithm regressions.

src/binding/database/db_registry.cpp
```diff
@@ -333,10 +334,35 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
...
+			// The column family is already open in this process (the DBDescriptor
+			// is process-global and shared across handles/envs). Compression is
+			// fixed per column family at creation, so a second open explicitly
+			// asking for a different algorithm or level cannot take effect on the
+			// reused handle — reject it rather than silently ignore the request. A
+			// plain reopen (compression defaulted, not explicit) inherits the live
+			// setting and skips this check.
+			rocksdb::ColumnFamilyHandle* cf = columns[name]->column.get();
+			rocksdb::Options current = entry.descriptor->db->GetOptions(cf);
+			bool algorithmDiffers = current.compression != *options.compression;
```
kriszyp, line 358: The reuse check compares only `current.compression`, even though this API promises to apply the requested algorithm to both SSTs and blobs and the new cold-open path deliberately preserves those fields independently. A legacy CF can have `compression=snappy` and `blob_compression_type=none`; after it is opened plainly, a second handle/worker explicitly requesting `snappy` passes this check, but values at the 2KB blob threshold remain uncompressed. Please also treat a live blob-algorithm mismatch as a conflict (or apply it dynamically), and add a seeded legacy-mismatch regression.

src/store.ts
```diff
@@ -74,13 +75,113 @@ export type CompactOptions = {
...
+	if (typeof option === 'string') {
+		algorithm = option;
+	} else if (typeof option === 'object' && option.algorithm !== undefined) {
+		algorithm = option.algorithm;
+		if (option.level !== undefined && option.level !== null) {
+			// TypeScript types `level` as a number, but coerce defensively (e.g. a
+			// numeric string from loosely-typed config) and reject anything that is
+			// not a finite 32-bit integer rather than letting the native layer
+			// silently drop it.
+			const coerced = Number(option.level);
```
kriszyp: `Number(option.level)` broadens the documented `level?: number` contract and prevents the stricter native backstop from seeing malformed types: `true`, `false`, `''`, whitespace, `[]`, `[6]`, and numeric strings are all accepted as levels, while a present `null` is silently treated as omitted. That still turns config mistakes into unintended or ignored compression tuning. Please require `typeof option.level === 'number'` before the finite/integer/int32 checks, reject present `null`, and cover representative non-number inputs.

scripts/native-test/run.mjs
```diff
@@ -35,12 +34,28 @@ const binaryName =
...
+// an earlier revision (e.g. 11.1.2-1 when the pin is now 11.1.2-2) would still
+// pass. Record the pinned version alongside the binary and force a rebuild when
+// it changes, so the runner never reuses a binary built for a different pin.
+const versionMarker = join(root, 'build', config, '.rocksdb-test-version');
+const builtVersion = existsSync(versionMarker) ? readFileSync(versionMarker, 'utf8').trim() : '';
+
+if (
+	!existsSync(binary) ||
+	process.env.NATIVE_TEST_REBUILD === '1' ||
+	builtVersion !== expectedVersion
```
kriszyp: This marker still does not identify the RocksDB artifact that node-gyp actually built against. `init-rocksdb/main.ts` selects `ROCKSDB_VERSION` (including the `.env` value it loads with `override: true`) before `package.json`, but this runner compares and writes only `pkg.rocksdb.version`. For example, with a package pin of `11.1.2-2`, build once under `ROCKSDB_VERSION=11.1.2-1`: the `-1` archive is linked but the marker records `-2`. Removing the override makes the next `pnpm test:native` see an apparently current marker and reuse `-1`; `MatchesPackagePin` also passes because it intentionally ignores the `-N` revision. Please derive the marker and `ROCKSDB_EXPECTED_VERSION` from the same effective build selection (including runtime/source identity), or record the installed `deps/rocksdb/rocksdb.json` identity after successful configure, and add an override-change regression around the runner's rebuild decision.

kriszyp (approved):
This is great. I am including the codex comments for your review, but I'm good with this stuff.

binding.gyp
cb1kenobi: Great catch — the earlier fix was indeed incomplete. Fixed in 20a9b7de:

- `deps/gtest.gyp` now quotes its command paths (and launches tsx via `node <cli.mjs>`), so the parse-time `<!()` no longer shell-splits at the space.
- `binding.gyp`'s POSIX library search dir is now module-relative, so gyp rebases it to `-L$(srcdir)/…` — a `../`-only path with no space to split.
- `librocksdb` is now linked as a per-platform `-l` token via `configure-rocksdb.mjs` (same mechanism as the compression libs) rather than an absolute file path, since gyp rebases `library_dirs` but passes a bare library *file* path through literally.

Verified end-to-end: I reproduced the original `exit 127` / undefined-symbol failures from a spaced checkout, then built **both** the `.node` and the native-test binary from `/tmp/rocks db …` (and unspaced) and ran them. Replaced the helper-only check with a real spaced-path `node-gyp configure` smoke test.

— Claude Opus 4.8

src/binding/database/db_descriptor.cpp
cb1kenobi: Good catch. Fixed in 19367c7f: `loadPersistedCompression` now returns the RocksDB status, and for an **existing** DB a non-OK `LoadLatestOptions` (missing/corrupt OPTIONS) fails the open with that status rather than falling back to defaults — the fallback would reopen the non-target CFs with the base default and silently restamp them. Added a regression that strips the `OPTIONS-*` files from an existing DB (still openable via the MANIFEST) and asserts the reopen throws.

— Claude Opus 4.8

src/binding/database/db_descriptor.cpp
cb1kenobi: Fixed in 19367c7f. `applyCompression` now resets `compression_opts.level` to `kDefaultCompressionLevel` when an explicit algorithm omits a level, so a zstd-level-19 CF cold-reopened as zlib no longer inherits 19. The registry warm-reopen check compares the **effective** requested level too (omitted == default, not "inherit"). Added both cold and warm cross-algorithm/level regressions.

— Claude Opus 4.8

src/binding/database/db_registry.cpp
cb1kenobi: Fixed in 19367c7f: the reuse check now also rejects a live blob-algorithm mismatch (`current.blob_compression_type != requested`), so a legacy CF opened plainly with `block=snappy`/`blob=none` isn't silently reused with uncompressed blobs by a later explicit `snappy` open.

One honest caveat: I couldn't build a seeded regression through the public API on this (LZ4-enabled) build — this layer always couples block and blob compression, so a live `block≠blob` CF only arises from externally-authored options. The guard is therefore defensive. Happy to add a lower-level fixture if you'd like one.

— Claude Opus 4.8

src/store.ts
cb1kenobi: Fixed in 19367c7f. `normalizeCompression` no longer lets a bare `Number()` coerce loose inputs: `true`/`false`/`''`/whitespace/`[]`/`[6]`/objects are now rejected. Per the earlier decision to keep string coercion, a number **or a non-blank numeric string** is still accepted (`'6'` and `'  6  '` → 6), and the throwing native int32 backstop remains. `null`/`undefined` stay "omitted → default". Added representative non-number rejection cases.

— Claude Opus 4.8

scripts/native-test/run.mjs
cb1kenobi: Fixed in 19367c7f. The runner now derives both the rebuild marker and `ROCKSDB_EXPECTED_VERSION` from the same effective selection init-rocksdb uses — `.env` override → `ROCKSDB_VERSION` → package pin, plus the Linux runtime and `ROCKSDB_PATH` source identity — not the pin alone. So a build made under `ROCKSDB_VERSION=11.1.2-1` is recorded as `-1`, and dropping the override flips the identity and forces a rebuild. Extracted the resolution into a pure `resolveBuildSelection()` with override-change unit tests.

— Claude Opus 4.8

PR #739, opened 2026-07-31
Fix: compression option as an empty object now behaves like unset instead of throwing

cb1kenobi:
## Summary

Opening a database with `compression` set to an empty object (or an object whose `algorithm` is `undefined`/`null`) threw a `TypeError`, inconsistent with `compression: undefined`, which defers to the native default (LZ4 when the build supports it). `normalizeCompression` now treats an object without an `algorithm` the same as an unset option and returns `{}`. Arrays and other non-object, non-string values still throw.

Since the normalized result is empty, such an open is also non-explicit for the second-open consistency check — same as `undefined`.

## Notes for reviewers

- A `level` on an object with no `algorithm` (e.g. `{ level: 3 }`) is now silently ignored rather than throwing — there is no explicit algorithm to tune, and this shape typically comes from loosely-typed config.
- Cross-model review (Codex) flagged that the public `CompressionOption` type still requires `algorithm`, so TS callers can't pass `{}` without a cast. Left as-is deliberately: the strict type steers TS users away from the ignored-`level` shape while the runtime stays lenient for JS/loose config. Happy to loosen the type if preferred.

Generated by Claude (Fable 5).

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -132,9 +132,9 @@ export function normalizeCompression(option: CompressionOption | undefined): {
 		algorithm = option;
 	} else if (typeof option === 'object' && !Array.isArray(option)) {
 		if (option.algorithm === undefined || option.algorithm === null) {
-			// An object without an algorithm (e.g. `{}` from loose config) is
-			// treated the same as an unset option; any `level` is ignored since
-			// there is no explicit algorithm to tune.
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Thanks!

PR #740, opened 2026-08-01
Add a bottommost option to compact() so existing data can be re-encoded

kriszyp:
`compactRange` passed a default-constructed `CompactRangeOptions`, whose `bottommost_level_compaction` is `kIfHaveCompactionFilter`. With no compaction filter installed that skips the bottommost level — where the bulk of the data sits — so `compact()` could not rewrite it.

That matters because a column family's `compression` governs **newly written files only**. Existing SST and blob files keep the codec they were written with until something rewrites them, and nothing available could. Changing the codec on a database that already held data left that data encoded as before, indefinitely.

Measured on 60k compressible records written uncompressed and reopened as zstd:

```
written uncompressed:            34.8M
reopened as:                     {"algorithm":"zstd"}
after plain compact():           34.8M   <-- bottommost skipped
after compact({bottommost:true}) 2.0M    <-- re-encoded
compactSync({bottommost:true})   2.0M    (idempotent)
```

```typescript
// Re-encode everything already on disk under the column family's current codec
await db.compact({ bottommost: true });
db.compactSync({ bottommost: true });
```

Opt-in rather than the default because it rewrites the entire range regardless of whether RocksDB judges it worthwhile, so it costs as much as the data is large — it should be a deliberate operator action, not something an upgrade or a routine compaction does implicitly.

## Why this is needed

`harper` [#2044](https://github.com/HarperFast/harper/pull/2044) makes Harper honor `storage.compression` on databases upgraded from 5.1, which had been silently ignored — an existing column family inherits its persisted codec on reopen rather than adopting the build default, so those instances were writing uncompressed indefinitely. That fix compresses new writes, which is the right default. This gives operators the separate, explicit action to convert the data already on disk; without it there is no path from an existing uncompressed database to a compressed one short of a dump and reload.

## Changes

- `db_descriptor.cpp` / `.h` — `compactRange` takes a `bottommost` flag and sets `BottommostLevelCompaction::kForce`
- `database.cpp` / `.h` — the flag is read from an added trailing argument on both `Compact` and `CompactSync`, and carried on `AsyncCompactState`
- `store.ts` — `CompactOptions.bottommost?: boolean`, forwarded to both native calls
- `load-binding.ts` — native signatures updated
- README — documents the option, and adds a note under **Compression** that a codec change affects new files only, and that omitting `compression` on an existing family inherits its codec rather than applying the default

## Testing

`test/compaction.test.ts` gains a case that writes uncompressed, reopens under zstd, asserts a plain `compact()` leaves the SST bytes **exactly** unchanged, then asserts `{ bottommost: true }` drops them below half — through both `compact()` and `compactSync()`. Skipped when the native build lacks zstd. 39 tests pass across the compaction and compression suites.

Verified against a RocksDB prebuild carrying the compression codecs (11.1.2); the repo's currently vendored prebuild has only none/zlib, so the new test self-skips there.

Generated by Claude Opus 5.

### DIFF
--- a/test/compaction.test.ts
+++ b/test/compaction.test.ts
@@ -12,7 +12,10 @@ function fileBytes(dir: string, extension: string): number {
 		recursive: true,
 	})) {
 		if (entry.isFile() && entry.name.endsWith(extension)) {
-			total += statSync(join(entry.parentPath ?? dir, entry.name)).size;
+			// `parentPath` was added in Node 20.11/21.5; README documents Node 18+ support, where
+			// this is `undefined` and the property was still named `path`.
+			const parentPath = entry.parentPath ?? (entry as unknown as { path?: string }).path ?? dir;
+			total += statSync(join(parentPath, entry.name)).size;
 		}
 	}
 	return total;
--- a/test/compaction.test.ts
+++ b/test/compaction.test.ts
@@ -4,19 +4,22 @@ import { readdirSync, statSync } from 'node:fs';
 import { join } from 'node:path';
 import { afterEach, describe, expect, it } from 'vitest';
 
-/** Total bytes of SST files under `dir` — what a codec change actually moves. */
-function sstBytes(dir: string): number {
+/** Total bytes of files with `extension` under `dir` — what a codec change actually moves. */
+function fileBytes(dir: string, extension: string): number {
 	let total = 0;
 	for (const entry of readdirSync(dir, {
 		withFileTypes: true,
 		recursive: true,
 	})) {
-		if (entry.isFile() && entry.name.endsWith('.sst')) {
+		if (entry.isFile() && entry.name.endsWith(extension)) {
 			total += statSync(join(entry.parentPath ?? dir, entry.name)).size;
 		}
 	}
 	return total;
 }
+
+const sstBytes = (dir: string) => fileBytes(dir, '.sst');
+const blobBytes = (dir: string) => fileBytes(dir, '.blob');
 
 describe('Compaction', () => {
 	afterEach(() => {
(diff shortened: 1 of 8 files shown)


cb1kenobi:
Reviewed `e2168272` — no issues found. This PR looks good, nice job!

Confirmed the native path: `kForceOptimized` forces bottommost compaction of pre-existing files (only skipping files the same run produced), the forced blob GC with `age_cutoff = 1.0` extends re-encoding to blob files, and the `bottommost=false` path is byte-for-byte the prior default-constructed `CompactRangeOptions` — no regression to the existing `compact()`. Threading (worker-thread execute callback, error propagated via `state->status`), memory safety, the opt-in default, and input validation all check out, and the SST + multi-generation blob tests (with value-integrity assertions) cover both entry points.

—
_Generated by Barber AI_

test/compaction.test.ts
```diff
@@ -139,6 +161,126 @@ describe('Compaction', () => {
...
 		}));
 
+	// RocksDB's CompactRangeOptions defaults bottommost_level_compaction to
+	// kIfHaveCompactionFilter, so with no compaction filter installed a plain compact() leaves the
+	// bottommost level alone — and that is where the bulk of the data ends up. Since a changed
+	// compression algorithm only governs newly written files, that level has to be rewritten for
+	// existing data to be re-encoded, which is what `bottommost: true` requests.
+	it('should re-encode existing data under a new codec only when bottommost is requested', () =>
+		dbRunner(async ({ db, dbPath }) => {
+			if (!supportedCompression.includes('zstd')) return;
```
cb1kenobi: If `zstd` is not supported, the test appears as passed. Instead we should use `it.skipIf` so that we can tell the test was actually skipped.

You will likely need to run `pnpm:fmt` to clean up the formatting.

```suggestion
	it.skipIf(!supportedCompression.includes('zstd'))('should re-encode existing data under a new codec only when bottommost is requested', () =>
		dbRunner(async ({ db, dbPath }) => {
```

test/compaction.test.ts
```diff
@@ -139,6 +161,126 @@ describe('Compaction', () => {
...
+		}));
+
+	// Values at or above the binding's 2048-byte blob threshold (`min_blob_size`) are stored in
+	// blob files, not SSTs. A bottommost SST compaction alone does not touch them: blob GC's
+	// default age cutoff only reclaims the oldest fraction of blob files, so `bottommost: true`
+	// forces blob GC across the full age range (see DBDescriptor::compactRange) to re-encode them
+	// too. Without that, this test's blob files would stay on the old codec.
+	it('should re-encode existing blob-backed data under a new codec when bottommost is requested', () =>
+		dbRunner(async ({ db, dbPath }) => {
+			if (!supportedCompression.includes('zstd')) return;
```
cb1kenobi: Same thing here:

```suggestion
		it.skipIf(!supportedCompression.includes('zstd'))('should re-encode existing blob-backed data under a new codec when bottommost is requested', () =>
		dbRunner(async ({ db, dbPath }) => {
```

cb1kenobi (changes requested):

cb1kenobi (approved):

Issue #695, opened 2026-07-08
Env-cleanup hooks can miss a descriptor mid-close/mid-reopen (registry reachability + dangling entry reference)


kriszyp:
Follow-up from the #694 cross-model review (adjudicated significant-narrow; shared with the pre-existing listener cleanup path, not introduced by #694).

The module env-cleanup hooks (`DBRegistry::RemoveListenersByEnv`, `DBRegistry::ReleaseCommitCompletionsByEnv`) find shared descriptors by walking `instance->databases[*].descriptor`. Two windows make a live descriptor unreachable from that map:

1. `OpenDB`'s wait predicate does `entry.descriptor.reset()` when it observes a closing descriptor (`db_registry.cpp`), dropping the registry's ref while the closer is still inside `finishClose()`. A worker env tearing down in that window is not scrubbed from the descriptor; the backstop is `finishClose()`'s own release pass (which runs after the commit thread is joined), but that pass can then call `napi_release_threadsafe_function` on a tsfn whose creating env has concurrently finished teardown — a narrow UAF.
2. Related pre-existing hazard in the same protocol: `OpenDB`'s waiting thread holds `auto& entry = entryIterator->second` (a reference into the map node) across the condition wait, while the closer's guarded erase can erase that node (the guard erases when `!eraseIt->second.descriptor`, which is exactly the state OpenDB's predicate creates). The woken waiter then evaluates its predicate against freed node storage.

Candidate categorical fix: give the registry a secondary set of all live descriptors (weak_ptr), maintained under `databasesMutex`, that the env-cleanup hooks walk instead of the entry map — descriptor reachability then no longer depends on entry-map lifecycle. The `entry` dangling reference wants a re-find-after-wake (or shared_ptr'd entry values) in `OpenDB`.

#694 already narrowed the surface: `DestroyDB` now closes before unlinking (CloseDB's discipline).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

kriszyp:
Reproduced this deterministically while triaging a Windows Bun CI failure on #742 (`db.setCompression()`) — the CI job failed with `Worker exited unexpectedly` right after the entire Vitest suite (all 722 tests) passed, i.e. during the single Bun process's final teardown, with no test-level failure. That symptom (crash after everything already passed) pointed at a process-lifecycle issue rather than anything in #742's diff.

[...]

PR #742, opened 2026-08-01
feat: add db.setCompression() for live per-CF compression changes

kriszyp:
## Summary

RocksDB's `DB::Open()` opens every existing column family in one call, so a caller can never re-specify compression for a column family that's already been transitively opened without closing every other handle sharing that process-global `DBDescriptor`. This blocks Harper's boot sequence, which necessarily opens its default CF (transitively opening every other CF) before it can read its own catalog CF to learn each table's desired codec — see the linked issue for the full repro.

Adds `db.setCompression(algorithm, level?)`, backed by `rocksdb::DB::SetOptions()`, which RocksDB's own headers document as dynamically mutable for exactly the fields needed (`ColumnFamilyOptions::compression`, `blob_compression_type`). Verified against the linked RocksDB build with a standalone C++ probe (not just assumed) — see the PR's self-review for the full writeup, including why this was chosen over a per-CF options map on the root open call (the alternative doesn't fit Harper's actual use case: it doesn't know a table's codec until *after* reading its own catalog CF, which is itself opened by that same call).

An independent cross-model review (codex + gemini + grok + a Harper-domain adjudicator) on the initial implementation surfaced real native-layer gaps — no in-flight operation registration (a shutdown-race UAF hazard matching AGENTS.md's documented teardown discipline), no read-only guard, no handling for a split live/durable state on an OPTIONS-persist failure, and an unbounded per-call cost given `SetOptions()` is RocksDB's own documented "slow, infrequent-use" call. All are fixed (second commit) — see the self-review for details.

## Rebase + follow-up hardening (this update)

Rebased onto latest `main` (fixed a genuine conflict where an earlier commit on this branch had accidentally deleted unrelated README sections — `compact({ bottommost: true })` and `txn.abandonWrites()` docs — now restored: [README.md](https://github.com/HarperFast/rocksdb-js/pull/742/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)).

Independent review of the rebased head found two real, fixed bugs in the already-merged `setCompression` code:

- `compressionPersistDirty` was unconditionally overwritten on every failed `SetOptions()` call, including one that never touched live options at all — clobbering a pending persist-retry flag an earlier failed call had set. Now only set `true` on an in-memory-only apply; cleared only by a successful persist ([database.cpp:1363](https://github.com/HarperFast/rocksdb-js/pull/742/files#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R1363)).
- `setCompression()` left `Store.compressionForAllColumnFamilies` untouched, so a same-instance `close()`+`open()` would restamp every other column family with the live-changed algorithm — the same "first open dictates every CF's algorithm" class AGENTS.md documents as a previously-caught bug. Now cleared unconditionally, with the live-state resync (including on a persist failure that still applied in memory) best-effort against a racing close ([database.ts:377-405](https://github.com/HarperFast/rocksdb-js/pull/742/files#diff-724f334a2b47a5e7cbcab252ce04fc232071d0bc058c96ed6fd4ac96e5dff308R377)).

Ran 8 rounds of the cross-model pre-push review CLI (codex + gemini, cursor/domain pruned as narrow-low-risk deltas) across the rebase + 5 follow-up commits; each round's actionable findings were fixed and verified before the next. Full receipts recorded against each pushed SHA.

## For the human reviewer

Two findings from the review were investigated and **declined** with evidence rather than fixed, since fixing either properly needs a design decision beyond this maintenance pass:

- **Concurrent-setter race on `compressionPersistDirty`** ([database.cpp:1363](https://github.com/HarperFast/rocksdb-js/pull/742/files#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R1363)): two `setCompression()` calls racing on the same column family have no mutual exclusion beyond RocksDB's own internal locking (`ACQUIRE_OPERATIONS_LOCK` is an in-flight refcount, not a mutex). A full fix means a lock spanning the whole read-modify-write-including-the-RocksDB-call sequence, which needs its own design/testing rather than a rushed patch. Reachability requires two concurrent setters on one CF.
- **`db.use()`-derived column-family views keep the original `compressionForAllColumnFamilies`** ([database.ts:377-405](https://github.com/HarperFast/rocksdb-js/pull/742/files#diff-724f334a2b47a5e7cbcab252ce04fc232071d0bc058c96ed6fd4ac96e5dff308R377)): `Store#createColumnFamilyStore` derives from the frozen constructor-time options snapshot, not the live mutable fields `setCompression()` updates. This is a pre-existing `Store` design characteristic that only becomes observable now that `setCompression()` is the first way to change compression post-open; whether a derived view should inherit the live values or its own frozen ones is a `Store` view-derivation semantics question, not a narrow bug fix.

A handful of other review findings were investigated and refuted with evidence (not applied): a claim that RocksDB's `GetOptions()` might return a resolved compression level instead of the `kDefaultCompressionLevel` sentinel is contradicted by the existing "omitting the level resets it to the algorithm default" test; a claim that `ColumnFamilyDescriptor` (and its `compressionPersistDirty` flag) might not be shared across handles on the same column family is contradicted by `DBDescriptor::columns` being a `shared_ptr` map (AGENTS.md invariant 6); a claim that the new native-boundary test's use of `'gzip'` could be build-dependently flaky is contradicted by `'gzip'` not being a recognized algorithm name in this binding's `compressionTypeFromName` at all (only `'zlib'`).

The CI workaround expanding the Windows/Bun job to three full-suite retries (pre-existing on this branch, unrelated to `setCompression`) was left as-is — it masks HarperFast/rocksdb-js#695 rather than fixing it, which is a known, already-tracked tradeoff.

## Test plan
- [x] `pnpm check` (type-check, lint, format) clean
- [x] `pnpm test` — 955 passed, 9 pre-existing skips (unrelated)
- [x] `pnpm test:native` — 194 passed
- [x] Native binding (`pnpm build:binding`) and TS bundle (`pnpm build:bundle`) both rebuild clean
- [x] Tests: live getter effect, level round-trip/reset, on-disk-size proof of live algorithm change (SST + blob threshold), cold-reopen durability, already-open-CF conflict-check interplay, read-only rejection, validation (JS wrapper + native boundary), same-instance reopen (both single-CF live-value sync and all-CF non-restamp)
- [x] README.md Compression section updated with the new API's semantics and verified caveats; two rebase-conflict-dropped sections restored
- [x] Independent cross-model review: 8 rounds total across the rebase and 5 follow-up commits (codex + gemini graded/finding legs each round); every actionable finding fixed and re-verified before the next round; converged (last two rounds surfaced only already-adjudicated repeats or refuted claims)

https://claude.ai/code/session_01WV3oB7zEeJvKFugn8DQ93e

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=8; full=1 @ 1f48c81e3010</sub>

<sub>Human-Review-Need: 4 @ 1f48c81e3010</sub>


kriszyp:
Parking this for future consideration rather than merging it now — moving to draft.

Opened #743 as the narrower alternative: `compressionForAllColumnFamilies` on the open call, which applies the requested codec to every column family instead of only the target. That covers the case actually blocking Harper today (a database upgraded from 5.1, where every family is persisted at `none` and the per-table reconcile open cannot move them), without live mutation of an already-open family.

The distinction is which model we want:

- **One codec per deployment** — the codec is known from config before any open, so it can just be supplied at open time. #743 is enough, and there is no `SetOptions()` cost or new mutable native state.
- **A codec per table** — you cannot know a table's codec until you have read the catalog, which is itself opened by that same `DB::Open`, so the only way to act on it afterwards is to change an open family's codec. That is this PR, and nothing else substitutes for it.

Harper's per-table dimension today is only enabled-vs-disabled, derived from one global `storage.compression` value at table-creation time, so the deployment-global model fits what exists. If we later want genuine per-table codecs (a `@table(compression:)` directive, say), this PR is the thing that unblocks it.

Nothing here is wrong — the native hardening in the second commit (in-flight operation registration, read-only guard, split live/durable state on an OPTIONS-persist failure) is good work and worth keeping on the branch. Parking rather than closing so it's ready if the per-table case comes up.

Generated by Claude Opus 5.

PR #743, opened 2026-08-01
Add compressionForAllColumnFamilies so a database can adopt one codec

kriszyp:
Alternative to #742 for the case Harper actually has: **one codec for the whole database**, rather than a codec per table.

## The problem

RocksDB opens every column family of a database in a single `DB::Open`, and `db_descriptor.cpp` gives each family the caller did not name its **persisted** algorithm, applying the request only to the target:

```cpp
if (cfName == name && options.compression && options.compressionExplicit) {
    applyCompression(cfo, *options.compression, options.compressionLevel);
}
```

That is deliberate and right when codecs are chosen per table — opening one family must not restamp another's. But it makes a database-wide codec inexpressible. A caller whose first open targets some other family (a catalog it has to read before it knows anything else) finds every sibling already open at its old algorithm, and a family's compression cannot change while it is open.

Reproduced against families persisted at `none`, opening a non-target family first:

```
default:    __catalog__ -> lz4,  records -> throws "already open with compression none"
with flag:  __catalog__ -> lz4,  records -> lz4,  orders -> lz4
```

This is exactly what blocks Harper on a database upgraded from 5.1: every column family was created before the prebuild carried any codecs, so they are all persisted at `none`, and the per-table reconcile open cannot move them.

## The change

`compressionForAllColumnFamilies: true` applies the requested algorithm to every family in the descriptor list instead of only the target.

```typescript
const db = RocksDatabase.open('/path/to/db', {
	name: 'catalog',
	compression: 'lz4',
	compressionForAllColumnFamilies: true,
});
```

- Requires an explicit `compression` — throws otherwise, so the lz4 default can still never restamp a persisted codec by accident.
- Only affects the open that creates the database handle; later opens of the same path reuse it.
- Governs newly written files, like any compression change. Use `compact({ bottommost: true })` (#740) to rewrite what is already on disk.

## Relationship to #742

#742 adds `setCompression()` via `SetOptions()` to change an already-open family's codec at any time. That is strictly more capable — it is what you need if a caller must discover a *per-table* codec from a catalog it can only read after opening everything.

This PR takes the narrower route that fits the deployment-global model: the codec is known from config before any open, so it can simply be supplied at open time. No live mutation of an open family, no new mutable native state, no `SetOptions()` cost. The trade-off is that it cannot express per-table codecs — a family deliberately persisted at `none` is restamped along with the rest, which is the intended behavior here but would be wrong under a per-table model.

Worth deciding which model we want before merging either; they are not mutually exclusive, but only one is needed for Harper today.

## Testing

Three cases in `test/compression.test.ts`: the default leaves un-named families at their persisted algorithm (asserting the sibling open throws), the flag makes both siblings reconcile to the requested codec, and omitting an explicit algorithm throws. Skipped when the native build lacks lz4. 54 tests pass across the compression, compaction, column-families and db-options suites.

Verified against a RocksDB prebuild carrying the codecs (11.1.2); the repo's vendored prebuild has only none/zlib, so the new tests self-skip there.

Generated by Claude Opus 5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -977,6 +977,27 @@ was compiled against, so it varies by build. Always check
 [`supportedCompression`](#supportedcompression) at runtime — opening with an unavailable algorithm
 throws. `'none'` is always available.
 
+**Adopting a codec across a whole database.** Compression is chosen per column family, and RocksDB
+opens every family of a database in one call — so by default the families you did not name keep
+their persisted algorithm. If your first open targets some other family (a catalog, say), the rest
+are already open at their old algorithm before you can ask for a new one, and a family's compression
+cannot be changed while it is open. Pass `compressionForAllColumnFamilies: true` to apply the
+requested algorithm to all of them instead:
+
+```typescript
+// Every column family in this database adopts lz4, not just 'catalog'
+const db = RocksDatabase.open('/path/to/db', {
+	name: 'catalog',
+	compression: 'lz4',
+	compressionForAllColumnFamilies: true,
+});
+```
+
+It requires an explicit `compression`, and only takes effect on the open that creates the database
+handle — later opens of the same path reuse that handle. As always the algorithm governs newly
+written files; use [`compact({ bottommost: true })`](#dbcompactoptions-promisevoid) to rewrite what
+is already there.
+
 **Scope and mutability.** Compression is genuinely **per-column-family**. Each `RocksDatabase`
 targets one column family (`name`), and its `compression` applies only to that CF — opening one CF
 never changes another's algorithm, regardless of open order. A column family keeps its own
--- a/src/binding/database/database.cpp
+++ b/src/binding/database/database.cpp
@@ -1474,6 +1474,21 @@ napi_value Database::Open(napi_env env, napi_callback_info info) {
 		dbHandleOptions.compression = rocksdb::kLZ4Compression;
 	}
 
+	NAPI_STATUS_THROWS(rocksdb_js::getProperty(
+		env,
+		options,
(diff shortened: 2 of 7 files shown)


cb1kenobi:
Reviewed `3b8a5356` — no issues found. This PR looks good, nice job!

Checked the native path carefully: the flag is applied to every column family in the descriptor list at open (`db_descriptor.cpp`), the default stays backward-compatible (`false` → the condition reduces exactly to the prior `isTarget && compression && compressionExplicit`, so existing DBs are never restamped), validation correctly requires an explicit `compression` (guards on `compressionExplicit`, set before the check), no ABI/`ColumnFamilyOptions` mismatch (binding-level flag only), and no new CF-handle lifetime/memory concerns. Transaction logs are `.txnlog` files rather than a reserved CF, so nothing internal is unexpectedly restamped. Docs and three tests cover the default, all-CF, and missing-algorithm cases.

—
_Generated by Barber AI_

cb1kenobi:
Re-reviewed `9ad628e6` — no issues found. This PR looks good, nice job!

The incremental change since `3b8a5356` is test-only (`test/compression.test.ts`): `seed()` now uses the `tempPath()` helper so seeded DB directories get cleaned up in `afterEach` instead of leaking, and the two lz4-gated cases were converted from manual `if (...) return` early-returns to `it.skipIf(...)` (consistent with the rest of the suite, and reports SKIPPED rather than a false PASS). No native, binding, or ABI changes — the every-CF application, backward-compatible default, explicit-compression validation, CF-handle memory safety, and per-store path are all untouched.

—
_Generated by Barber AI_

Issue #672, opened 2026-06-23
Async DB ops (backup, createCheckpoint) defer registry purge when closed before settling


cb1kenobi:
## Summary

`Database::Backup` and `Database::CreateCheckpoint` run on the worker pool and pin the `DBDescriptor` (`shared_ptr`) in their async state to keep the underlying `rocksdb::DB` alive for the duration of the copy. Because `DBRegistry::CloseDB` only purges a descriptor when its `use_count() <= 1`, an async op that is still in flight holds an extra reference and **defers the registry purge**.

## Behavior today

If a caller starts a backup/checkpoint, does **not** await it, and calls `db.close()` before it settles:

- `close()` cancels and waits up to `waitForAsyncWorkCompletion()`'s 5s timeout, then resets the handle's descriptor ref.
- At the purge check the async state's descriptor copy keeps `use_count` above 1, so the purge is skipped and **never retried** — the descriptor lingers in the registry (holding the DB open + its file lock) until process exit / `shutdown()` / `PurgeAll`.
- A reopen of the same path during this window reuses the live descriptor, so reads/writes still work; the cost is the deferred cleanup (memory, file handles, lock) for a closed-and-not-reopened database.

This is a pre-existing characteristic of `backup` and now also `createCheckpoint` (#577), which intentionally matches the backup pattern rather than diverging.

## Why the naive fix is not enough

Pinning the `rocksdb::DB` shared_ptr instead of the descriptor lets the purge proceed, but introduces a **worse** race: if the copy outlives the 5s close timeout, `CloseDB` erases the registry entry while the worker still holds the DB + lock open, so a subsequent `open(samePath)` finds no entry and **cannot acquire RocksDB's lock** until the copy finishes. (Surfaced by cross-model review on #577.)

## Proposed fix

Keep the descriptor pinned (so the registry entry persists during the copy and reopen reuses the live DB — no lock race), and add a path-based purge retry fired from the async **complete** callback once the descriptor ref is released:

- After the op completes and the state's descriptor ref is dropped, re-check the registry entry under `databasesMutex`; if `use_count() <= 1 && beginClose()`, run `finishClose()` + erase + notify.
- No-op when a live handle remains (normal awaited-then-used path).

This must be coordinated cross-env per the teardown invariants in `AGENTS.md` (note #6) and should be exercised under the Guard Malloc teardown harness, since it touches the registry purge path.

## Scope

- `src/binding/database/backup.cpp` (`AsyncBackupState`)
[...]

PR #744, opened 2026-08-01
fix(transaction): bound the coordinated-retry park with a descriptor-owned timeout

### LINKED ISSUE #741: Worker-env teardown destroys transactions on the shared DBDescriptor, corrupting the heap under concurrent commits
## Summary

When a `worker_threads` env shuts down while **other live envs are still committing** through the same database path, `DBRegistry::Shutdown` → `DBDescriptor::finishClose` → `TransactionHandle::close` destroys RocksDB transactions belonging to that shared descriptor and corrupts the native heap. Reproduces as `SIGABRT` (glibc `double free or corruption (!prev)`) or `SIGSEGV`.

A `DBDescriptor` is process-global and shared by every env that opens the path (AGENTS.md note 6), so one env's teardown is destroying transaction state other envs — and the dedicated commit thread — are still using.

## Reproduction

Repro script: 4 `worker_threads` open the same DB path and drive `coordinatedRetry` transactions at a **single hot row**; one worker is recycled every 2s.

| config | result |
|---|---|
| recycling off | clean (445k+ `RETRY_NOW` — the park path is heavily exercised) |
| recycling on, `worker.terminate()` | 4/5 runs crashed (3× SIGABRT, 1× SIGSEGV) |
| recycling on, **graceful** exit (`db.close(); process.exit(0)`) | 3/3 runs crashed (2× SIGABRT, 1× SIGSEGV) |

Graceful exit crashing rules out abrupt `terminate()` as the trigger — an ordinary clean worker exit is sufficient.

Tested against the **deployed artifact**: `rocksdb-js.node` (2.5.0) taken from a running harper-pro 5.2.0-beta.4 container, executed locally on Node 24.18.0 (ABI 137, matching the container's 24.18.1). RocksDB 11.1.2.

## Stack

```
node::worker::Worker::Run
  node::FreeEnvironment -> node::Environment::RunCleanup -> node::CleanupQueue::Drain
    rocksdb_js::DBRegistry::Shutdown
      rocksdb_js::DBDescriptor::finishClose
        rocksdb_js::TransactionHandle::close
          rocksdb::OptimisticTransaction::~OptimisticTransaction
            rocksdb::TransactionBaseImpl::~TransactionBaseImpl
              _int_free   <-- "double free or corruption (!prev)"
```

## Why this may matter beyond the crash

Corruption is the loud outcome of this race. The quiet outcome would be a **commit completion never delivered to its originating env** — i.e. `Transaction.commit()`'s promise never settling. AGENTS.md already flags that the per-env commit-completion tsfn is released on env teardown and that "a per-commit tsfn acquire is NOT sufficient (env teardown does not honor tsfn acquire counts)" — same subsystem, same race.

That silent variant is being investigated as the mechanism behind HarperFast/harper#2001 (a worker thread whose commit promise never settles, rejecting every subsequent write with 503 until restart — no error logged, no CPU burn, RocksDB reporting 0% stall). **That link is not proven** — this issue reports only the reproducible crash. Filing separately so the memory-safety bug is tracked on its own merits.

## Notes

Harper recycles http worker threads in normal operation while multiple workers write the same hot rows, which is precisely this repro's shape.

Reported by Claude Opus 5 on behalf of @kriszyp.

kriszyp:
## Summary

Fixes #741 by bounding every coordinated-retry park. A commit blocked behind a holder that never releases now resolves `RETRY_NOW` after `ROCKSDB_JS_PARK_TIMEOUT_MS` (default 5000 ms) instead of leaving the write path hung indefinitely.

- Adds a descriptor-owned timeout registry with monotonic deadline ordering and a lazily started thread, joined during descriptor close.
- Keeps the VT wake callback off `DBDescriptor` and `DBRegistry`: it captures only weak timeout-registry and exactly-once state, avoiding registry re-entry and descriptor reference-count perturbation while the process-global VT writer mutex is held ([wake path](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR595)).
- Cancels timeout TSFNs during worker-environment cleanup while snapshotting only independently owned registries, so cleanup cannot make a concurrent descriptor purge skip ([env cleanup](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R671)).
- Handles timeout-thread and N-API/TSFN setup failures with a bounded immediate retry or terminal promise rejection; no setup failure can silently leave the commit promise pending ([failure path](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR303)).
- Reads `ROCKSDB_JS_PARK_TIMEOUT_MS` once per process behind a function-local static, matching `commitThreadMode()`/`commitDelayMs()` in the same file: a park runs on whichever env's JS thread owns the transaction, and `::getenv` is not safe against a `process.env` write on another thread ([`parkTimeoutMs()`](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR251)).
- Guards both TSFN resolution paths (`ParkTimeoutRegistry::resolve` and the no-timeout-thread degrade) on `napi_call_threadsafe_function`'s status before releasing: a `napi_closing` result means the tsfn may already be mid-teardown on Node's side, so releasing it again is a use-after-close ([`ParkTimeoutRegistry::resolve`](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR639)) — found by the pre-push review during this rebase, not present in earlier rounds.
- Measures every elapsed-time assertion on the monotonic clock, including the conflict-wake bound in the concurrent-transactions test, which was still reading `Date.now()` on both ends ([`lock-tracker.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-eb8cd8f6a4cf0503ceeeb6418b69a712c67b883a30e7f427953c1e2830f4e3a7R126)) — an NTP step between the two reads was enough to push it past the 3000 ms bound with nothing wrong in the wake path.
- Documents the bound where the behavior is specified. The README described the coordinated-retry wait as lasting "until the conflicting transaction has committed and released its write intent", which this change makes untrue; it now states the timeout, that a timeout consumes a retry attempt, the resulting `ERR_TRANSACTION_ABANDONED` at roughly `maxRetries x` the timeout, and that `0`/negative/malformed values select the 5000 ms default rather than disabling the bound ([README retry section](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R782)).
- Drops the exactly-once gate on the no-timeout-thread degrade. Every path that returns 0 from `ParkTimeoutRegistry::schedule` does so before the entry publishes `fired`, and the `fireOnce` wake closure is not built until after this branch breaks out, so that side owns the park outright and the compare-exchange could never fail ([degrade path](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR568)).
- Raises the fork fixture's parent-side kill timer from 4 s to 20 s. The child's own `elapsed` assertion is the deadline check; the parent's timer additionally covers node boot, type-stripping, addon load and the seed writes, so at the same 4 s budget a slow runner and a real regression both surfaced as `SIGTERM` with the fixture's diagnostic never thrown. 20 s stays under vitest's 30 s `testTimeout` so the kill path still dumps the child's stderr ([backstop](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-eb8cd8f6a4cf0503ceeeb6418b69a712c67b883a30e7f427953c1e2830f4e3a7R197)).
- Adds deterministic timeout and close-lifecycle regressions, including an isolated process that proves the test fails against `origin/main` ([tests](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-eb8cd8f6a4cf0503ceeeb6418b69a712c67b883a30e7f427953c1e2830f4e3a7R187)).

## For the human reviewer

The implementation deliberately preserves the task's policy choices:

- A timeout consumes one of the finite coordinated-retry attempts. A legitimately slow holder can therefore fail after roughly `maxRetries × timeout`; operators can raise the timeout but cannot disable the bound with `0`. This is now stated in the README rather than only in the code, so it is a documented contract from here on.
- The knob is process-wide (env var, read once at the first park) and the implementation uses one lazy timeout thread per descriptor. Moving this to a per-database open option or a process-global timer would be a separate API/architecture change.
- `LockTracker` has no callback-removal API. Timed-out callbacks become inert immediately, but registrations remain until the holder wakes; adding cancellation is deferred because it changes the VT concurrency surface. Note the shape of the residual: registrations now accumulate per *retry* rather than per incident, so a long-lived abandoned holder under heavy contention grows one tracker's callback vector for as long as the incident lasts.

Declined, with reasons — each is a real observation, none is required to make the fix correct:

- **Park timeouts are operationally silent.** A synthetic wake is indistinguishable from a real one, so an exhausted retry reports the same generic `Transaction did not commit after N coordinated retries` a hot key produces. A `transaction:parkTimeout` global event (the `transactionLog:warning` shape) would make the abandoned-holder incident diagnosable — new API surface, so it belongs in its own change. Partially mitigated this round on the configuration half: a misconfigured `ROCKSDB_JS_PARK_TIMEOUT_MS` is still silent at runtime, but the README now says which values are ignored, so an operator who sets `0` can find out why it had no effect.
- **`db.transaction()` orchestration is not covered end-to-end.** [`test/fixtures/fork-park-timeout.mts`](https://github.com/HarperFast/rocksdb-js/pull/744/changes?w=1#diff-e0ee3c42fe7484eef76516f98a7a7c05d511073d428089adbfe40c275c52ad7aR42) drives `Transaction.commit()` directly and asserts the native `RETRY_NOW`. I traced `src/database.ts:1014-1027` and the loop does terminate correctly, but nothing pins it.
- **A parked commit does not hold the event loop** (`napi_unref_threadsafe_function`, pre-existing — the fixture's `setInterval` keep-alive exists for this). Ref-ing for the park's duration is only safe now that the wait is bounded, but it is a behavior change beyond this fix.
- **Comment density in `db_descriptor.cpp`.** All three lenses flag the inline narration around `ParkTimeoutRegistry` (`// Fire while still holding the mutex`, `// Notify + join outside the lock`) as saying what the next line says. I trimmed only the two blocks this round touched; rewriting comments from earlier rounds under a nit that has been adjudicated as a nit five times running seemed like the wrong kind of churn for a review-feedback pass. The two blocks that do carry invariants — AGENTS.md note 12 and the `ParkTimeoutRegistry` class header — are the reason the #672 purge-skip does not come back, and the domain lens explicitly asked that they stay.
- **`fire()` takes the park registry's mutex while the global VT `writerMutex_` is held**, so a wake can wait behind `runLoop()`/`shutdown()` draining a batch of due parks. Everything done under that mutex is non-blocking — `napi_call_threadsafe_function` is `napi_tsfn_nonblocking`, and `shutdown()`'s `join()` is deliberately outside the lock — and N is bounded by one descriptor's concurrent coordinated-retry commits. The suggested `try_lock` would silently drop a real early wake and fall back to the full timeout, which is worse than the latency it avoids.
- **The first `::getenv` in `parkTimeoutMs()` can still race a `process.env` write.** True, and exactly the exposure `commitThreadMode()`/`commitDelayMs()` already carry three lines below — the previous round asked for this shape specifically. Closing it properly means capturing all three at module init, which is a separate change to code this PR does not otherwise touch.
- **`std::bad_alloc` between the ref transfer and the park registration would terminate the process.** Pre-existing: before this change the same window ended in `addWakeCallback()` pushing onto a `std::vector<std::function>`, which can throw identically, and no N-API callback in this codebase is exception-guarded. Worth noting `ParkTimeoutRegistry::schedule()` does not leak on a partial insert — an orphaned deadline whose park is missing is erased by `runLoop()`'s existing `parks.find() == end()` branch.
- **`retry: 1` on `commit-teardown.test.ts` can let a probabilistic native abort pass.** It is the narrowing the previous round asked for (Deno-on-macOS only, one test, rather than re-rolling the whole suite); a `skipIf` would remove the crash guard on that platform outright rather than weaken it.
- **`retryNowFinalize` deletes its refs without an `env == nullptr` guard** — checked and declined as a false positive. The NULL-`env` convention is documented for `napi_threadsafe_function_call_js` only; the finalize callback runs on the JS thread with a live env when the thread count hits zero, which is the contract `event_emitter.cpp:216-219` already states and relies on. The function is also unchanged by this PR.

- **Join-at-close is still not covered by a test.** I re-ran the mutation myself: `join()` → `detach()` in `ParkTimeoutRegistry::shutdown()` leaves the file 12/12 green, because `shutdown()` resolves pending parks under its mutex *before* joining. The close test's comment no longer claims otherwise. Real coverage needs the ASan lane, not another deadline assertion.

Cross-model review used independent Codex and Gemini lenses plus Harper-domain adjudication across nine rounds. It found and drove fixes for the original wake-path deadlock, transient descriptor pins, deadline lookup, thread-creation failure, promise-settlement failures, `getenv` thread-safety, comment accuracy, the wall-clock assertion above, and — in round 8, rebasing onto latest `main` — the `napi_closing` use-after-close noted above. Round 9 (delta, after that fix) confirmed no new defect and that the other findings carry forward unchanged. Rounds 10-12 covered this review-feedback round: round 10 verified the CAS removal is sound (confirming `schedule` returns 0 only before it moves `fired` into an entry, and that `fireOnce` is constructed afterwards) and raised one new finding caused by the README addition itself - documenting the knob as operator-facing while every misconfiguration of it stays silent - which round 11 closed by stating the parsing rules; round 12 covered the comment trim. No round found a defect in the delta. The Cursor lens has never run on this branch: `cursor-review` refuses any diff that touches agent instructions, and this one edits `AGENTS.md`.

**Rebase note (rounds 8–9):** rebased onto latest `origin/main` (`da0ef135`) with no semantic conflicts — the only overlap was an independent upstream `AGENTS.md` invariant renumbering (main inserted its own invariant 13; this PR's invariant 13 became 14), resolved by keeping both. `pnpm build`, `pnpm test:native` (151 passed), and the park/lock-tracker/commit-teardown/orphan-gc Vitest suites (21 passed) were re-run clean against the rebased tip before and after the `napi_closing` fix above.

## Verification

- `pnpm build`
- `pnpm check` — clean (type-check, oxlint, oxfmt)
- `pnpm test:native` — 151 passed
- `pnpm test` — 58 files passed, 805 tests passed, 3 skipped
- `node --expose-gc ./node_modules/vitest/vitest.mjs test/lock-tracker.test.ts test/commit-teardown.test.ts` — 14 passed after the final delta
- Re-verified from a clean worktree at `65153100` (Linux, Node 26.2.0, RocksDB 11.8.1): `pnpm check` clean, `pnpm test:native` 151 passed, `pnpm test` 58 files / 805 passed / 3 skipped, and the two park files 14 passed.
- Re-verified again after the rebase onto `main` and the `napi_closing` fix, at `dc1bacfd` (Linux, Node 26.2.0, RocksDB 11.8.1): full `pnpm build`, `pnpm test:native` 151 passed, and `lock-tracker.test.ts` + `commit-teardown.test.ts` + `transaction-orphan-gc.test.ts` 21 passed.
- Re-verified after this review-feedback round at `d2587a27` (Linux, Node 26.2.0, RocksDB 11.8.1): `pnpm build:binding` + `pnpm build:bundle`, `pnpm check` clean, `pnpm test` 59 files / 819 passed / 3 skipped (the one initial failure was `transaction-log-crash-recovery.test.ts` needing a `dist` bundle that did not exist in the fresh worktree; it passes once built), and `lock-tracker.test.ts` 12/12 in 2.93 s with the raised backstop.
- Fails-on-base proof: the isolated timeout regression was applied to `origin/main`, where the child exceeded its 4-second guard and was terminated; the branch completes in under 500 ms.
- Env-read regression check: with the value read once per process, the fork fixture still completes in 369 ms and still asserts `elapsed >= 40`, so both halves of the parse rule (the child's `ROCKSDB_JS_PARK_TIMEOUT_MS=1` taking effect, and the 50 ms clamp) stay covered.

Refs #741

<sub>Review-Coverage: authored=claude; ran=codex,gemini; adjudicated=domain; blocked=cursor-grok(no-receipt); declined=cursor-composer; rounds=12 @ d2587a273272</sub>

<sub>Human-Review-Need: 4 (decisions: timeout-consumes-a-retry-attempt, default-5000ms-and-no-opt-out, bound-at-the-park-vs-reap-the-holder, env-var-only-configuration, plain-std-thread-per-descriptor) @ d2587a273272</sub>

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -214,12 +214,24 @@ jobs:
         run: pnpm build
 
       - name: Run tests on Deno
-        # macos-latest gets an extra retry for a known pre-existing native crash in
-        # commit-teardown.test.ts (HarperFast/rocksdb-js#746, mutex lock failed:
-        # Invalid argument during worker-env teardown vs. the shared commit thread)
-        # that has hit multiple unrelated PRs on this OS/runtime combo. Not a fix
-        # for #746 itself, just enough retry budget that a rare pre-existing race
-        # doesn't block unrelated PRs -- same pattern as the windows-latest retry
-        # below (HarperFast/rocksdb-js#695).
+        # macos-latest gets an extra retry ONLY for the known pre-existing native
+        # crash in commit-teardown.test.ts (HarperFast/rocksdb-js#746, "mutex lock
+        # failed: Invalid argument" during worker-env teardown vs. the shared commit
+        # thread), which has hit multiple unrelated PRs on this OS/runtime combo.
+        # Gating on that crash string, rather than re-rolling any failure like the
+        # windows-latest retry below (HarperFast/rocksdb-js#695), keeps this OS's
+        # signal for every other regression. Not a fix for #746 itself.
         shell: bash
-        run: pnpm test:deno || ([ "${{ matrix.os }}" = "windows-latest" -o "${{ matrix.os }}" = "macos-latest" ] && pnpm test:deno)
+        run: |
+          # pipefail is what makes the `| tee` below report the test exit code
+          # instead of tee's.
+          set -o pipefail
+          if pnpm test:deno 2>&1 | tee "$RUNNER_TEMP/deno-test.log"; then
+            exit 0
+          fi
+          case "${{ matrix.os }}" in
+            windows-latest) ;;
+            macos-latest) grep -q 'mutex lock failed' "$RUNNER_TEMP/deno-test.log" || exit 1 ;;
+            *) exit 1 ;;
+          esac
+          pnpm test:deno
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -214,24 +214,5 @@ jobs:
         run: pnpm build
 
       - name: Run tests on Deno
-        # macos-latest gets an extra retry ONLY for the known pre-existing native
-        # crash in commit-teardown.test.ts (HarperFast/rocksdb-js#746, "mutex lock
-        # failed: Invalid argument" during worker-env teardown vs. the shared commit
-        # thread), which has hit multiple unrelated PRs on this OS/runtime combo.
-        # Gating on that crash string, rather than re-rolling any failure like the
-        # windows-latest retry below (HarperFast/rocksdb-js#695), keeps this OS's
-        # signal for every other regression. Not a fix for #746 itself.
         shell: bash
-        run: |
-          # pipefail is what makes the `| tee` below report the test exit code
-          # instead of tee's.
-          set -o pipefail
-          if pnpm test:deno 2>&1 | tee "$RUNNER_TEMP/deno-test.log"; then
-            exit 0
-          fi
-          case "${{ matrix.os }}" in
-            windows-latest) ;;
-            macos-latest) grep -q 'mutex lock failed' "$RUNNER_TEMP/deno-test.log" || exit 1 ;;
-            *) exit 1 ;;
-          esac
-          pnpm test:deno
+        run: pnpm test:deno || ([ "${{ matrix.os }}" = "windows-latest" ] && pnpm test:deno)
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -214,5 +214,12 @@ jobs:
         run: pnpm build
 
       - name: Run tests on Deno
+        # macos-latest gets an extra retry for a known pre-existing native crash in
+        # commit-teardown.test.ts (HarperFast/rocksdb-js#746, mutex lock failed:
+        # Invalid argument during worker-env teardown vs. the shared commit thread)
+        # that has hit multiple unrelated PRs on this OS/runtime combo. Not a fix
+        # for #746 itself, just enough retry budget that a rare pre-existing race
+        # doesn't block unrelated PRs -- same pattern as the windows-latest retry
+        # below (HarperFast/rocksdb-js#695).
         shell: bash
-        run: pnpm test:deno || ([ "${{ matrix.os }}" = "windows-latest" ] && pnpm test:deno)
+        run: pnpm test:deno || ([ "${{ matrix.os }}" = "windows-latest" -o "${{ matrix.os }}" = "macos-latest" ] && pnpm test:deno)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -193,7 +193,7 @@ sufficient (env teardown does not honor tsfn acquire counts); see
   each completion callback (widens teardown race windows)
 - `ROCKSDB_JS_TXN_GET_DELAY_MS` - Test-only: delay a transaction's cold-cache async get before
   it reads (exercises orphan cleanup past the async-work wait timeout)
-- `ROCKSDB_JS_PARK_TIMEOUT_MS` - Bounded wait (default `3000`) before a
+- `ROCKSDB_JS_PARK_TIMEOUT_MS` - Bounded wait (default `5000`) before a
   coordinated-retry commit parked on a conflicting holder's VT lock resolves
   RETRY_NOW unconditionally, in case the holder never releases (see
   "Coordinated retry" note below)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -196,7 +196,10 @@ sufficient (env teardown does not honor tsfn acquire counts); see
 - `ROCKSDB_JS_PARK_TIMEOUT_MS` - Bounded wait (default `5000`) before a
   coordinated-retry commit parked on a conflicting holder's VT lock resolves
   RETRY_NOW unconditionally, in case the holder never releases (see
-  "Coordinated retry" note below)
+  "Coordinated retry" note below). Read per park; values below `50` are clamped
+  up to it, and `0` (an ambiguous "disable the bound") falls back to the default
+  like any malformed value. There is no opt-out: a deployment that would rather
+  wait than fail a legitimately slow holder raises the value instead
 
 ## Test Structure
 
@@ -423,37 +426,58 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     that lock's last holder releases (`VerificationTable::releaseWriteIntent` → `LockTracker::wake`).
     A holder that never releases — a leaked/abandoned transaction, or a wake lost to a bug elsewhere —
     would otherwise park forever (harper#2001: a worker's write path disabled for 5+ hours until
-    restart). `DBDescriptor::scheduleParkTimeout` (`db_descriptor.{h,cpp}`) bounds this with
+    restart). `ParkTimeoutRegistry` (`db_descriptor.{h,cpp}`) bounds this with
     `ROCKSDB_JS_PARK_TIMEOUT_MS` (default `5000` — the top of this fix's requested 2-5s range, to
     leave maximum headroom for a holder that is merely slow rather than abandoned, since a timeout
     consumes a `coordinatedRetry` attempt exactly like a real wake does and `maxRetries` is finite):
-    one park-timeout thread **per descriptor** — lazily started, joined at `finishClose()` (and again,
-    idempotently, from the destructor as a safety net, matching `commitWorker`) — tracks every
-    outstanding deadline instead of spawning a thread per park (the contention path is exactly where
-    an abandoned holder makes parks dense, so per-park threads would be a resource cliff, not a fix).
+    one registry, and one lazily-started timeout thread, **per descriptor** — joined at
+    `finishClose()` (and again, idempotently, from the destructor as a safety net, matching
+    `commitWorker`) — tracks every outstanding deadline instead of spawning a thread per park (the
+    contention path is exactly where an abandoned holder makes parks dense, so per-park threads would
+    be a resource cliff, not a fix). Deliberately a plain `std::thread`, not a `uv_timer_t`: this addon
+    ships one prebuilt binary across Node ABI versions via N-API, and libuv's struct layout is not part
+    of that stable surface.
+
+    **A `LockTracker` wake callback runs under the process-global VT `writerMutex_`, so it must not
+    block and must not re-enter the VT or `DBRegistry`.** `LockTracker::wake()` invokes its callbacks
+    inline and both callers (`releaseWriteIntent`, `cancelForDB`) hold that mutex across the whole
+    function. Re-entering the registry from there self-deadlocks: `DBRegistry::PurgeIfUnreferenced`
+    can claim the purge and call `finishClose()` → `cancelForDB()` → a second lock of the same
+    non-recursive `writerMutex_`, wedging every database's write-intent path process-wide — the exact
+    symptom this note exists to fix. It is also an AB-BA against `finishClose`'s `txnsMutex` →
+    `writerMutex_` order, and it would run a flush, a manual compaction, `WaitForCompact` and thread
+    joins under the global VT lock. That is why `ParkTimeoutRegistry` is a standalone object owned by
+    the descriptor through a `shared_ptr` rather than state on the descriptor itself: the wake closure
+    captures a **`std::weak_ptr<ParkTimeoutRegistry>`** and calls only `fire(id)`, which touches one
+    mutex and one map. Weak, not raw, because a park can end up registered on a tracker installed by a
+    _different_ database on a colliding VT slot (`VerificationTable::lockSlotForWrite` joins an existing
+    tracker without retagging its `dbId`), so that lock's eventual release wakes a park whose own
+    database may already have closed — `cancelForDB()` only wakes trackers tagged with _its own_
+    `vtEpoch`, so it cannot be relied on to have resolved a foreign-`dbId` park first. Weak **to the
+    registry and not to the descriptor** because a `weak_ptr<DBDescriptor>::lock()` is a transient extra
+    reference, and `PurgeIfUnreferenced` decides on `use_count() <= 1`: a racing close would see the
+    inflated count, skip the purge, and leak the registry entry plus the open RocksDB — the
+    HarperFast/rocksdb-js#672 hazard, which the wake path cannot repair by retrying the purge (that is
+    the re-entrancy above). `.lock()` failing is the expected outcome once the owning database closes:
+    `ParkTimeoutRegistry::shutdown()` (called from `finishClose()` right after `cancelForDB`, before
+    the descriptor can be destroyed) unconditionally resolves every park it still holds regardless of
+    whether the real holder ever wakes it, so by the time the weak reference can fail, the park has
+    already settled.
+
     Each park is identified by a monotonic `uint64_t id`, not its entry's address: `LockTracker::wakeCallbacks`
     has no removal API (see the gap noted below), so a stale closure can outlive its entry, and an
     address-keyed lookup risks resolving a _different_, later park that reused the same freed heap
     address. The timeout thread and the LockTracker wake callback race through one heap-allocated
     `std::atomic<bool>` per park (independent of the per-park `RetryNowContext`, whose refs/TSFN the
-    winning side's release eventually frees) — whichever fires first calls+releases the TSFN under
-    `parkTimeoutMutex` and erases the entry; the loser finds it already gone and touches nothing. That
-    same mutex is what a dying env's `releaseParkTimeoutsByEnv` (wired into the module env-cleanup
+    winning side's release eventually frees) — whichever fires first calls+releases the TSFN under the
+    registry's `mutex` and erases the entry; the loser finds it already gone and touches nothing. That
+    same mutex is what a dying env's `releaseByEnv` (wired into the module env-cleanup
     hook next to `ReleaseCommitCompletionsByEnv`) takes to cancel — release without calling — that
     env's pending parks before Node frees their tsfns; `retryNowCallJs` also guards `env == nullptr`
-    like `commitCompletionCallJs` does, for the same tsfn-queue-drained-during-teardown reason. The
-    LockTracker wake closure captures a **`std::weak_ptr<DBDescriptor>`**, not a raw pointer: a park
-    can end up registered on a tracker installed by a _different_ database on a colliding VT slot
-    (`VerificationTable::lockSlotForWrite` joins an existing tracker without retagging its `dbId`), so
-    that lock's eventual release wakes a park whose own descriptor may have already closed and been
-    destroyed — `cancelForDB()` only wakes trackers tagged with _its own_ `vtEpoch`, so it cannot be
-    relied on to have resolved a foreign-`dbId` park before the descriptor goes away. `.lock()` failing
-    is the expected outcome once that happens: `shutdownParkTimeouts()` (called from `finishClose()`
-    right after `cancelForDB`, before the descriptor can be destroyed) unconditionally resolves every
-    park still in its own `parkTimeouts` regardless of whether the real holder ever wakes it, so by the
-    time the weak reference can fail to lock, the park has already settled. Deliberately a plain
-    `std::thread`, not a `uv_timer_t`: this addon ships one prebuilt binary across Node ABI versions via
-    N-API, and libuv's struct layout is not part of that stable surface. Known gap: `LockTracker::wakeCallbacks`
+    like `commitCompletionCallJs` does, for the same tsfn-queue-drained-during-teardown reason. Parks
+    are indexed twice, by id and by deadline (`std::multimap`): `fire()` needs an O(1) lookup because it
+    runs under the global VT mutex, and the timeout thread needs the earliest deadline on every wakeup
+    without an O(N) scan on that same lock. Known gap: `LockTracker::wakeCallbacks`
     itself has no removal API. Before this change an abandoned holder accrued one inert callback per
     waiter and then everything hung; now each waiter re-parks (and re-registers) every
     `ROCKSDB_JS_PARK_TIMEOUT_MS` up to `maxRetries`, so registrations accumulate per _retry_ rather than
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -415,6 +415,7 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     if a zero were taken as a terminator, would let a chain "end" anywhere in megabytes of padding.
     Resolve it only on a break — `getLogFileSize` crosses into native and takes the store mutex, so
     a per-frame call would tax every healthy read.
+
 12. **Coordinated retry parks on a lock, bounded by a descriptor-owned timeout**: a `coordinatedRetry`
     commit that loses a conflict (`IsBusy`) parks instead of rejecting immediately —
     `completeCommitWork` (`src/binding/transaction/transaction.cpp`) registers a wake callback on the
@@ -460,7 +461,7 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     memory-growth concern, not a correctness one) — deferred rather than risking an unreviewed change to
     `verification_table.cpp`'s concurrency invariants under this fix's scope.
 
-12. **A dropped transaction must release itself**: `DBDescriptor::transactionAdd` holds a **strong**
+13. **A dropped transaction must release itself**: `DBDescriptor::transactionAdd` holds a **strong**
     `shared_ptr` (the parallel `closables` entry is weak), so the registry alone keeps a
     `TransactionHandle` alive and `~TransactionHandle` — hence `close()`, the only `ClearSnapshot()`
     path — is unreachable while it is registered. The `NativeTransaction` finalizer therefore calls
@@ -482,7 +483,7 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     handle fields that are fixed before publication (`id`, `createdAt`), because `txnsMutex` covers
     map membership while mutable-field writers hold no lock.
 
-13. **A recovered active transaction-log file ends on a transaction boundary when recovery can
+14. **A recovered active transaction-log file ends on a transaction boundary when recovery can
     prove one**: only a batch's final entry
     carries `TRANSACTION_LOG_ENTRY_LAST_FLAG`, so a crash mid-batch leaves whole, well-framed
     entries that are a _prefix_ of a transaction. `recoverTail()` discards them
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -193,6 +193,10 @@ sufficient (env teardown does not honor tsfn acquire counts); see
   each completion callback (widens teardown race windows)
 - `ROCKSDB_JS_TXN_GET_DELAY_MS` - Test-only: delay a transaction's cold-cache async get before
   it reads (exercises orphan cleanup past the async-work wait timeout)
+- `ROCKSDB_JS_PARK_TIMEOUT_MS` - Bounded wait (default `3000`) before a
+  coordinated-retry commit parked on a conflicting holder's VT lock resolves
+  RETRY_NOW unconditionally, in case the holder never releases (see
+  "Coordinated retry" note below)
 
 ## Test Structure
 
@@ -411,6 +415,50 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     if a zero were taken as a terminator, would let a chain "end" anywhere in megabytes of padding.
     Resolve it only on a break — `getLogFileSize` crosses into native and takes the store mutex, so
     a per-frame call would tax every healthy read.
+12. **Coordinated retry parks on a lock, bounded by a descriptor-owned timeout**: a `coordinatedRetry`
+    commit that loses a conflict (`IsBusy`) parks instead of rejecting immediately —
+    `completeCommitWork` (`src/binding/transaction/transaction.cpp`) registers a wake callback on the
+    conflicting VT slot's `LockTracker` (via `addWakeCallback`) and resolves `RETRY_NOW` only when
+    that lock's last holder releases (`VerificationTable::releaseWriteIntent` → `LockTracker::wake`).
+    A holder that never releases — a leaked/abandoned transaction, or a wake lost to a bug elsewhere —
+    would otherwise park forever (harper#2001: a worker's write path disabled for 5+ hours until
+    restart). `DBDescriptor::scheduleParkTimeout` (`db_descriptor.{h,cpp}`) bounds this with
+    `ROCKSDB_JS_PARK_TIMEOUT_MS` (default `5000` — the top of this fix's requested 2-5s range, to
+    leave maximum headroom for a holder that is merely slow rather than abandoned, since a timeout
+    consumes a `coordinatedRetry` attempt exactly like a real wake does and `maxRetries` is finite):
+    one park-timeout thread **per descriptor** — lazily started, joined at `finishClose()` (and again,
+    idempotently, from the destructor as a safety net, matching `commitWorker`) — tracks every
+    outstanding deadline instead of spawning a thread per park (the contention path is exactly where
+    an abandoned holder makes parks dense, so per-park threads would be a resource cliff, not a fix).
+    Each park is identified by a monotonic `uint64_t id`, not its entry's address: `LockTracker::wakeCallbacks`
+    has no removal API (see the gap noted below), so a stale closure can outlive its entry, and an
+    address-keyed lookup risks resolving a _different_, later park that reused the same freed heap
+    address. The timeout thread and the LockTracker wake callback race through one heap-allocated
+    `std::atomic<bool>` per park (independent of the per-park `RetryNowContext`, whose refs/TSFN the
+    winning side's release eventually frees) — whichever fires first calls+releases the TSFN under
+    `parkTimeoutMutex` and erases the entry; the loser finds it already gone and touches nothing. That
+    same mutex is what a dying env's `releaseParkTimeoutsByEnv` (wired into the module env-cleanup
+    hook next to `ReleaseCommitCompletionsByEnv`) takes to cancel — release without calling — that
+    env's pending parks before Node frees their tsfns; `retryNowCallJs` also guards `env == nullptr`
+    like `commitCompletionCallJs` does, for the same tsfn-queue-drained-during-teardown reason. The
+    LockTracker wake closure captures a **`std::weak_ptr<DBDescriptor>`**, not a raw pointer: a park
+    can end up registered on a tracker installed by a _different_ database on a colliding VT slot
+    (`VerificationTable::lockSlotForWrite` joins an existing tracker without retagging its `dbId`), so
+    that lock's eventual release wakes a park whose own descriptor may have already closed and been
+    destroyed — `cancelForDB()` only wakes trackers tagged with _its own_ `vtEpoch`, so it cannot be
+    relied on to have resolved a foreign-`dbId` park before the descriptor goes away. `.lock()` failing
+    is the expected outcome once that happens: `shutdownParkTimeouts()` (called from `finishClose()`
+    right after `cancelForDB`, before the descriptor can be destroyed) unconditionally resolves every
+    park still in its own `parkTimeouts` regardless of whether the real holder ever wakes it, so by the
+    time the weak reference can fail to lock, the park has already settled. Deliberately a plain
+    `std::thread`, not a `uv_timer_t`: this addon ships one prebuilt binary across Node ABI versions via
+    N-API, and libuv's struct layout is not part of that stable surface. Known gap: `LockTracker::wakeCallbacks`
+    itself has no removal API. Before this change an abandoned holder accrued one inert callback per
+    waiter and then everything hung; now each waiter re-parks (and re-registers) every
+    `ROCKSDB_JS_PARK_TIMEOUT_MS` up to `maxRetries`, so registrations accumulate per _retry_ rather than
+    per incident for as long as it lasts (each is inert once its own park resolves, so this is a
+    memory-growth concern, not a correctness one) — deferred rather than risking an unreviewed change to
+    `verification_table.cpp`'s concurrency invariants under this fix's scope.
 
 12. **A dropped transaction must release itself**: `DBDescriptor::transactionAdd` holds a **strong**
     `shared_ptr` (the parallel `closables` entry is weak), so the registry alone keeps a
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -196,10 +196,14 @@ sufficient (env teardown does not honor tsfn acquire counts); see
 - `ROCKSDB_JS_PARK_TIMEOUT_MS` - Bounded wait (default `5000`) before a
   coordinated-retry commit parked on a conflicting holder's VT lock resolves
   RETRY_NOW unconditionally, in case the holder never releases (see
-  "Coordinated retry" note below). Read per park; values below `50` are clamped
-  up to it, and `0` (an ambiguous "disable the bound") falls back to the default
-  like any malformed value. There is no opt-out: a deployment that would rather
-  wait than fail a legitimately slow holder raises the value instead
+  "Coordinated retry" note below). Read once per process (a function-local
+  `static`, like the other two above — `::getenv` is not safe against a
+  concurrent `::setenv` from a `process.env` write, and a park runs on whichever
+  env's JS thread owns the transaction), so it must be set in the environment a
+  process is started with. Values below `50` are clamped up to it, and `0` (an
+  ambiguous "disable the bound") falls back to the default like any malformed
+  value. There is no opt-out: a deployment that would rather wait than fail a
+  legitimately slow holder raises the value instead
 
 ## Test Structure
 
--- a/src/binding/database/db_descriptor.cpp
+++ b/src/binding/database/db_descriptor.cpp
@@ -10,6 +10,7 @@
 #include "rocksdb/utilities/options_util.h"
 #include <algorithm>
 #include <memory>
+#include <system_error>
 #include <unordered_map>
 
 namespace rocksdb_js {
@@ -586,6 +587,18 @@ uint64_t DBDescriptor::scheduleParkTimeout(
 		// registering with the LockTracker at all (see the header comment).
 		return 0;
 	}
+	if (!this->parkTimeoutThreadStarted) {
+		try {
+			this->parkTimeoutThread = std::thread([this]() { this->runParkTimeoutLoop(); });
+		} catch (const std::system_error&) {
+			// Thread creation failed (e.g. thread/resource exhaustion): leave
+			// the flag false so the next park retries, and tell the caller to
+			// resolve inline now rather than register a park nothing will
+			// ever time out.
+			return 0;
+		}
+		this->parkTimeoutThreadStarted = true;
+	}
 	auto entry = std::make_unique<ParkTimeout>();
 	entry->id = this->nextParkTimeoutId++;
 	entry->deadline = std::chrono::steady_clock::now() + std::chrono::milliseconds(timeoutMs);
@@ -593,11 +606,7 @@ uint64_t DBDescriptor::scheduleParkTimeout(
 	entry->tsfn = tsfn;
 	entry->fired = std::move(fired);
 	uint64_t id = entry->id;
-	this->parkTimeouts.push_back(std::move(entry));
-	if (!this->parkTimeoutThreadStarted) {
-		this->parkTimeoutThreadStarted = true;
-		this->parkTimeoutThread = std::thread([this]() { this->runParkTimeoutLoop(); });
-	}
+	this->parkTimeouts.emplace(id, std::move(entry));
 	this->parkTimeoutCv.notify_all();
 	return id;
 }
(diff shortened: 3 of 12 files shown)


src/binding/transaction/transaction.cpp
```diff
@@ -422,25 +475,82 @@ static void completeCommitWork(napi_env env, TransactionCommitState* state) {
...
+						// the shape that can make a racing close()'s
+						// PurgeIfUnreferenced observe use_count() > 1 and skip
+						// (HarperFast/rocksdb-js#672) -- retry it after
+						// dropping our ref, like BackupState/checkpoint state
+						// do for the same reason.
+						d->fireParkTimeout(parkId);
+						std::string path = d->path;
+						bool readOnly = d->readOnly;
+						d.reset();
+						DBRegistry::PurgeIfUnreferenced(path, readOnly);
```
cb1kenobi: **Critical: `PurgeIfUnreferenced` runs under the process-global VT `writerMutex_` — self-deadlock in `finishClose`**

`LockTracker::wake()` invokes its callbacks inline, and both of its callers hold `writerMutex_` for the whole function — `VerificationTable::releaseWriteIntent` (`verification_table.cpp:237`) and `cancelForDB` (`:300`). So this entire `fireOnce` body, `PurgeIfUnreferenced` included, executes with the process-global VT writer mutex held.

If the purge claims the descriptor (`use_count() <= 1 && beginClose()` — exactly the transient-`.lock()` window the comment above cites as the reason for the retry), it calls `descriptor->finishClose()`, which takes `txnsMutex` (`db_descriptor.cpp:452`) and then calls `vt->cancelForDB(vtEpoch)` (`:478`) → `std::lock_guard` on the **same non-recursive `writerMutex_` this thread already holds**. That is UB / a hard self-deadlock, and it wedges every database's write-intent path process-wide — the same symptom (harper#2001) this PR exists to fix. There is also a plain AB-BA: `finishClose` takes `txnsMutex` → `writerMutex_`, while this path takes `writerMutex_` → (`databasesMutex`) → `txnsMutex`. And `finishClose` does a flush, a manual compaction, `WaitForCompact`, and thread joins — all of that would run under the global VT lock.

Even when the purge does *not* claim, every wake now acquires the registry's `databasesMutex` while holding the global VT `writerMutex_` — a new global-lock nesting on the conflict-release path. Because `wakeCallbacks` has no removal API, this also runs for parks that already resolved via the timeout, so the exposure is every stale callback, not just live parks.

Suggested fix: treat "a `LockTracker` wake callback runs under `writerMutex_`, so it must not block and must not re-enter the VT or the registry" as the invariant. Drop the `PurgeIfUnreferenced` retry from `fireOnce` and perform it somewhere off the wake path — or avoid the transient ref entirely by having `scheduleParkTimeout` keep the descriptor reference for the park's lifetime, so there is no purge-skip window to retry. Worth adding that rule to AGENTS.md note 12 next to the rest of the design notes.

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
```diff
@@ -422,25 +475,82 @@ static void completeCommitWork(napi_env env, TransactionCommitState* state) {
...
+			// Exactly-once gate: independent of `ctx` so a loser arriving after
+			// the winner's ctx/tsfn are already gone just loses the CAS.
+			auto fired = std::make_shared<std::atomic<bool>>(false);
+
+			// #741 bound (harper#2001): resolves RETRY_NOW after
+			// ROCKSDB_JS_PARK_TIMEOUT_MS if the holder never releases. 0 both
+			// when the descriptor is closing and when there is no descriptor
+			// at all (DBHandle::close() can reset it concurrently) -- either
+			// way there is no timeout thread behind this park.
+			uint64_t parkId = descriptor ? descriptor->scheduleParkTimeout(env, parkTimeoutMs(), tsfn, fired) : 0;
```
cb1kenobi: **Medium: a timeout-driven `RETRY_NOW` spends a retry attempt, so a merely-slow holder now fails the write**

`database.ts:981-1025` caps coordinated retries at `maxRetries` (default 3) and then rejects with `Transaction did not commit after 3 coordinated retries`. Since the timeout resolves `RETRY_NOW` indistinguishably from a genuine wake, a holder that is slow rather than abandoned (large batch commit, compaction backpressure) burns all three attempts in ~15s and the write is **rejected** — where before this change it parked once and then succeeded. The loop is bounded, so there is no spin, but there is no backoff and no way for JS to tell a synthetic wake from a real one.

This ships on by default and `ROCKSDB_JS_PARK_TIMEOUT_MS` gives no escape hatch: `0` folds back to the 5000ms default, so the bound cannot be disabled for a deployment that would rather park than fail.

Suggested fix: mark the timeout-driven resolve (a distinct sentinel, or a flag alongside `RETRY_NOW`) so `database.ts` can re-park without consuming an attempt, or retry it with backoff. Failing that, give operators a real opt-out. The PR body documents this as deferred, but since it is default-on it is worth an explicit decision rather than a note.

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
```diff
@@ -211,6 +224,44 @@ static void retryNowFinalize(napi_env env, void* finalizeData, void* /*hint*/) {
...
+/**
+ * Bounded wait (ms) for a coordinated-retry commit parked on a conflicting
+ * holder's VT lock, selected by ROCKSDB_JS_PARK_TIMEOUT_MS (harper#2001).
+ * Default is the top of the 2-5s range this was scoped to: a timeout
+ * consumes a coordinatedRetry attempt like a genuine wake would, so a
+ * higher default leaves more headroom for a merely-slow (not abandoned)
+ * holder before maxRetries exhausts. Malformed input falls back to the
+ * default rather than producing a degenerate effective timeout.
+ */
+static unsigned parkTimeoutMs() {
```
cb1kenobi, line 251: **Low: process-wide `static` timeout, and no lower bound on the parsed value**

`ms` is a function-local `static`, so `ROCKSDB_JS_PARK_TIMEOUT_MS` is read once per process at the first park. Two consequences:

- Nothing can vary it per test, which is why the only regression test has to burn ≥5s of real wall clock and assert the loose `[4500, 10000)` window. A per-call read would let it use ~200ms and assert tightly, and would take 5s off every CI run.
- There is no floor. `ROCKSDB_JS_PARK_TIMEOUT_MS=1` is accepted and makes every conflict exhaust `maxRetries` within a few milliseconds — as degenerate as the `0` case that is (correctly) rejected two lines below.

The rejection of `0` / negative / non-numeric / `ERANGE` / `> UINT32_MAX` all fall back to the default, which is the safe direction. Suggested: clamp the accepted value to a sane minimum (e.g. 50ms), and read the env per call so tests can drive it.

—
_Generated by Barber AI_

src/binding/database/db_descriptor.cpp
```diff
@@ -565,6 +575,158 @@ void DBDescriptor::releaseCommitCompletionsByEnv(napi_env env) {
...
+	std::unique_lock<std::mutex> lock(this->parkTimeoutMutex);
+	for (;;) {
+		if (this->parkTimeoutStopped) {
+			return;
+		}
+		if (this->parkTimeouts.empty()) {
+			this->parkTimeoutCv.wait(lock);
+			continue;
+		}
+		auto earliest = std::min_element(
```
cb1kenobi: **Low: O(N) `min_element` under `parkTimeoutMutex` partially undoes the map's stated rationale**

The `ParkTimeout` header comment picks `unordered_map` over a vector specifically so `fireParkTimeout`'s lookup stays O(1) while `LockTracker::wake()` holds the process-global VT `writerMutex_`. But this loop re-scans the whole map with `min_element` on every wakeup, and `scheduleParkTimeout` calls `notify_all()` on every registration — so registering N parks costs O(N²) scanning, on the same mutex `fireParkTimeout` must acquire while holding `writerMutex_`. Under an abandoned holder, where parks are dense by design, that is the contention the map choice was meant to avoid.

Suggested fix: keep a `std::multimap<time_point, uint64_t>` (or a min-heap of `{deadline, id}`) alongside the id map so the earliest deadline is O(1), and skip the `notify_all` when the new deadline is not earlier than the current earliest.

—
_Generated by Barber AI_

test/lock-tracker.test.ts
```diff
@@ -168,6 +176,42 @@ describe('Coordinated retry (Phase 3)', () => {
...
+			await db.put(key, valueWithVersion(2.2e12));
+			txn.putSync(key, valueWithVersion(2.3e12));
+
+			const start = Date.now();
+			const result = await txn.commit();
+			const elapsed = Date.now() - start;
+
+			expect(result).toBe(RETRY_NOW);
+			expect(elapsed).toBeGreaterThanOrEqual(4500);
+			expect(elapsed).toBeLessThan(10000);
```
cb1kenobi: **Low: join-at-close is unverified — mutation results**

I mutation-tested this file against a locally built addon (`node-gyp build` clean each time, banner stable at `rocksdb-js/2.7.1 RocksDB/11.8.1`, runs serialized):

| mutation | result |
|---|---|
| `scheduleParkTimeout` deadline → `now + 1h` (timeout never fires) | **red** — `Test timed out in 30000ms`, 1 failed / 10 passed |
| exactly-once gate removed (no erase, no CAS in `fireParkTimeout` + `runParkTimeoutLoop`) | **red** — `SIGABRT`, vitest exit 134, on `resolves with RETRY_NOW ... on IsBusy` |
| `shutdownParkTimeouts()` `join()` → `detach()` | **green — 11/11 passed** |

So the deadline and the once-guard are genuinely covered (the once-guard only as a hard crash rather than an assertion, but it does go red). Nothing exercises join-at-close, which is the path where the timeout thread can outlive `parkTimeouts` / `parkTimeoutMutex`. The PR body already lists this as deferred — flagging it on the diff so it stays visible. A `db.close()` with a park still outstanding (the `test/commit-teardown.test.ts` child-process harness pattern) would cover it, and it gets much cheaper to write if the deadline becomes configurable per test (see the note on `parkTimeoutMs()`).

—
_Generated by Barber AI_

.github/workflows/pr.yml
```diff
@@ -214,5 +214,12 @@ jobs:
...
+        # macos-latest gets an extra retry for a known pre-existing native crash in
+        # commit-teardown.test.ts (HarperFast/rocksdb-js#746, mutex lock failed:
+        # Invalid argument during worker-env teardown vs. the shared commit thread)
+        # that has hit multiple unrelated PRs on this OS/runtime combo. Not a fix
+        # for #746 itself, just enough retry budget that a rare pre-existing race
+        # doesn't block unrelated PRs -- same pattern as the windows-latest retry
+        # below (HarperFast/rocksdb-js#695).
         shell: bash
-        run: pnpm test:deno || ([ "${{ matrix.os }}" = "windows-latest" ] && pnpm test:deno)
+        run: pnpm test:deno || ([ "${{ matrix.os }}" = "windows-latest" -o "${{ matrix.os }}" = "macos-latest" ] && pnpm test:deno)
```
cb1kenobi: **Low: the macOS retry re-rolls the whole Deno suite, not just the flaky case**

`pnpm test:deno` is retried in full, so a genuine macOS-only regression a future PR introduces also gets a second roll of the dice — this buys tolerance for #746 at the cost of that OS's signal on every unrelated failure. Since #746 is narrowed to `commit-teardown.test.ts`, scoping the retry keeps both properties: give that case a per-test `retry` (vitest `it(..., { retry: 1 })` or a config entry for that file), or retry only that file, so anything else still fails on the first run.

—
_Generated by Barber AI_

test/lock-tracker.test.ts
```diff
@@ -168,6 +176,42 @@ describe('Coordinated retry (Phase 3)', () => {
...
+			// below, so RocksDB's optimistic conflict check has something to
+			// validate against.
+			const txn = new Transaction(db.store, { coordinatedRetry: true });
+			await txn.get(key);
+			await db.put(key, valueWithVersion(2.2e12));
+			txn.putSync(key, valueWithVersion(2.3e12));
+
+			const start = Date.now();
+			const result = await txn.commit();
+			const elapsed = Date.now() - start;
```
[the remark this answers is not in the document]
kriszyp: Addressed in 4bcd981f: the elapsed-time assertions now use performance.now(), which is monotonic and unaffected by wall-clock adjustments.

src/binding/transaction/transaction.cpp
kriszyp: Addressed in 2a5ae5be and follow-ups: the wake callback no longer captures or re-enters DBDescriptor/DBRegistry. It holds weak ParkTimeoutRegistry and fired-state references only, so it cannot call finishClose()/cancelForDB() while VerificationTable::writerMutex_ is held, and stale callbacks return before taking the registry mutex.

src/binding/transaction/transaction.cpp
kriszyp: Reviewed as an explicit policy decision. The timeout intentionally consumes a finite retry and ROCKSDB_JS_PARK_TIMEOUT_MS=0 intentionally falls back to the default: this fix guarantees eventual settlement, with no opt-out that restores the production hang. Operators can raise the timeout for slow-holder workloads. The tradeoff is documented in AGENTS.md and called out under Human reviewer notes in the PR body.

src/binding/transaction/transaction.cpp
kriszyp: Addressed in 2a5ae5be: the environment variable is read per park rather than cached in a function-static value, positive values below 50 ms are clamped to 50 ms, and zero/negative/malformed/overflow values fall back to the 5000 ms default.

src/binding/database/db_descriptor.cpp
kriszyp: Addressed in 2a5ae5be: the registry now maintains a deadline-ordered multimap plus an id-to-entry map, so selecting the next deadline is O(1) and removing/firing a park is logarithmic instead of scanning all parks under the mutex.

test/lock-tracker.test.ts
kriszyp: Addressed in 4bcd981f: the regression test closes the database with a park live and asserts the commit settles before the timeout bound, exercising wake-during-close and joining the descriptor-owned timeout thread. The PR body explicitly records the remaining live worker-env teardown coverage gap for human review.

.github/workflows/pr.yml
kriszyp: Addressed in 4bcd981f: the workflow-wide macOS Deno retry was reverted. The known #746 suppression is now retry: 1 on only the affected commit-teardown test, and only when running Deno on macOS.

kriszyp (commented):
Addressed all seven unresolved review threads. The PR body now records the verification results, remaining coverage gap, and explicit timeout-policy decisions for human review.

src/binding/transaction/transaction.cpp
```diff
@@ -211,22 +225,96 @@ static void retryNowFinalize(napi_env env, void* finalizeData, void* /*hint*/) {
...
+ * the conflict path, not a hot path) so a process can be started with a small
+ * value for tests without depending on which park happens to run first.
+ * Malformed input falls back to the default rather than producing a
+ * degenerate effective timeout; a positive value below kMinimum is clamped up
+ * to it instead, since the intent there is unambiguous.
+ */
+static unsigned parkTimeoutMs() {
+	constexpr unsigned kDefault = 5000;
+	constexpr unsigned kMinimum = 50;
+	const char* v = ::getenv("ROCKSDB_JS_PARK_TIMEOUT_MS");
```
cb1kenobi: **Low: the per-park `getenv` is thread-unsafe, and a `static` would have served the test just as well**

The move off the function-local `static` (my earlier note on this function) fixed the "can't vary it per test" half, but `getenv` is not safe against a concurrent `setenv`. Node's `process.env` writes on the main thread go through `uv_os_setenv` → `setenv(3)`, which may reallocate `environ`; a worker thread sitting in this `getenv` while that happens can be handed a pointer into freed memory. Every commit that hits the coordinated-retry path now takes that read, on whatever thread owns the transaction's env.

The reason given in the comment above — "so a process can be started with a small value for tests" — does not actually require the per-call read: the new regression spawns a **child process** (`test/fixtures/fork-park-timeout.mts`) with `ROCKSDB_JS_PARK_TIMEOUT_MS=1` already in its environment, so a first-park `static` initializer reads the right value. And `test_seam.h` notes that worker-thread `process.env` writes never reach `::getenv` at all, so a per-call read buys no in-process variability either.

Suggested fix: restore the `static const unsigned ms = []{...}()` initializer and keep the new `kMinimum` clamp and `0`-is-malformed rule inside it. The fork fixture keeps working unchanged, and the read happens once per process on a thread-safe path.

—
_Generated by Barber AI_

test/lock-tracker.test.ts
```diff
@@ -168,6 +180,82 @@ describe('Coordinated retry (Phase 3)', () => {
...
+				});
+			});
+
+			expect(signal).toBeNull();
+			expect(code).toBe(0);
+		}));
+
+	// Covers the wake-during-close path (the holder's intents are released from
+	// inside finishClose(), so the park's LockTracker callback runs while the
+	// close is in progress) and the timeout thread being joined with a park
```
cb1kenobi: **Low: this still does not cover the join — re-ran the mutation at `88580c00`**

The comment above claims this test covers "the timeout thread being joined with a park live". I re-ran the same mutation against a clean local build of this head (`node-gyp build`, mutation confirmed in the artifact — `nm -u .../db_descriptor.o` shows `std::__1::thread::detach()` where the clean build shows `std::__1::thread::join()`):

| mutation in `ParkTimeoutRegistry::shutdown()` | result |
|---|---|
| `toJoin.join()` → `toJoin.detach()` | **green — 12/12 passed** |

That is the same result as at `07b2fd9f`. `shutdown()` resolves every pending park under `mutex` *before* the join, so the promise settles on time whether or not the thread is ever joined — the assertion here can't tell the two apart. What a detached thread actually costs is a `runLoop()` that can outlive its `ParkTimeoutRegistry` and touch freed `this`, which is a teardown/ASan-shaped failure, not a timing one.

Also worth recording: I rebuilt with `src/binding/` at the merge base (`23f21726`, artifact confirmed to have zero `ParkTimeoutRegistry` symbols) and ran this file. The fork-based timeout regression above **fails on base** (child SIGTERM'd at its 4s guard) — genuine coverage. **This** test passes on base too, so it is not a regression test for anything the PR adds; it exercises the close path but nothing about it is new-code-specific.

Not a blocker, and the code itself does join correctly. Flagging only so the comment's coverage claim doesn't get taken at face value later.

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
cb1kenobi: Confirmed resolved at `88580c00` — verified in the code and in the built artifact, not from this thread.

- `nm -u build/Release/obj.target/rocksdb-js/src/binding/transaction/transaction.o | c++filt` lists **42** undefined `rocksdb_js::` symbols and **none** of them is a `DBRegistry::` symbol. The only registry calls left from this translation unit are `ParkTimeoutRegistry::schedule` and `::fire`. The re-entrant leg is gone at link level, not merely guarded.
- Negative control for that method: injecting `DBRegistry::PurgeIfUnreferenced("CB1_NEGCTRL_744B_MARKER", false)` back into `fireOnce` and rebuilding makes both the undefined symbol and the marker string appear (`strings` on the `.node`). Reverting and rebuilding removes both. So the absence above is real, not a detection gap.
- `ParkTimeoutRegistry` is `std::make_shared`, i.e. its own control block — the `__shared_ptr_emplace<rocksdb_js::ParkTimeoutRegistry, ...>` symbol is in the binary. That is what makes `weakParks.lock()` structurally unable to perturb `DBDescriptor::use_count`, so the #672 purge-skip window this used to compensate for no longer exists rather than being narrowed.
- Disassembling `fire`/`take`/`resolve` in the linked `.node`: the only `rocksdb_js` call targets are `take()` and `~unique_ptr<ParkTimeout>`; everything else is libSystem/N-API. So the wake path's lock order is now `writerMutex_` → `ParkTimeoutRegistry::mutex`, and that mutex is a leaf — nothing acquires it and then reaches for `writerMutex_`, `databasesMutex`, or `txnsMutex`. Both the self-deadlock and the AB-BA are eliminated, and no flush/compaction/`WaitForCompact`/join can run under the global VT lock any more.
- `DBRegistry::ReleaseParkTimeoutsByEnv` now snapshots `entry.descriptor->parkTimeouts` instead of the descriptor, so env cleanup no longer inflates the `use_count` either.

Worth keeping in mind for future edits: `ParkTimeoutRegistry::shutdown()`'s `if (stopped && !threadStarted) return;` early exit is load-bearing. `finishClose()` always shuts the registry down before the descriptor is destroyed, so by the time a wake-path `weakParks.lock()` could be the last owner, the destructor's `shutdown()` is a no-op — which is what keeps a `std::thread::join()` from ever running under `writerMutex_`.

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
cb1kenobi: Acknowledged as a deliberate policy call — leaving this open for human sign-off rather than re-filing it, since it is a product decision and not a defect.

Re-verified at `88580c00` that the behaviour is unchanged: `src/database.ts` and `src/transaction.ts` are untouched by this PR (`git diff 23f21726..88580c00 -- src/database.ts src/transaction.ts` is empty), so `database.ts:987-1026` still caps at `maxRetries` (default 3) and abandons with `Transaction did not commit after 3 coordinated retries`, and JS still cannot distinguish a synthetic wake from a real one. A holder that is slow rather than abandoned still fails the write after roughly 3 × 5s.

One thing the new commits do improve here: `parkTimeoutMs()` is read per park now, so raising `ROCKSDB_JS_PARK_TIMEOUT_MS` no longer depends on which park happened to run first in the process.

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
kriszyp: Agreed — fixed in 48f985cf. `parkTimeoutMs()` is back to the `static const unsigned ms = []{...}()` form used by `commitThreadMode()`/`commitDelayMs()` in the same file, with the 50 ms clamp and the zero-is-malformed rule moved inside the initializer. Your reasoning that the per-call read bought nothing is right: worker-thread `process.env` writes never reach `::getenv`, so the only way to vary this is a child process started with the value set, which is exactly what `test/fixtures/fork-park-timeout.mts` does — it still passes in 369 ms (it would need ≥5 s if the env value were ignored, and its `elapsed >= 40` lower bound still pins the clamp). AGENTS.md's env-var entry now says read-once-per-process and why.

— Claude Opus 5

test/lock-tracker.test.ts
kriszyp: You're right, and I reproduced it rather than taking it on faith: mutating `toJoin.join()` to `toJoin.detach()` in `ParkTimeoutRegistry::shutdown()`, rebuilding (`nm -u db_descriptor.o` confirms `std::thread::detach()` replacing `std::thread::join()`) and re-running the file gives 12/12 green. `shutdown()` resolves every pending park under `mutex` before the join, so the deadline assertion cannot discriminate the two.

Fixed the claim rather than the test, in 87e7378c and trimmed in a45e8493: the comment now says it catches a hang on the wake-during-close path and names the two things a deadline cannot see (the join — whose real cost is a detached `runLoop()` touching a freed registry, an ASan-shaped failure — and the still-pending-park drain). Real join coverage needs the ASan lane, which I've logged as a follow-up rather than growing this PR.

— Claude Opus 5

cb1kenobi:
Reviewed `a45e8493` — no issues found in the incremental diff. This PR looks good, nice job!

Re-review scope: `88580c00..a45e8493` is a fast-forward (both heads share merge base `23f21726`, verified with `git range-diff` — 11 commits identical, 3 new). Genuine new work: 3 commits / 3 files, +53/-38.

Both of my remaining Low findings are addressed:

- **`getenv` thread-safety** — `48f985cf` restores the function-local `static`, matching `commitThreadMode()`/`commitDelayMs()`, with the 50 ms clamp and zero-is-malformed rule inside the initializer. Confirmed the only consumer of the env var is a spawned child process (`test/lock-tracker.test.ts:195`), so nothing depended on the per-park read; `lock-tracker` still runs 12/12 in 3.32s.
- **Coverage claim** — `87e7378c` + `a45e8493` correct it rather than overstating. I re-ran the mutation at this head to check: `toJoin.join()` → `toJoin.detach()` in `ParkTimeoutRegistry::shutdown()`, rebuilt, mutation confirmed in the artifact (`nm -u db_descriptor.o` shows `std::__1::thread::detach()`, absent from the clean build), still **12/12 green**. The corrected comment matches that result.

Re-verified from the prior round, since a later commit could have reintroduced it:

- The `PurgeIfUnreferenced`-under-`writerMutex_` self-deadlock is still structurally impossible here. `nm -u transaction.o` shows 42 undefined `rocksdb_js::` symbols and zero `DBRegistry::`; `PurgeIfUnreferenced` is absent. Validated with a negative control — re-injecting the call into `fireOnce` makes the symbol appear (42 → 43) and a marker string land in both the `.o` and the `.node`; reverting removes both. Note this is symbol-level and structural evidence, not a dynamic repro.
- Lock ordering is unchanged: `txnsMutex → writerMutex_` already exists at the merge base (`cancelForDB` under `txnsLock`), and `ParkTimeoutRegistry`'s mutex remains a leaf, so there is no ordering constraint against #787 in either direction.

Test results at this head: `lock-tracker` 12/12 (3.32s), full suite 807 passed / 1 skipped, CI 24 pass / 2 skipping. Against the merge base, the fork-based timeout regression fails (child SIGTERM'd at its 4s guard) while the close-path test passes — so the coverage split is exactly what the corrected comment now describes.

—
_Generated by Barber AI_

test/lock-tracker.test.ts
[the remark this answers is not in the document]
kriszyp: Following up because the reply above was not quite complete: the same commit added a second elapsed-time bound to this file (the concurrent-conflict wake check) that was still reading `Date.now()` on both ends, so one of the three assertions was not actually on a monotonic clock. Fixed in 65153100 — all three now use `performance.now()`.

— Claude Opus 5

test/lock-tracker.test.ts
```diff
@@ -168,6 +180,84 @@ describe('Coordinated retry (Phase 3)', () => {
...
+	it('a commit parked behind a never-releasing holder settles with RETRY_NOW within the deadline', () =>
+		dbRunner({ skipOpen: true }, async ({ dbPath }) => {
+			const { code, signal } = await new Promise<{
+				code: number | null;
+				signal: NodeJS.Signals | null;
+			}>((resolve, reject) => {
+				const child = spawn(process.execPath, [parkTimeoutFixturePath, dbPath], {
+					env: { ...process.env, ROCKSDB_JS_PARK_TIMEOUT_MS: '1' },
+				});
+				const timer = setTimeout(() => child.kill(), 4000);
```
cb1kenobi: **Medium: the parent's 4000 ms kill and the child's own 4000 ms ceiling are the same number, but the parent's clock starts earlier**

`fork-park-timeout.mts` asserts `elapsed < 4000` measured around `await transaction.commit()` only. This timer starts at `spawn()`, so its budget also has to cover node boot, `.mts` type-stripping, loading the 12 MB addon, `db.open()`, the seed put and the two transaction bodies. A child that satisfies its own assertion at `elapsed = 3900` is still killed here, and the failure surfaces as `signal === 'SIGTERM'` with the fixture's diagnostic `Error` never thrown — so a real regression and a slow runner look identical.

Measured locally on an idle M-series/14-cpu box the whole child is 820 ms, so the headroom is ~5x on the fastest environment this will ever run in. The suite already treats child spawn as the slow part on macOS/Windows — `commit-teardown.test.ts` budgets 120 s for the same shape.

The child's assertion is already the real deadline check; this timer only needs to be a backstop against a true hang.

Suggested fix:
```suggestion
				const timer = setTimeout(() => child.kill(), 30_000);
```

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
```diff
@@ -428,27 +520,96 @@ static void completeCommitWork(napi_env env, TransactionCommitState* state) {
...
+			uint64_t parkId = parkTimeouts
+				? parkTimeouts->schedule(env, parkTimeoutMs(), tsfn, fired)
+				: 0;
+
+			if (parkId == 0) {
+				// No timeout thread behind this park -- resolve now rather
+				// than register with the LockTracker unbounded.
+				vt->unrefTracker(t);
+				bool expected = false;
+				if (fired->compare_exchange_strong(expected, true)) {
```
cb1kenobi: **Low: this CAS can never fail — `fired` is uniquely owned on every path that reaches here**

All three paths that make `parkId` zero return before `fired` is stored anywhere: `parkTimeouts` null (never passed to `schedule`), `stopped` (returns at the top), and thread-creation failure (returns before `parks.emplace`). `schedule()` only publishes `fired` on the path that returns a non-zero id, so at this point the local `shared_ptr` has exactly one owner and no other thread can observe it. The exactly-once gate and the comment about `ParkTimeoutRegistry::resolve` describe a race that cannot occur here.

Suggested fix — drop the gate and keep the call+release:

```cpp
vt->unrefTracker(t);
if (::napi_call_threadsafe_function(tsfn, nullptr, napi_tsfn_nonblocking) == napi_ok) {
	::napi_release_threadsafe_function(tsfn, napi_tsfn_release);
}
parked = true;
break;
```

—
_Generated by Barber AI_

AGENTS.md
```diff
@@ -193,6 +193,17 @@ sufficient (env teardown does not honor tsfn acquire counts); see
   each completion callback (widens teardown race windows)
 - `ROCKSDB_JS_TXN_GET_DELAY_MS` - Test-only: delay a transaction's cold-cache async get before
   it reads (exercises orphan cleanup past the async-work wait timeout)
+- `ROCKSDB_JS_PARK_TIMEOUT_MS` - Bounded wait (default `5000`) before a
```
cb1kenobi, line 196: **Low: `README.md:774-780` still documents the unbounded park and is now wrong**

The knob is documented here, but the user-facing contract is in the README, which says the commit "waits until the conflicting transaction has committed and released its write intent." After this change that wait is capped at `ROCKSDB_JS_PARK_TIMEOUT_MS`, and because a timeout spends a `coordinatedRetry` attempt, a holder slower than roughly `maxRetries x 5000` (~15 s at defaults) now ends in `ERR_TRANSACTION_ABANDONED` where it previously waited and succeeded.

That policy is already adjudicated at `transaction.cpp:504` — this is only that the one place a user would look does not say it. AGENTS.md is the right home for the env var itself (no `ROCKSDB_JS_*` var appears in the README), but the behavior change belongs in the paragraph that describes the behavior.

Suggested fix — append to `README.md:780`:

> The wait is bounded: if the conflicting transaction never releases its write intent, the commit resolves after `ROCKSDB_JS_PARK_TIMEOUT_MS` (default `5000`) and consumes a retry attempt, so a holder that stays blocked for longer than roughly `maxRetries x` that timeout abandons the transaction rather than waiting indefinitely.

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `dc1bacfd` — no issues found in the incremental diff. This PR looks good, nice job!

**Re-review scope — derived with `git range-diff`, not the compare API.** Merge bases computed separately per head: `merge-base 65153100 origin/main` = `23f21726`, `merge-base dc1bacfd origin/main` = `da0ef135`. They differ, so this was a **rebase** onto 14 new `main` commits, not a fast-forward — which means the delta had to be checked for content that came from neither side. Three independent methods agree it is clean:

1. `git range-diff 23f21726..65153100 da0ef135..dc1bacfd` — 15 commits map 1:1, 1 appended. The four marked `!` (`15bbcb77`, `15088793`, `d09d9ff7`, `94906e72`) differ only in `AGENTS.md`/`transaction.cpp` **context** lines that `main` moved.
2. `git merge-tree --write-tree 65153100 da0ef135` conflicts in `AGENTS.md` and `transaction.cpp`. Diffing that auto-merge tree against `dc1bacfd^` shows the resolution keeps `main`'s `wrapperCollected` → `close()` block and this PR's relocation of the `RetryNowContext` transfer — both derivable from the two sides, nothing foreign introduced.
3. Comparing the PR's own net patch (added/removed lines only) between the two ranges differs by exactly **4 lines**, all of them the `AGENTS.md` invariant renumber (`12 → 13`, `13 → 14`) forced by `main` inserting its own invariant 12.

**Genuine new work: 1 commit, 2 files, +13/−4** — `dc1bacfd` "don't release a closing park timeout tsfn twice".

That commit is sound. Guarding `napi_release_threadsafe_function` on the preceding `napi_call_threadsafe_function` status is the pattern already used at `db_descriptor.cpp:1132-1136` and `:1172-1176`; the `== napi_ok` form is behaviourally identical to `!= napi_closing` here because the park tsfn is created with `max_queue_size = 0` (unlimited), so `napi_queue_full` is unreachable, and the exactly-once `fired` CAS means the thread count cannot reach zero before this call — so `napi_closing` at this point really only means env teardown, where Node destroys the tsfn itself and the skipped release cannot leak. (Small nit on the commit message only: `dispatchCommitCompletion`/`releaseCommitCompletionsByEnv` don't actually implement this pattern — the first never releases, the second releases without calling. `lockReleaseByKey` is the real precedent.)

**Prior findings re-checked against this head, in the artifact rather than from thread state:**

- **Critical (wake-path re-entry) — still fixed.** `fireOnce` still captures only `weak_ptr<ParkTimeoutRegistry>` + `weak_ptr<std::atomic<bool>>` + `parkId` and calls `fire(id)`. `nm -u build/Release/obj.target/rocksdb-js/src/binding/transaction/transaction.o | c++filt`: **175 undefined / 43 `rocksdb_js::` / 0 `DBRegistry::`**. The only registry symbols this TU references are `ParkTimeoutRegistry::schedule` and `::fire`. The 174→175 / 42→43 drift from the prior head is exactly one symbol, `TransactionHandle::close()`, which is `main`'s code arriving via the rebase. Negative control re-run through a real `make BUILDTYPE=Release` (not `node-gyp build`, which can silently no-op): re-injecting a `DBRegistry::PurgeIfUnreferenced` call with a unique marker takes the counts to 176/44/1 and puts the marker in both the `.o` and the `.node`; reverting restores a byte-identical symbol set and removes the marker from both. This is source + symbol-graph evidence with a validated detector — **not** a dynamic repro; forcing the interleaving would require mutating the code under review.
- **Low (join-at-close uncovered) — still open, and now conclusively so.** `toJoin.join()` → `toJoin.detach()` in `ParkTimeoutRegistry::shutdown()`, rebuilt and confirmed in the artifact (`nm -u db_descriptor.o` gains `std::__1::thread::detach()`, absent from the clean build): `lock-tracker` 12/12 green, and this time I also ran the **whole suite** against the mutated build — 60 files, 814 passed / 1 skipped, all green. So nothing anywhere in the suite kills it. The test comment at `test/lock-tracker.test.ts:219-227` says exactly this, so the claim matches the measurement; the gap is documented, not misstated.
- **Medium (a timeout-driven `RETRY_NOW` spends a `coordinatedRetry` attempt) — unchanged by this push.** `src/database.ts` and `src/transaction.ts` remain untouched by the entire PR at this head, `parkTimeoutMs()` is still the function-local `static` with the 5000 ms default, and no opt-out, backoff, or change to the retry accounting was added. It stays where it was: a documented policy decision for the human reviewer, not something for you to action.

**Testing.** Local build at `dc1bacfd`: `lock-tracker` 12/12 in 3.11s, `lock-tracker` + `commit-teardown` 14/14, full suite 60 files / 814 passed / 1 skipped. Merge base `da0ef135` built from scratch for comparison (`nm` confirms zero `ParkTimeoutRegistry` symbols in its `.node`): the same two files give 12/12. The PR's own test files are byte-identical between `65153100` and `dc1bacfd` — the `test/` delta between those heads is all `main`'s (`transaction-orphan-gc.test.ts`, `stats.test.ts`, a fixture), which is also what moves the suite total from 808 to 815 — so the earlier fails-on-base result for the fork-based timeout regression carries forward unchanged. CI at this head: 24 pass / 2 skipping — the full matrix is green.

One detection note for anyone repeating the join mutation: the linked `.node` carries an undefined `std::__1::thread::detach()` **even on a clean build** — it comes from `deps/rocksdb/lib/librocksdb.a`, not from this project (no project object references it, and there is no `.detach()` anywhere in `src/`). Only the `nm -u` on `db_descriptor.o` discriminates. That joins the two artifacts already known here: the residual `join()` traced to `commit_worker.h:110`, and the `DEBUG_LOG` marker being a release-build no-op (`debug.h:39`).

Generated by Barber AI

cb1kenobi (approved):
Tested. Looks good!

PR #745, opened 2026-08-01
fix: N-API use-after-free during worker-env teardown, plus two VT/transaction lifecycle races (#741)

kriszyp:
Human-Review-Need: 4 @ c740d2e26ace
## Root cause found: N-API calls are unsafe inside `napi_add_env_cleanup_hook`

Node's `Environment::RunCleanup()` runs `principal_realm_->RunCleanup(); cleanup_queue_.Drain();`
**in that order**. So by the time our module's env-cleanup hook runs, `Realm::RunCleanup()` has
already destroyed the env's `BaseObject`s and freed N-API per-env state.

`TransactionHandle::close()` guarded its `napi_delete_reference` with *"am I on the owning JS
thread?"* — but **env teardown runs on exactly that thread**, so the guard passed and the call
wrote through a freed env (`napi_delete_reference` → `napi_clear_last_error` writes `last_error`).
That corrupts glibc heap metadata; the abort then surfaces later and somewhere unrelated — usually
inside RocksDB's own allocators (`~Arena` / `~WriteBatchWithIndex` / `~TransactionBaseImpl`), which
is why this read as a RocksDB/VT bug for so long.

Confirmed with ThreadSanitizer against a from-source TSan Node — three heap-use-after-free writes
in `napi_clear_last_error`, reached from `TransactionHandle::close()` ← `DBDescriptor::finishClose()`
← `DBRegistry::Shutdown()` ← our cleanup hook.

**Fix:** `src/binding/napi/env_teardown.h` — a thread-local `EnvTeardownScope` set for the duration
of the cleanup hook; close paths check `isEnvTearingDown()` before making N-API calls. During
teardown the reference is reclaimed with the env anyway, so skipping is correct.

### Measured effect

`repro-crossthread.mjs` (`GRACEFUL=1 RECYCLE_MS=2000`, 4 workers, 15s, ~1M transactions/run), all in
one worktree with runs verified to have actually done work:

| build | result |
|---|---|
| baseline `ea83ff46` | 9/12 crashed |
| before this commit | 8/12 crashed |
| **with the fix** | **0/12 crashed** |

Still clean **8/8** at a harsher 800ms recycle. TSan goes from 3 races to 0.

## Also included: two VT/transaction lifecycle races

1. **TOCTOU on cross-env `close()` vs. in-flight async commit** — `close()` could delete `txn` while
   a commit that registered *after* `waitForAsyncWorkCompletion()` saw zero work was still using it.
   Fixed with a `stateMutex`, with `close()` flipping `closed` under it *before* draining.
2. **A parked coordinated-retry wake-callback's TSFN outliving its env** — fixed with
   `ParkedFlagRegistry`, mirroring the existing `commitCompletions`/`ReleaseCommitCompletionsByEnv`
   per-env teardown pattern.

Both were confirmed against gdb backtraces matching the issue's signatures. Note these two do *not*
move the repro's crash rate on their own (bisected) — the UAF above was what that repro was hitting.

## Known remaining (not root-caused)

`test/vt-lock-tracker-churn.test.ts` still fails ~25% (1/4 reps at 4000ms, 1/6 at 1500ms) with the
same `corrupted size vs. prev_size`, so it stays `it.skip`. It reproduces **only** through that
fixture's tsx-transpiled-worker path, never through the plain-`.mjs` repro, and TSan does not catch
it. Also still open: `getSync`/`putSync`/`removeSync`/`CommitSync`/`getCount` remain outside
`tryRegisterAsyncWork()` (AGENTS.md item 10a).

Because of that residual, I'd suggest keeping #741 open after this merges, scoped to what's left.

## Notes for reviewers

- **`db_handle.cpp` has further `napi_delete_reference` sites reachable from
  `DBRegistry::Shutdown()`** that this repro didn't exercise. They likely need the same
  `isEnvTearingDown()` guard — I left them alone rather than change code I couldn't test.
- AGENTS.md item 12 documents the root cause, the working TSan-Node recipe (needs
  `-DTHREAD_SANITIZER`; `V8_IS_TSAN` must stay **off** or V8 won't compile under
  `v8_enable_sandbox=0`), why ASan is structurally blind here, and a methodology warning about
  false-negative baseline runs that cost me significant time this session.
- `binding.gyp` gains a `ROCKSDB_TSAN=1` toggle alongside the existing `ROCKSDB_ASAN`.

## Test coverage

Full Vitest suite 710 passed / 2 skipped; native GoogleTest 104/104.

Refs #741


kriszyp:
Update on the blocking finding: spent another round investigating per Kris's request, ruling out three hypotheses and confirming ThreadSanitizer isn't practical here without a from-source Node build. Full detail in `AGENTS.md` item 12 (substantially rewritten in c86e78ef).

**Ruled out this round:**
1. VT's `writerMutex_` held across N-API wake-callback calls — moving `wake()` outside the lock didn't help.
2. Un-closed/leaked `TransactionHandle`s inflating `finishClose()`'s teardown backlog — this leak is real (confirmed via debug-log correlation: `completeCommitWork`'s non-`IsBusy` rejection path never calls `close()`), but patching the repro script to always `abort()` on rejection didn't reduce the crash rate.
3. Raw `TransactionHandle::close()` latency — tested decisively by injecting an artificial delay into `close()` on **unmodified baseline code** via the existing `ROCKSDB_JS_TXN_CLOSE_DELAY_MS` seam. Clean 20/20 at both 10ms and 100ms delays. This rules out "the fix just makes `close()` slower, widening some pre-existing window" as the explanation.

**ThreadSanitizer**: wired up (`ROCKSDB_TSAN=1` in `binding.gyp`), but running the repro under it produces 100-150+ races per run that all resolve to pure V8-internal GC/JIT machinery (confirmed via `addr2line`) — a known limitation of running vanilla Node under TSan without Node's own build-time instrumentation. A real attempt needs Node built from source with `-fsanitize=thread`.

Net: the regression traces to something in this fix's actual new logic (the `stateMutex` critical sections or `ParkedFlagRegistry`), not simply added latency in the teardown path. Still unresolved — still recommend not merging until it is.

kriszyp:
**Retraction of my earlier comment on this PR.**

The comment above claiming this fix converts the repro from "15/15 clean" to "14/15 crashing" is
**wrong**, and I've rewritten the PR description accordingly.

The baseline arm in that comparison was a false negative. It ran in a separate worktree where
`pnpm install` had run but `pnpm build:bundle` had not, so `dist/index.mjs` didn't exist. Every
worker failed with `Cannot find module`, did zero work, and the harness *still* printed
`RESULT: no stuck commit` and exited 0. All ~35 "clean baseline" data points — including the
10ms/100ms `ROCKSDB_JS_TXN_CLOSE_DELAY_MS` runs I used to "decisively rule out close() latency" —
were no-op runs.

Re-measured in a single worktree (identical RocksDB prebuild, node_modules and build config; only
`src/` differing; competing build paused; each run verified to have actually issued ~580k
transactions):

- baseline `ea83ff46`: **9/12 crashed**
- this PR: **8/12 crashed**

So the fix neither causes nor fixes that crash, and the crash matches the 7/12 rate #741 itself
reports. The real conclusion is less dramatic but more useful: **this PR closes two genuine races
but does not resolve #741's headline repro**, so #741 should stay open.

Lesson recorded in `AGENTS.md` item 12: a green result from these repro scripts is only meaningful
if the run actually did work — check per-worker `issued=` counts and grep for `Cannot find module`
before trusting any before/after comparison, and compare revisions within one worktree rather than
across worktrees.

Apologies for the noise from the incorrect earlier analysis.

kriszyp:
**Pre-push review round (DLC step 10) — outcome and what changed.**

`independent-review: FAILED — codex leg timed out (rc=-1 after 973s); same-family fallback skipped for budget. Gemini + Grok legs did produce findings and were acted on manually, but per the skill's rules this run does NOT count as independent coverage.`

Both non-Claude legs independently flagged one thing I had only *noted* rather than fixed, so I fixed it:

- **`DBHandle::close()` had the same teardown UAF** (Grok, major). It releases transaction-log JS refs via `napi_delete_reference` and is reachable from `DBRegistry::Shutdown()` inside the env cleanup hook — the same shape as the `TransactionHandle` path. Now guarded with `isEnvTearingDown()` (`b3e8130d`). Only the `close()` loop is guarded; `unrefLog()`/`useLog()` are called from live JS, not teardown.
- **Comment verbosity** (both legs, nit): trimmed the comments added for this fix down to the invariant they encode, per repo convention.

Findings I did **not** act on, with reasons:

- **`CommitSync`/`getSync`/`putSync`/`removeSync`/`getCount` bypass `tryRegisterAsyncWork()`** (Gemini: blocker; Grok: major). Real, but **pre-existing on `main`** — not introduced here — and documented as a deliberately deferred gap in AGENTS.md item 10a. An earlier attempt to close it on `CommitSync` reproduced a separate heap-corruption regression that wasn't root-caused, so it needs its own change with its own verification rather than being appended to this one.
- **`waitForAsyncWorkCompletion()`'s 5s timeout** (both legs, major). Also pre-existing and unchanged by this PR (item 10b). Removing it trades a UAF risk for a shutdown-hang risk; that's a design call worth making deliberately, not here.
- **`lockVTSlot` now takes `stateMutex` on the VT write path** (Gemini major / Grok minor, perf). Accepted, deliberate: it is the lock that makes the intent vectors safe against a concurrent cross-env `close()`. The hottest paths (`getSync`/`putSync`/`removeSync` themselves) were explicitly left unlocked for exactly this cost reason.
- **The churn test is `it.skip`** (both legs, major). Fair criticism and I agree it's a real coverage gap — but it still fails ~25% on a residual crash this PR doesn't fix (unchanged at 2/6 after the `db_handle` guard), and shipping it enabled would make CI flaky for an unrelated reason.

Post-fix verification: `repro-crossthread.mjs` 0/10, full suite 709 passed / 3 skipped, native 104/104.

cb1kenobi:
I have fixed the merge conflicts, but I have some reservations about this PR.

The description states "N-API calls are unsafe inside", but that's false. This PR adds the env-teardown guard and mutexes, and that's cool. But the friction is this is a 947-line three-race fix with an unresolved residual crash, a skipped test, and its only real evidence living outside the repo.

src/binding/transaction/transaction_handle.cpp
```diff
@@ -513,10 +598,21 @@ void TransactionHandle::getCount(
...
+	// An earlier attempt to add that registration to a sibling synchronous
+	// path (CommitSync) reproduced a heap-corruption regression under worker
+	// churn testing that wasn't root-caused within the fix for #741; this
+	// path shares that same new-registration shape and was reverted out of
+	// caution rather than risk the same defect untested.
 	this->ensureSnapshot();
 	if (this->snapshotSet) {
 		itOptions.readOptions.snapshot = this->txn->GetSnapshot();
 	}
+	}
```
cb1kenobi: **Critical: stray `}` — the current PR head does not compile**

Commit `a209f6b5` (an accepted `gemini-code-assist` suggestion) added the `if (!this->txn)` guard to `getCount()` but also left an extra closing brace here. It closes `getCount()` early, so the `itHandle` declaration and `for` loop below fall into namespace scope referencing `this` — a hard compile error, and the whole binding fails to build on this head.

Suggested fix — delete this line:
```suggestion
```

—
_Generated by Barber AI_

test/vt-lock-tracker-churn.test.ts
```diff
@@ -0,0 +1,74 @@
...
+	// 8/8 at an even harsher 800ms recycle.
+	//
+	// But this fixture still fails ~25% of the time (1/4 reps at the cadence
+	// below, 1/6 at 1500ms) with the same `corrupted size vs. prev_size`. That
+	// residual crash only reproduces through this tsx-transpiled-worker path,
+	// not through the plain-.mjs repro, and ThreadSanitizer did not catch it
+	// (0 races over 3 aggressive runs -- TSan's ~15x slowdown likely closes the
+	// window). It is NOT root-caused, so enabling this would make CI flaky for
+	// a reason this PR does not address. See AGENTS.md item 12.
+	it.skip(
```
cb1kenobi, line 69: **Medium: three teardown fixes ship with no runnable regression coverage**

This is the only test the PR adds and it's `it.skip`, because the ~25% residual crash it exercises is not root-caused. The evidence cited for the dominant fix (`repro-crossthread.mjs`, "9/12 → 0/12") is not committed to the repo, so neither a reviewer nor CI can reproduce it. The stateMutex, ParkedFlag, and env-teardown fixes have no passing automated test — nothing here catches a regression.

Suggested fix: commit the `repro-crossthread.mjs` scenario as an opt-in/long-form spawned-child test. The existing `ROCKSDB_JS_TXN_CLOSE_DELAY_MS` / `ROCKSDB_JS_COMMIT_DELAY_MS` seams make the close-vs-commit race deterministic without relying on the flaky fixture — at minimum, don't leave all three fixes uncovered.

—
_Generated by Barber AI_

src/binding/napi/env_teardown.h
```diff
@@ -0,0 +1,40 @@
+#ifndef __NAPI_ENV_TEARDOWN_H__
+#define __NAPI_ENV_TEARDOWN_H__
+
+namespace rocksdb_js {
+
+/**
+ * True while the calling thread is inside its env's cleanup hook.
+ *
+ * Node runs `Realm::RunCleanup()` — which frees N-API per-env state — before
```
cb1kenobi, line 9: **Medium: the documented root cause doesn't match Node's teardown ordering**

Checked against Node source: `Realm::RunCleanup()` only resets binding-data and BaseObjects — it does not free the `napi_env` struct where `last_error` lives. The `napi_env`'s own teardown (`Unref`→`DeleteMe`) is itself a cleanup-queue hook registered at env creation, and the queue drains in reverse (LIFO) order, so it runs *after* this module's hook. For the env being torn down, `last_error` is still valid memory when the hook runs.

A use-after-free *write* to `last_error` therefore implies the reference belongs to a *different, already-destroyed* worker env — i.e. the cross-env lifetime problem the `stateMutex`/`ParkedFlag` changes address, not "the realm freed my own env's N-API state." The `isEnvTearingDown()` guard is cheap and safe, so keep it — but please reconcile this explanation against the actual TSan/gdb trace (is the crashing env the current thread's, or foreign?) before enshrining it in AGENTS.md item 14, and don't treat #741 as closed on this basis.

—
_Generated by Barber AI_

src/binding/database/db_handle.cpp
```diff
@@ -148,10 +149,14 @@ void DBHandle::close() {
...
 	}
 
-	// clean up transaction log references
-	for (auto& [name, ref] : this->logRefs) {
-		DEBUG_LOG("%p DBHandle::close Releasing transaction log JS reference \"%s\"\n", this, name.c_str());
-		::napi_delete_reference(this->env, ref);
+	// Skip during env teardown: this is reachable from DBRegistry::Shutdown()
+	// inside the env cleanup hook, by which point Node has already freed the
+	// env's N-API state and the refs go with it (napi/env_teardown.h).
+	if (!isEnvTearingDown()) {
```
cb1kenobi, line 155: **Medium: sibling `napi_delete_reference` sites left unguarded by this fix's own logic**

This guard is correct, but `napi_delete_reference` is still called unconditionally in `DBHandle::unrefLog()` (line 371) and `DBHandle::useLog()` (line 392). The PR description itself notes db_handle.cpp "has further `napi_delete_reference` sites reachable from `DBRegistry::Shutdown()`" that were left alone — by this fix's own reasoning those are latent heap corruption during teardown.

Suggested fix: either wrap those sites in the same `isEnvTearingDown()` check, or confirm and document that they are unreachable from `DBRegistry::Shutdown()`. Shipping a self-documented reachable UAF defers the crash rather than fixing it.

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
```diff
@@ -188,6 +222,87 @@ static void retryNowFinalize(napi_env env, void* finalizeData, void* /*hint*/) {
...
+		// sweep when the vector is about to reallocate anyway (size ==
+		// capacity): this bounds sweeps to O(log n) over n registrations --
+		// the same amortized cost as the vector's own growth -- instead of
+		// an O(n) scan under this process-global lock on every single park.
+		if (flags.size() == flags.capacity()) {
+			flags.erase(
+				std::remove_if(flags.begin(), flags.end(), [](const std::weak_ptr<ParkedFlag>& w) { return w.expired(); }),
+				flags.end()
+			);
+		}
```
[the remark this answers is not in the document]
cb1kenobi: Fixed

src/binding/transaction/transaction_handle.cpp
cb1kenobi: Fixed

cb1kenobi:
Closing this PR in favor of https://github.com/HarperFast/rocksdb-js/pull/780.

Issue #749, opened 2026-08-05
Uncommitted transaction-log reads bound corruption checks by mapped capacity, so a torn frame with a plausible length goes undetected


kriszyp:
## Summary

An uncommitted transaction-log read bounds its per-entry corruption checks by the **mapped capacity** of the memory map rather than by the written extent. The map is pre-extended far past the data, so a torn frame only trips the check when its declared length exceeds that whole capacity. Any smaller garbage length is accepted, and the reader advances `position` by it — walking over valid entries, or handing back a misframed one, without ever reporting corruption.

## Where

`src/transaction-log-reader.ts`, in the `query()` iterator:

```ts
const limit = readUncommitted ? logBuffer!.length : Math.min(size, logBuffer!.length);
if (position + TRANSACTION_LOG_ENTRY_HEADER_SIZE > limit) { /* corrupt */ }
const length = dataView.getUint32(position + 8);
if (position + TRANSACTION_LOG_ENTRY_HEADER_SIZE + length > limit) { /* corrupt */ }
```

For a committed read `limit` is the committed watermark, which is a true entry boundary, so the checks are tight. For `readUncommitted` it is `logBuffer.length` — the mapped capacity.

## Why it matters

`readUncommitted` is the boot-replay path (Harper's `replayLogs.ts` reads with `startFromLastFlushed: true, readUncommitted: true`), which is precisely the reader that has to cope with a log damaged by an interrupted append.

The field error in HarperFast/harper#1977 declared a length of 1,688,494,450 — comfortably past any mapped capacity, so it was caught. A partial write that leaves a plausible-looking smaller length is not: the frame is silently accepted, and iteration continues from a position that is no longer an entry boundary.

This also gates corrupt-frame resync (HarperFast/harper#2016, #2063): recovery can only run where detection fires, so on this path resync is inert for the class of corruption whose declared length happens to land under the mapped capacity.

## Why it is not simply tightened

The looser bound is deliberate — an uncommitted reader is allowed to read past the `size` it captured at query time, because a concurrent append may have grown the file since. Replacing `logBuffer.length` with the extent captured at query start would report healthy, concurrently-appended entries as corrupt.

Fetching the current extent per entry is also not acceptable: `getLogFileSize` crosses into native and takes the store mutex, so it must not sit on the healthy read path (this was measured as a regression in PR review and moved off that path).

## Suggested direction

Check against a cached written extent and refresh it only on an apparent overrun:

[...]

Issue #748, opened 2026-08-05
A failed transaction-log append orphans its partial bytes, baking a mid-file framing break into the log


kriszyp:
## Summary

A transaction-log append that fails part-way through leaves the bytes it already wrote on disk, unaccounted for, and the next successful append lands *after* them. The result is a permanently torn frame in the middle of the file with valid entries on both sides — the one shape open-time recovery deliberately refuses to repair.

This is distinct from #572 (partial `writev` bookkeeping, fixed in 2.0.0/1.4.1) and from the torn tail that `recoverTail()` handles. Both of those are closed; this path is still live on `main`.

## Mechanism

Four pieces compose:

1. The log fd is opened `O_RDWR | O_CREAT | O_APPEND` (`transaction_log_file_posix.cpp:128`), so every write lands at the current physical end of file regardless of `TransactionLogFile::size`.
2. `writeBatchToFile` returns `-1` on a hard error and **discards `totalWritten`** (`transaction_log_file_posix.cpp`). Bytes that already reached the file are never reported to the caller.
3. `writeEntriesV1` throws on `-1` and does **not** advance `this->size`, and does **not** remove the bytes that landed (`transaction_log_file.cpp:365-371`).
4. The next successful append goes to physical EOF — i.e. after the orphaned partial bytes.

So a single failed append leaves:

- a partial frame permanently embedded mid-file, with well-formed entries after it once writing resumes; and
- `size` under-reporting the physical extent by the orphan length from that point on.

`recoverTail()` classifies this as `RecoveryScan::Kind::MidFileCorruption` and leaves it intact by design (`transaction_log_file.cpp:178`) — correctly, since truncating would discard committed entries after the break:

> Leave the file intact: entries are still framed after the break, so truncating would discard committed/replicated transactions. […] Reads past this point will fail until the file is repaired.

It also only ever runs on the active (highest-sequence) file, on the stated assumption that "rotated files are immutable and already complete" (`transaction_log_store.cpp:1078`). This path violates that assumption: the process survives the failed append and keeps writing, so the break gets baked into a file that later rotates and is never scanned again.

## Why this matters

[...]

PR #750, opened 2026-08-05
fix(txnlog): resync past a mid-log corrupt frame instead of ending the log

kriszyp:
Reader half of [HarperFast/harper#2016 — Mid-log corrupt transaction-log frame silently truncates replay and replication — acknowledged writes lost](https://github.com/HarperFast/harper/issues/2016) and [HarperFast/harper#2063 — Corrupt-log containment in RocksTransactionLogStore.getRange is per-drain, so one bad entry head-of-line-blocks a replication stream forever](https://github.com/HarperFast/harper/issues/2063). The consumer change is [harper#2087](https://github.com/HarperFast/harper/pull/2087).

## The defect

A framing break has two shapes and the reader treated them as one.

A **torn tail** really is end-of-log — `recoverTail()` truncates it at open. A **mid-log break** is not: a partial append (`ENOSPC`/`EDQUOT`) that the process survives leaves intact, already-acknowledged entries after it. `recoverTail()` deliberately leaves that file alone rather than discard them, and only ever scans the active file on the assumption that "rotated files are immutable and already complete" — an assumption this case violates, because the process keeps writing and the file later rotates. So the reader is the only thing standing between those entries and a consumer.

It threw and left `position` at the bad entry. Every later entry in the file was therefore unreachable, permanently: a consumer restarts from the same resume cursor on the next drain and re-throws at the same offset. #2016 rolled a table back 2.2 days on a crash-recovery replay; #2063 starved a replication stream for 11 days, clearing only when the log aged out of retention.

## The fix

`query()` reports a break as `CorruptFrameError` carrying `resyncPosition` — where valid framing resumes — plus the unreadable byte count, and advances `position` there *before* throwing. One loud report per break, and a caller that wants the entries past it calls `next()` again. A caller that treats the error as terminal behaves exactly as before.

The resume point uses the same heuristic as `validFramingResumes()` in the recovery scan, which already computes this answer at open time and keeps only the yes/no. With no resume point (a torn tail) `position` lands past both the readable end and `size`, so a retrying caller gets `done` rather than the same throw again.

**The scan is bounded by the written extent, not the read limit.** This is the load-bearing detail. An uncommitted read's limit is the pre-extended memory map, and every offset in that zero fill reads as an end-of-entries marker — so scanning against it loses the exact-end signal entirely, leaving only the eight-frame rule. A mid-log break followed by 1–7 entries would then be unrecoverable on exactly the boot-replay path (`replayLogs.ts` reads `readUncommitted`) that lost 2.2 days in #2016. `getLogFileSize()` returns the append-owned `TransactionLogFile::size` (invariant 5), not the physical or mapped size, so the end is a single offset on every platform including Windows, and a chain landing on it keeps the ~1/2³² argument. It is resolved only on a break — it crosses into native and takes the store mutex, so a per-frame call would tax every healthy read.

## Tests

`test/transaction-log.test.ts`, 69 passing. A 12-entry log with a broken second frame now yields **11 entries where the old reader delivered 1**. Plus the edges: a sub-eight-frame run recovered via the written extent, the same under `readUncommitted` against a zero-padded map, and a torn tail terminating when committed `size` over-reports the buffer.

## Worth a reviewer's attention

**The unbounded scan is a deliberate, contested call.** Round 2 of review asked for a 1 MiB cap to be removed (entries can legitimately be large, and a partial frame wider than the cap puts every entry behind it out of reach — the amputation this exists to prevent, reintroduced at a threshold). Round 3 then flagged that removing it raises the false-resync probability to roughly N/2³² over the scan span, and that a byte-by-byte scan over a large log is a one-off event-loop stall. Both directions are right; I landed on matching `validFramingResumes()`, whose identical exposure was adjudicated acceptable in round 1 because a false resume yields at most one garbage frame that re-fails framing, not silent acceptance of bad data. Worth a second opinion — this is the one place I'd expect a reviewer to disagree.

**Detection on the uncommitted path is still bounded by mapped capacity**, so a torn frame whose declared length lands under that capacity is not detected at all, and resync cannot run for it. Pre-existing, not introduced here, and deliberately left out of scope because tightening it means changing uncommitted-read semantics and risks reporting concurrently-appended entries as corrupt. Filed as #749 with a suggested approach. Note this affects the *boot-replay* path; replication broadcast uses committed reads, where the bound is already the watermark.

**No end-to-end proof.** Every test drives synthetic or mmap-substituted buffers. Nothing reopens a genuinely damaged log on disk and watches a real consumer replicate past the break — that lives in harper-pro and is not written. Whether #2016/#2063 are actually closed rests on the consumer change in harper#2087.

**This does not stop corrupt frames being created.** The write-side producer — a failed append orphaning its partial bytes, which `O_APPEND` then writes past — is #748.

---

Cross-model review: 3 rounds, codex + gemini + harper-domain each round. Round 1 caught that the scan was bounded by the read limit, which would have left the fix half-working on the boot-replay path. Round 2 caught that the resulting `getLogFileSize` call sat inside the read loop — a per-frame native call taking the store mutex. Both fixed here. Unresolved items are the three above. Round 3's remaining findings were repeats of already-adjudicated items, except the cap trade-off noted above.

Generated with Claude Opus 5.

<sub>Human-Review-Need: 4 @ f8457ea7ad79</sub>

### DIFF
--- a/src/binding/transaction/transaction.cpp
+++ b/src/binding/transaction/transaction.cpp
@@ -599,12 +599,10 @@ napi_value Transaction::Commit(napi_env env, napi_callback_info info) {
 	CommitThreadMode mode = commitThreadMode();
 	bool completionsClosed = false;
 	if (mode != CommitThreadMode::Legacy && dbHandle && dbHandle->descriptor) {
-		// The commit lanes are owned by the descriptor and drained/joined before
-		// the descriptor is destroyed; and an in-flight commit's state pins the
-		// descriptor alive (state -> txnHandle -> dbHandle -> descriptor). So a
-		// raw pointer captured into the worker task is valid for the task's
-		// lifetime and cannot form a reference cycle with the worker thread.
-		DBDescriptor* descriptor = dbHandle->descriptor.get();
+		// Keep the descriptor alive until every queued stage has completed. Worker
+		// teardown can close the transaction handle between the log and commit
+		// stages, releasing its DBHandle and its descriptor reference.
+		auto descriptor = dbHandle->descriptor;
 
 		// Ensure this env has a completion tsfn and account the dispatch (refs
 		// the tsfn as the env goes idle->busy so the event loop stays alive).
@@ -1223,4 +1221,3 @@ void Transaction::Init(napi_env env, napi_value exports) {
 }
 
 } // namespace rocksdb_js
-
--- a/src/binding/transaction/transaction.cpp
+++ b/src/binding/transaction/transaction.cpp
@@ -601,7 +601,9 @@ napi_value Transaction::Commit(napi_env env, napi_callback_info info) {
 	if (mode != CommitThreadMode::Legacy && dbHandle && dbHandle->descriptor) {
 		// Keep the descriptor alive until every queued stage has completed. Worker
 		// teardown can close the transaction handle between the log and commit
-		// stages, releasing its DBHandle and its descriptor reference.
+		// stages, releasing its DBHandle and its descriptor reference. The registry
+		// retains its reference until it drains and joins these lanes, so this task
+		// capture cannot be the last descriptor reference on a worker thread.
 		auto descriptor = dbHandle->descriptor;
 
 		// Ensure this env has a completion tsfn and account the dispatch (refs
--- a/src/transaction-log-reader.ts
+++ b/src/transaction-log-reader.ts
@@ -125,15 +125,22 @@ function corruptFrame(
 ): { error: CorruptFrameError; nextPosition: number } {
 	// An uncommitted read's `limit` is the pre-extended map, so ask the file for its written extent;
 	// a committed read is already bounded at the watermark.
-	const writtenExtent = readUncommitted ? transactionLog.getLogFileSize(logBuffer.logId) : 0;
-	const dataEnd = writtenExtent > 0 ? Math.min(logBuffer.length, writtenExtent) : limit;
+	let dataEnd = limit;
+	if (readUncommitted) {
+		try {
+			const writtenExtent = transactionLog.getLogFileSize(logBuffer.logId);
+			dataEnd = writtenExtent > 0 ? Math.min(logBuffer.length, writtenExtent) : 0;
+		} catch {
+			dataEnd = 0;
+		}
+	}
 	const resyncPosition = findResyncPosition(dataView, position + 1, dataEnd);
 	return {
 		error: new CorruptFrameError(message, logBuffer.logId, position, resyncPosition),
 		// With no resume point iteration is over, so land past `size` as well as the readable end:
 		// a committed `size` that over-reports the mapped buffer would otherwise leave
 		// `position < size`, and the next call would re-enter the loop and read off the end.
-		nextPosition: resyncPosition ?? Math.max(dataEnd, size, position + 1),
+		nextPosition: resyncPosition ?? Math.max(dataEnd || limit, size, position + 1),
 	};
 }
 
--- a/src/transaction-log-reader.ts
+++ b/src/transaction-log-reader.ts
@@ -99,7 +99,7 @@ function findResyncPosition(dataView: DataView, from: number, dataEnd: number): 
 	for (let start = from; start <= lastStart; start++) {
 		let pos = start;
 		let frames = 0;
-		while (frameFits(dataView, pos, dataEnd)) {
+		while (frames < RESYNC_MIN_FRAMES && frameFits(dataView, pos, dataEnd)) {
 			pos += TRANSACTION_LOG_ENTRY_HEADER_SIZE + dataView.getUint32(pos + 8);
 			if (++frames >= RESYNC_MIN_FRAMES || pos === dataEnd) {
 				return start;
--- a/src/transaction-log-reader.ts
+++ b/src/transaction-log-reader.ts
@@ -328,11 +328,15 @@ Object.defineProperty(TransactionLog.prototype, 'query', {
 						throw broken.error;
 					}
 					const length = dataView.getUint32(position + 8);
-					if (position + TRANSACTION_LOG_ENTRY_HEADER_SIZE + length > limit) {
+					if (length === 0 || position + TRANSACTION_LOG_ENTRY_HEADER_SIZE + length > limit) {
 						const broken = corruptFrame(
-							`Corrupt transaction log entry at position ${position.toString(16)} of log ${
-								logBuffer!.logId
-							}: declared length ${length} overruns the log (limit=${limit})`,
+							length === 0
+								? `Corrupt transaction log entry at position ${position.toString(16)} of log ${
+										logBuffer!.logId
+									}: zero-length entry`
+								: `Corrupt transaction log entry at position ${position.toString(16)} of log ${
+										logBuffer!.logId
+									}: declared length ${length} overruns the log (limit=${limit})`,
 							transactionLog,
 							dataView,
 							logBuffer!,
--- a/src/transaction-log-reader.ts
+++ b/src/transaction-log-reader.ts
@@ -60,13 +60,6 @@ export class CorruptFrameError extends RangeError {
 }
 
 /**
- * How far past a break to look for a resume point. A partial append leaves one interrupted batch,
- * so a real resume point is close; bounding the search keeps the cost predictable and keeps the
- * number of candidate offsets — and so the odds of accepting a false one — small.
- */
-const MAX_RESYNC_SCAN_BYTES = 1 << 20;
-
-/**
  * Whether a complete, in-bounds frame begins at `pos`. The only sane bound on an entry's length is
  * the written extent: a single entry can legitimately exceed the rotation threshold (the first
  * entry written to a fresh file is always written in full). A zero timestamp is the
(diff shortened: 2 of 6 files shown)


kriszyp:
### Guided tour

**Read `src/transaction-log-reader.ts` first, and only the three new pieces**: `findResyncPosition` (~line 90), `corruptFrame` (~line 120), and the two throw sites in `query()` (~line 300). Everything else in the diff is `AGENTS.md`, an export, and tests.

The whole change is one idea: **the throw ends the entry, not the log.** `corruptFrame` computes where framing resumes and assigns `position` there *before* the error is thrown, so a caller that calls `next()` again continues past the break. If you only check one thing, check that `position` is always assigned before the `throw` on both paths — a throw that leaves it unmoved restores the exact bug this fixes.

**Where to look hardest: the `dataEnd` argument.** It must be the written extent (`getLogFileSize`, which returns the append-owned `TransactionLogFile::size`, not `st_size` and not the mapped length). Two ways to get this wrong, both of which I did in earlier rounds:

- Bound the scan by the *read limit*. For `readUncommitted` that is the pre-extended memory map, so the exact-end signal can never fire and only the eight-frame rule survives — which silently drops recovery for a break followed by 1–7 entries, on the boot-replay path, which is the path that lost 2.2 days in harper#2016. Caught in round 1.
- Fetch it per frame. It crosses into native and takes the store mutex. Caught in round 2; it now resolves inside `corruptFrame`, i.e. only when framing is already broken.

**The decision I'd most expect you to question** is the unbounded scan. Round 2 of review said to remove the 1 MiB cap (a partial frame wider than it puts everything behind it out of reach — the amputation this exists to prevent, at a threshold). Round 3 said removing it raises false-resync odds to ~N/2³² over the span and makes a large scan a one-off event-loop stall. I matched `validFramingResumes()`, on the grounds that its identical exposure is already shipped and a false resume yields one garbage frame that re-fails framing. I don't think this is clear-cut; if you'd rather have a cap, say so and name a number.

**What the tests prove:** that a mid-log break yields 11 of 12 entries instead of 1, that a sub-eight-frame run is recovered via the written extent, that the same holds under `readUncommitted` against a zero-padded map, and that a torn tail terminates when committed `size` over-reports the buffer.

**What they do not prove:** anything end-to-end. Every test substitutes a mock buffer for the memory map. Nothing reopens a genuinely damaged log on disk, and nothing exercises a real consumer — so whether harper#2016/#2063 actually close depends on harper#2087, not on this PR's green checks.

**Coverage:** codex ✓, gemini ✓, harper-domain ✓, three rounds each. Cursor legs pruned by policy. Rounds 1 and 2 each found a real defect (both listed above, both fixed). Round 3's findings were repeats of adjudicated items plus the cap trade-off. Unresolved and deliberately out of scope: the uncommitted detection bound (#749) and the write-side producer (#748).

src/binding/transaction/transaction.cpp
```diff
@@ -599,12 +599,10 @@ napi_value Transaction::Commit(napi_env env, napi_callback_info info) {
...
-		// The commit lanes are owned by the descriptor and drained/joined before
-		// the descriptor is destroyed; and an in-flight commit's state pins the
-		// descriptor alive (state -> txnHandle -> dbHandle -> descriptor). So a
-		// raw pointer captured into the worker task is valid for the task's
-		// lifetime and cannot form a reference cycle with the worker thread.
-		DBDescriptor* descriptor = dbHandle->descriptor.get();
+		// Keep the descriptor alive until every queued stage has completed. Worker
+		// teardown can close the transaction handle between the log and commit
+		// stages, releasing its DBHandle and its descriptor reference.
+		auto descriptor = dbHandle->descriptor;
```
cb1kenobi, line 607: **Low: the shared_ptr capture is a correct UAF fix, but its safety is non-local and now undocumented**

Copying the descriptor into the worker-task lambdas correctly closes the use-after-free the old raw pointer was exposed to (teardown can reset `dbHandle` between the log and commit stages, dropping the state's descriptor ref while a lane still dereferences it). It also reintroduces the descriptor -> worker -> task -> descriptor cycle the previous comment deliberately avoided. That is safe only because DBRegistry always keeps a map ref that outlives the tasks: the `use_count() <= 1` purge gate, plus drain+join under a live external ref before the map entry is erased. If a task's captured ref were ever the last one, `~DBDescriptor` would run on the worker thread and re-enter `commitWorker.shutdown()` (self-join under an already-held mutex) -> deadlock/terminate.

The old code documented why the raw pointer was cycle-free; the new comment explains why the ref is needed but no longer records why the reintroduced cycle is safe. Suggested fix: add one line here (or on `CommitWorker::shutdown`) noting the invariant, e.g. "must never be the last ref while a task is outstanding; guaranteed by DBRegistry drain-before-erase," so a future registry change can't silently make the self-join live. (Out of scope for the txnlog fix — flagging because it reverses the prior anti-cycle decision.)

—
_Generated by Barber AI_

test/transaction-log.test.ts
```diff
@@ -1512,6 +1531,267 @@ describe('Transaction Log', () => {
...
+					Object.defineProperty(log, '_getMemoryMapOfFile', realDescriptor);
+				}
+			}));
+	});
+
+	// The edges the synthetic mid-log case doesn't reach: a run shorter than RESYNC_MIN_FRAMES
+	// (recovered only by landing exactly on the written extent), and the readUncommitted bound —
+	// the boot-replay path, where the read limit is the pre-extended mmap and scanning against it
+	// would find no resume point at all.
+	describe('corrupt-frame resync bounds', () => {
```
cb1kenobi, line 1795: **Low: no test covers multiple corrupt regions in a single log**

The design is "one report per break, call `next()` again," and `drainPastCorruption` loops precisely so repeated breaks can be collected — yet every test here yields exactly one `CorruptFrameError`. The repeat path (a second break after resyncing past the first, each triggering a fresh `findResyncPosition` scan from the new position) is unexercised, so a regression that mishandles the second break wouldn't be caught.

Suggested fix: add a case that clobbers two separate entries' length fields (e.g. the 2nd and 7th of 12) and asserts two `CorruptFrameError`s with the expected `position`/`resyncPosition` for each, plus recovery of the surviving entries between and after them.

—
_Generated by Barber AI_

test/transaction-log.test.ts
```diff
@@ -1512,6 +1531,267 @@ describe('Transaction Log', () => {
 				(log as { _currentLogBuffer?: unknown })._currentLogBuffer = undefined;
 				expect(Array.from(log.query({ start: 0 })).length).toBe(3);
 			}));
+
+		// A mid-log break — a partial append the process survived, so valid entries were appended
+		// after it. Those stay reachable: the break costs one frame, not the rest of the log
+		// (HarperFast/harper#2016, #2063).
+		it('query() resyncs past a mid-log corrupt frame and still yields the entries after it', () =>
```
cb1kenobi, line 1530: **Low: no test plants a deceptive frame header inside the unreadable span**

The 8-frame / exact-written-extent rule is the load-bearing safety property of the resync (the PR itself flags it as the contested call), but no test constructs a plausible-looking false frame inside the corrupt gap. Every test's gap contains only the original clobbered entry, so the heuristic's ability to reject a lone false header — rather than resyncing onto it — is never actually verified.

Suggested fix: in the broken span write bytes that form one frame-shaped run (nonzero timestamp + in-bounds length) that neither chains `RESYNC_MIN_FRAMES` deep nor lands exactly on the written extent, then assert `resyncPosition` is still the true next-entry boundary and not that false header.

—
_Generated by Barber AI_

src/transaction-log-reader.ts
```diff
@@ -11,6 +11,132 @@ const UINT32_FROM_FLOAT = new Uint32Array(FLOAT_TO_UINT32.buffer);
...
+		let pos = start;
+		let frames = 0;
+		while (frameFits(dataView, pos, dataEnd)) {
+			pos += TRANSACTION_LOG_ENTRY_HEADER_SIZE + dataView.getUint32(pos + 8);
+			if (++frames >= RESYNC_MIN_FRAMES || pos === dataEnd) {
+				return start;
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
cb1kenobi: Thanks — the event-loop-stall concern behind this is real, but I'm going to leave `findResyncPosition` byte-by-byte, for two reasons:

1. **It's the JS mirror of `validFramingResumes()` in `transaction_log_recovery.cpp`, which is deliberately byte-by-byte (`++start`).** The two are kept in step on purpose (AGENTS.md invariant 10) so the reader and the open-time recovery scan classify a break identically. Skipping ahead in one but not the other reintroduces exactly the divergence that invariant exists to prevent.

2. **The skip can return a *later* resync than the earliest valid one.** Unlike the C++ side, which only needs a yes/no, the JS function must return the *earliest* resume point — every byte before it is an unreadable, committed entry. If a false run of 2–7 frames starting at `start` overruns the true resume point `R` and then fails, byte-by-byte finds `R`; jumping `start = pos` steps past it. That drops recoverable committed entries — the precise amputation this PR exists to fix — at roughly the same ~1/2^32 probability the heuristic already lives with. (The ~1/2^64 figure is for a false-positive resync, which is a different event from skipping past an earlier true one.)

The scan cost was weighed in an earlier review round and settled on matching `validFramingResumes()`: non-resyncing offsets fail after ~one frame check, so it's effectively linear in practice, and a genuinely huge corrupt span is a one-off event, not a hot path.

— Generated by Claude Opus 4.8

cb1kenobi (approved):

PR #751, opened 2026-08-05
fix(txnlog): erase a failed append's partial bytes instead of baking a framing break into the log

### LINKED ISSUE #748: A failed transaction-log append orphans its partial bytes, baking a mid-file framing break into the log

kriszyp:
Fixes #748.

A transaction-log append that fails part-way through — a full disk, an exhausted quota, a short
write on a failing volume — left the bytes it had already written on disk, unaccounted for. The log
fd is opened `O_APPEND`, so the next successful append landed *after* those bytes: a partial entry
permanently embedded mid-file with valid entries on both sides. That is the one shape open-time
recovery deliberately refuses to repair (truncating there would discard committed, replicated
transactions), so every reader stops at the break and everything written after it becomes
unreachable — the reader-side consequence seen in HarperFast/harper#2016 and HarperFast/harper#2063.

`writeBatchToFile()` now reports how many bytes reached the file through a `bytesLanded` out-param
instead of collapsing a hard error to `-1`, and `writeEntriesV1()` erases that range before
throwing. Erasing is safe: `writeEntriesV1` holds `fileMutex` across the whole append, `size` never
advanced over the bytes, and the commit throws — nothing acknowledged them. The erase reuses
`eraseTail()` from #723. Both platforms truncate back to `size` (the zero-fill this description
originally claimed for Windows was replaced by a physical truncate in #723) — with the caveat that
the Windows truncate refuses while another owner holds the memory map, so a failed append there
under a live JS reader retires the file instead of erasing it.

Three failure edges around that, each from a review round:

- **The erase itself fails** (ENOSPC then EIO). The file is retired rather than warned about: an
  `appendBoundaryLost` flag makes `writeEntriesV1` return without writing, which the store already
  reads as "no progress" and answers by rotating to the next sequence. The orphaned bytes stay the
  *trailing* partial that `recoverTail()` can truncate, instead of becoming the mid-file break it
  must leave intact. The double fault degrades to a repairable torn tail. Retirement is set *before*
  the erase and lifted only on success, so no way out of the erase short of success — a refusal, a
  failed truncate, a throw — can leave the file appendable.
- **The landed extent is unreportable.** Windows takes it from the file pointer rather than
  `lpNumberOfBytesWritten`, which `WriteFile` does not promise to set on failure; if that query also
  fails, the extent is reported as `TRANSACTION_LOG_BYTES_LANDED_UNKNOWN` and the file is retired
  rather than erased against a guess. A file pointer *behind* the write origin is nonsense for a
  write and is reported as unknown too, not as "nothing landed" — reporting `0` there would skip
  both the erase and the retirement and leave the file appendable over the orphan. A figure larger than the bytes handed to the OS is treated
  the same way — an over-large erase would cut into committed entries.
- **A short header write.** `open()` initialized a new file's header with three unchecked writes; a
  header that only partly landed still set `size` to the full header length, framing the first
  append from the wrong offset for the life of the file. It is now one checked write, and a short
  one removes the file — a size in `(0, HEADER_SIZE)` fails `open()`'s "too small" check on every
  future open, and freeing disk space would not heal it. (`removeFile()` splits into a
  `fileMutex`-acquiring wrapper and a per-platform `removeFileLocked()` body, since `open()` already
  holds the lock.)
- **A throw during cleanup.** `batch.currentEntryIndex` is rolled back before the erase and the
  warning emit, both of which allocate; a `bad_alloc` there must not leave the batch claiming
  entries that never reached disk.

## Stacked on #723

This branches from `kris/txnlog-committed-position-recovery` (#723), which introduces `eraseTail()`.
Review/merge that one first; the diff here is only this change.

## Where to look

- `writeEntriesV1()`'s error branch — the erase range (`committedSize` .. `committedSize +
  bytesLanded`) has to match where the bytes actually landed on *both* platforms: POSIX appends at
  physical EOF (which equals `size`), Windows seeks to `size` before writing and its file is
  pre-extended to `maxFileSize`, so `size` there is the logical end of entries, not the physical
  one. Both truncate back to `size`; a Windows orphan is overwritten by the next append, but a
  *shorter* next batch would leave its stale bytes reading as an entry, so the rule is the same.
- The retirement path and its contract with `TransactionLogStore::writeBatch` — an unchanged `size`
  is the existing "no progress" signal that triggers rotation, the same mechanism a
  max-size file uses. The flag is deliberately sticky for the object's life: it can only
  over-retire, never under-retire, and a fresh process rebuilds the object with it clear after
  recovery has had its chance at the trailing partial.
- `bytesLanded` accounting in both `writeBatchToFile()` implementations.

## Verification

Native GoogleTest, POSIX (`pnpm test:native`). Failure injection uses the existing macro-seam
pattern (`ROCKSDB_JS_WRITEV`, plus new `ROCKSDB_JS_WRITE` and `ROCKSDB_JS_FTRUNCATE`), defined only
in the `rocksdb-js-native-tests` gyp target — production builds resolve to the bare syscalls.

- `FailedAppendLeavesTheLogOnAnEntryBoundary` — **fails on base** with exactly the reported shape:
  file size 43 vs committed 37 (6 orphaned bytes), and after the next append the recovery scan
  returns `MidFileCorruption` with the entry count truncated at the break. Passes with the fix
  (scan `Clean`, both entries readable).
- `UnerasableOrphanRetiresTheFile` — ENOSPC + failing `ftruncate`: the retired file takes no further
  entries and the scan still classifies it `TruncateTail`.
- `ShortHeaderWriteDiscardsTheFileInsteadOfBrickingIt` — **fails on base**: the 5-byte file survives
  and the reopen throws "File is too small to be a valid transaction log file". Passes with the fix.
- `UnerasableExtent/Unknown` and `UnerasableExtent/OverReported` — both **fail on base** with the
  guard removed: the file is either appended past or erased against an untrustworthy range.
- `AppendThatWritesNothingLeavesTheFileUntouched`, plus two `WriteBatchToFile` tests for
  `bytesLanded` on a hard error and on a nothing-written error.

Full gates on macOS: `pnpm test:native` 120 passed (3 madvise tests skipped — Linux-only),
`pnpm test` 728 passed / 1 skipped, `pnpm check` clean.

## Open concerns from the pre-push review

Carried here rather than resolved, so a reviewer does not have to rediscover them:

- **Coverage stops at `TransactionLogFile`.** Nothing exercises the retire → rotate → retry contract
  through `TransactionLogStore::writeBatch`; the native-test target does not link the store (it
  needs RocksDB), and inducing a real `ENOSPC` from the JS suite would need a filled volume. The
  store side was verified by reading it (`transaction_log_store.cpp:831` rotates on unchanged size),
  not by a test.
- **Windows has no fault injection.** Its `bytesLanded` accounting and truncating erase are
  exercised only by #723's own `eraseTail` coverage (the arithmetic that turns a file pointer into a
  landed extent is now factored into `landedBytesFromFilePointer()` and unit-tested on every
  platform, but the `WriteFile`/`SetFilePointerEx` calls around it are not); there is no `WriteFile` seam, and I cannot build Windows
  locally — CI is the only verification for that backend. The *decision* logic those values feed
  (retire-vs-erase) is shared, and a test-only `forcedBytesLandedForTests` override drives it from
  the POSIX build.
- **Pre-existing, not introduced here:** `writeBatch`'s open-retry loop is unbounded, so a
  persistently failing volume spins through sequence numbers. A short header write is a new way to
  reach that loop, but `openFile()` already reached it under the same ENOSPC condition. Worth its
  own issue rather than widening this one.

Verified false and dismissed along the way: `removeFileLocked()` closes the file handle before
unlinking (no Windows sharing violation), and `bytesWritten` is zero-initialized before every
`WriteFile`, so it could never have contained garbage — the repeated "uninitialized stack memory"
framing of the Windows finding was wrong, even though the underlying under-reporting risk was real
and is now fixed.

## One unrelated commit

`test(block-cache): await the second read so teardown cannot abort it` fixes a pre-existing flake
that was blocking this PR's CI, not anything introduced here. `test/block-cache.test.ts` asserted
`db.get()` returns a Promise without awaiting it; when `dbRunner` closed the database first, the
in-flight read was aborted and the unhandled rejection failed the run with every test passing. It
took out the Deno and Node 24 macOS jobs on this branch while the same commit passed everywhere
else. Identical code is on `main`, and it is the only unawaited get-promise assertion in the suite.

Reviewed by codex + gemini + harper-domain across four rounds. Generated by Claude Opus 5.

---

**Review-feedback pass (2026-08-19).** Addressing the open threads on this PR:

- Windows `writeBatchToFile` reports a negative file-pointer delta as unknown, not `0` — the case
  that previously skipped both the erase and the retirement. The derivation is extracted into
  `landedBytesFromFilePointer()` with GoogleTest coverage of every branch on all platforms.
- `snapshotForBackup()` seeds the extent of a registered-but-never-opened log file before reading
  it. `size` is only meaningful after `open()`, so on the directory-iteration orders that left an
  older segment lazy, a backup silently omitted it. Done under `dataSetsMutex` (which `purge()`
  also holds) and the borrowed handle is closed again, so no fd or mapping is pinned.
- POSIX `truncateFile()` logs a truncation below the `MAP_SHARED` overlay boundary — the documented
  SIGBUS case — and lowers the overlay bookkeeping afterwards.
- Comment and AGENTS.md corrections: the Windows erase is a truncate, not an allocating zero-fill,
  and invariant 5's "`size` is the physical extent" clause is POSIX-only.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ fbf72fb7a892</sub>

<sub>Human-Review-Need: 4 @ fbf72fb7a892</sub>


src/binding/transaction_log/transaction_log_file_windows.cpp
```diff
@@ -414,13 +414,17 @@ int64_t TransactionLogFile::writeBatchToFile(iovec* iovecs, int iovcnt) {
 				std::string errorMessage = getWindowsErrorMessage(error);
 				DEBUG_LOG("%p TransactionLogFile::writeBatchToFile WriteFile failed (error=%lu: %s, iovec %d/%d)\n",
 					this, error, errorMessage.c_str(), i, iovcnt);
+				// Synchronous WriteFile (NULL lpOverlapped) always populates
+				// lpNumberOfBytesWritten, including on failure.
+				bytesLanded = totalBytesWritten + bytesWritten;
```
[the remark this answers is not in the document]
kriszyp: Adopted, with one adjustment. The "uninitialized stack memory" version of this concern does not apply — `bytesWritten` is `DWORD bytesWritten = 0` before every `WriteFile`, so it can never hold garbage and `eraseTail` could not have truncated a random range. The real exposure is the other direction: if Windows leaves it at 0 after a partial write, `bytesLanded` *under*-reports, and since the caller erases exactly that range, the bytes it missed stay on disk and reopen the framing break this PR exists to close.

So the file pointer is the better source either way — it measures rather than trusts. Taken in 884fdadf, using `SetFilePointerEx` instead of `SetFilePointer` (the `DWORD` form makes `INVALID_SET_FILE_POINTER` ambiguous with a legitimate 4GB offset), clamping a negative delta to 0, and keeping the accumulated count as the fallback when the query fails.

Note for human reviewers: this path has no local coverage — there is no `WriteFile` seam and I cannot build Windows here, so CI is the only verification.

src/binding/transaction_log/transaction_log_file_windows.cpp
```diff
@@ -412,13 +412,29 @@ int64_t TransactionLogFile::writeBatchToFile(iovec* iovecs, int iovcnt) {
...
+				// promise to set lpNumberOfBytesWritten on failure, and the caller
+				// erases exactly the range reported here — so if the pointer cannot be
+				// read either, report the extent as unknown rather than guess with a
+				// count that may be short.
+				LARGE_INTEGER zero, current;
+				zero.QuadPart = 0;
+				if (::SetFilePointerEx(this->fileHandle, zero, &current, FILE_CURRENT)) {
+					int64_t landed = current.QuadPart -
+						static_cast<int64_t>(this->size.load(std::memory_order_relaxed));
+					bytesLanded = landed > 0 ? landed : 0;
```
cb1kenobi: **Medium: a nonsensical file-pointer delta is reported as `0` (do nothing) rather than unknown**

The comment three lines up sets the policy — "report the extent as unknown rather than guess with a count that may be short" — but the clamp does the opposite for the one case it was written to catch. `landed == 0` legitimately means nothing landed, whereas `landed < 0` means the pointer is *behind* the offset we seeked to, i.e. the extent is exactly what this commit calls unknown. Reporting `0` makes the caller take the `if (extentUnknown || bytesLanded > 0)` branch as false: no erase, no retire, and the file stays appendable — reopening the mid-file framing break this PR exists to close.

Suggested fix:
```suggestion
					bytesLanded = landed >= 0 ? landed : TRANSACTION_LOG_BYTES_LANDED_UNKNOWN;
```

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -455,13 +480,59 @@ void TransactionLogFile::writeEntriesV1(TransactionLogEntryBatch& batch, const u
...
+		// to leave the break in place because valid entries follow it.
+		if (extentUnknown || bytesLanded > 0) {
+			uint32_t committedSize = this->size.load(std::memory_order_relaxed);
+
+			// Retire up front and lift it only once the erase has actually succeeded:
+			// the Windows erase allocates, so every way out of it short of success —
+			// including a throw — has to leave this file closed to further appends.
+			this->appendBoundaryLost.store(true, std::memory_order_relaxed);
+			if (!extentUnknown &&
+				this->eraseTail(committedSize, committedSize + static_cast<uint32_t>(bytesLanded))) {
```
cb1kenobi: **Medium: on Windows this erase is refused whenever a JS reader is tailing the log, and the rationale comment is stale after the rebase**

Two things drifted when the branch was rebased onto #723, which changed the Windows `eraseTail` from a zero-fill to a plain truncate:

1. **The erase now degrades to retire under a live reader.** Windows `eraseTail` (`transaction_log_file_windows.cpp:494`) delegates to `truncateFile`, which refuses outright when `memoryMap.use_count() > 1` (`transaction_log_file_windows.cpp:471`). The current file's map is handed to JS as an external Buffer whose native `shared_ptr` lives until GC (`transaction_log.cpp:231`, finalizer at `:238`), and `transaction-log-reader.ts:434` caches exactly that Buffer. So on Windows, any process tailing the live log keeps `use_count() > 1` and every failed append takes the `else` branch: `appendBoundaryLost` stays set and the file is retired instead of erased. That is still safe (the orphan remains a repairable trailing partial and the store rotates), but it means the PR's primary repair path is the *exception* on Windows in precisely the deployment shape #748 came from, while POSIX erases normally — the two backends reach different outcomes from the same on-disk state. Worth documenting, and worth deciding whether Windows should drop the map and retry rather than give up.

2. **"the Windows erase allocates" is no longer true.** Post-rebase `truncateFile` is `SetFilePointerEx` + `SetEndOfFile` + `FlushFileBuffers` — no buffer, no `std::string`, nothing that can throw. Retiring before the erase is still fine as belt-and-braces, but the stated reason for the ordering is now false, and the PR description's "Windows zero-fills the range to restore its end-of-entries marker" is out of date too. Both are load-bearing explanations in crash-durability code, so they're worth correcting before someone relies on them.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_file_windows.cpp
```diff
@@ -374,7 +372,9 @@ bool TransactionLogFile::removeFile() {
 	return true;
 }
 
-int64_t TransactionLogFile::writeBatchToFile(iovec* iovecs, int iovcnt) {
+int64_t TransactionLogFile::writeBatchToFile(iovec* iovecs, int iovcnt, int64_t& bytesLanded) {
```
cb1kenobi, line 498: **Medium: the Windows verification story isn't actually running — the Windows native-test target does not compile**

The PR notes that Windows has no fault injection and "CI is the only verification for that backend". On this head, that CI job fails at the *build* step, so none of the Windows code added here has been compiled by CI, let alone exercised:

```
test/native/transaction_log_erase_tail_test.cc(32,44): error C2589: 'constant': illegal token on right side of '::'
```

`TRANSACTION_LOG_FILE_TIMESTAMP_POSITION` / `TRANSACTION_LOG_FILE_HEADER_SIZE` are `#define`s (`transaction_log_file.h:50-51`), so the `rocksdb_js::`-qualified uses in that test expand to `rocksdb_js::5`. It only bites on Windows because of the transitive include order there.

This is **inherited from #723, not introduced here** — that file is not in this PR's diff, and #723's own CI fails with the identical error at the same base commit `ba0a59c7`. Flagging it because it invalidates the mitigation this PR relies on: dropping the `rocksdb_js::` qualifier on those macros in #723 would let the Windows job build and give this backend its first real coverage.

Same story for the red macOS/Windows test jobs: `transaction-log-crash-recovery.test.ts > seeds from the immediately preceding file when the active file is header-only` (`expected 1 to be 9`) reproduces deterministically on base `ba0a59c7` with none of this PR's commits applied, so it is also #723's to fix — nothing here regressed it.

—
_Generated by Barber AI (Claude Opus 5)_

AGENTS.md
```diff
@@ -213,7 +213,16 @@ sufficient (env teardown does not honor tsfn acquire counts); see
    append). Read/index paths (e.g. `findPositionByTimestamp`) must never truncate it — a zero
    timestamp seen mid-index during concurrent appends is a not-yet-visible memory-map artifact, not
    EOF. Reads during writes are bounded by the committed position, not `size` (see
-   `hasAppendedSinceOpen`; HarperFast/harper#1148).
+   `hasAppendedSinceOpen`; HarperFast/harper#1148). The other half of that contract is that
+   `size` is also the _physical_ extent: an append that fails part-way (ENOSPC, a short write
```
cb1kenobi: **Low: the new invariant is POSIX-only, stated unconditionally**

`size` is *not* the physical extent on Windows: `getMemoryMapLocked` deliberately pre-extends the file to `maxFileSize` with `SetEndOfFile` before mapping (`transaction_log_file_windows.cpp:256-297`), so an active Windows segment's on-disk size is `maxFileSize` with a zero-padded tail for its whole life. The supporting clause — "with the file opened `O_APPEND`, writes go to physical EOF, not to `size`" — is POSIX-only as well; Windows `writeBatchToFile` explicitly seeks to `size` first (`:383`).

AGENTS.md is where agents go for invariants they're expected to enforce, so an unqualified version of this one will produce wrong Windows code. Suggest scoping it: `size` is the logical end of entries on both platforms and the physical extent on POSIX; on Windows the physical file is pre-extended and end-of-entries is found via the zero-timestamp convention instead.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_file_posix.cpp
```diff
@@ -400,14 +420,14 @@ int64_t TransactionLogFile::writeToFile(const void* buffer, uint32_t size, int64
...
-	return static_cast<int64_t>(::write(this->fd, buffer, size));
+	return static_cast<int64_t>(ROCKSDB_JS_WRITE(this->fd, buffer, size));
 }
 
 bool TransactionLogFile::truncateFile(uint32_t newSize) {
 	if (this->fd < 0) {
 		return false;
 	}
-	if (::ftruncate(this->fd, static_cast<off_t>(newSize)) != 0) {
+	if (ROCKSDB_JS_FTRUNCATE(this->fd, static_cast<off_t>(newSize)) != 0) {
```
cb1kenobi, line 588: **Low: POSIX `truncateFile` has no outstanding-map guard, unlike its Windows counterpart**

Windows refuses to truncate while a map is outstanding (`transaction_log_file_windows.cpp:471`); this path just calls `ftruncate` and leaves `this->memoryMap` and any JS-held copies pointing at the old mapping. The new `eraseTail` call site is safe today, but only by caller discipline: `updateMemoryMapOverlay()` runs solely on the write-success path (`transaction_log_file.cpp:539-541`), so the `MAP_SHARED` overlay never covers the failed append's bytes, and the erase always truncates back to a `size` that was already published. Nothing structural enforces that — a future caller passing a `newSize` below the current overlay boundary would leave readers faulting `MAP_SHARED` pages past EOF, which is the documented SIGBUS case.

A cheap `assert`/`DEBUG_LOG` that `newSize >= lastOverlaySize` would pin the invariant where it is actually relied upon.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
```diff
@@ -311,6 +311,11 @@ LogPosition TransactionLogStore::findPositionByTimestamp(double timestamp) {
 	}
 	while (it != this->sequenceFiles.end()) {
 		auto logFile = it->second.get();
+		// Directory iteration order is unspecified, so registerLogFile() may not
+		// have opened an older file before a higher sequence became current.
+		if (!logFile->isOpen()) {
+			logFile->open(this->latestTimestamp);
+		}
```
cb1kenobi, line 321: **Medium: the same registered-but-never-opened state still silently drops files from `snapshotForBackup()`**

The diagnosis here is right and the fix is real — I traced it: `TransactionLogFile`'s constructor (`transaction_log_file_posix.cpp:59` / `_windows.cpp:16`) seeds nothing but `path` and `sequenceNumber`, so `size` stays at its declared `0` (`transaction_log_file.h:108`) until `openFile()` runs, and `close()` never resets it. Feeding that `0` into `TransactionLogFile::findPositionByTimestamp` as `mapSize` makes `getMemoryMapLocked()` hand back `nullptr`, which returns `0xFFFFFFFF` (`transaction_log_file.cpp:588-591`); the store then `break`s and returns `{ HEADER_SIZE, sequenceNumber + 1 }`, positioning the reader at the *start of the next file* and skipping every entry in the unopened one. Directory-order dependent, so it would show up on ext4 and hide on APFS.

The concern is that this patches one consumer of that state rather than the state itself, and there is at least one more consumer with a worse failure mode. `snapshotForBackup()` (`transaction_log_store.cpp:426-431`) reads the same field:

```cpp
uint64_t byteLimit = file->size.load(std::memory_order_relaxed);
if (byteLimit == 0) {
	// A just-created current file whose header has not landed yet — nothing
	// to back up, and a header-less file would confuse recovery on restore.
	continue;
}
```

That comment enumerates only one way to reach `size == 0`; a registered-but-never-opened older file is the other, and it takes the `continue` too — so on the same iteration orders this commit fixes for reads, a **backup silently omits whole rotated log segments** with no error and no warning. `collectStats()` (`:504`) has the benign version of it, under-reporting `totalSizeBytes` for the same files.

Suggested fix: enforce the invariant where the state is created instead of at each reader. `registerLogFile()` already does a `std::filesystem::last_write_time(path)` stat for `fileLastWriteTime` (`:708-715`) — seeding `logFile->size` from `std::filesystem::file_size(path)` in that same `try` makes "registered" imply "has a usable extent" for every consumer at once, and leaves `size == 0` meaning only what the backup comment already assumes it means. The lazy `open()` here is still needed for the *mapping*, so it can stay; it just stops being load-bearing for correctness at each new call site someone adds.

One smaller note on this hunk: `open()` throws on a bad token/version/undersized file, and unlike `load()`'s scan loop (`:1116-1140`, which wraps `open()` in try/catch and closes what it opened) this call site lets it propagate, so one corrupt older `.txnlog` now makes every timestamp-based reader positioning throw instead of skipping past it. Arguably the better behavior — worth being deliberate about it rather than inheriting it.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_file_windows.cpp
cb1kenobi: Confirmed fixed at \`de43892c\` — the \`rocksdb_js::\` qualifiers are gone from the macro uses, and the Windows native C++ job on this head now builds and runs: [105 tests from 14 test suites, all passed](https://github.com/HarperFast/rocksdb-js/actions/runs/32212654104/job/95951426849), including all three \`TransactionLogEraseTail\` cases. That backend has real coverage for the first time. Resolving.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
kriszyp: Fixed for `snapshotForBackup()` in 06ef658 — but not by seeding at `registerLogFile()`, because `file_size(path)` is not the written extent on Windows: `getMemoryMapLocked()` pre-extends an active segment to `maxFileSize` with `SetEndOfFile`, so the on-disk size is padding plus entries and only the open-time index scan can tell them apart. Seeding from a stat would have made "registered implies a usable extent" true on POSIX and wrong on Windows.

Instead the extent is resolved where it is consumed, but with the handle *borrowed*: inside the same `dataSetsMutex` block `purge()` serializes against (so a concurrent unlink cannot slip between the `isOpen()` check and `open()`, whose `O_CREAT`/`OPEN_ALWAYS` would resurrect a header-only ghost segment), and closed again if it was closed before, so no fd or Windows mapping is pinned. `close()` leaves `size` set, so the walk costs its syscalls once per process rather than once per backup.

On the smaller note: yes, `open()` propagating is deliberate — an unopenable segment hard-fails the backup rather than silently omitting itself, since a gap in a backed-up log is invisible until a restore needs it. `collectStats()` is left as-is (it is deliberately syscall-free) with a comment naming the under-report.

— KrAIs (Claude Opus 5)

src/binding/transaction_log/transaction_log_file.cpp
kriszyp: Both corrected in 48c650a.

1. Documented rather than changed. Dropping the map to force the truncate is not available — JS owns that mapping — so under a live Windows reader the failed append retires the file instead of erasing it, and the store rotates. The comment now says so, and the retire-up-front rationale is stated as "every way out short of success" rather than the (now false) allocation claim.

2. The neighbouring `O_APPEND` comment was POSIX-only too. It now notes that a Windows append seeks back to `size`, so the orphan is overwritten — but a *shorter* next batch would leave the orphan's stale bytes past its own end, reading as an entry rather than the end-of-entries marker, which is why erase-or-retire is the rule on both platforms. PR description updated.

— KrAIs (Claude Opus 5)

src/binding/transaction_log/transaction_log_file_windows.cpp
kriszyp: Fixed in 48c650a — `landed >= 0 ? landed : TRANSACTION_LOG_BYTES_LANDED_UNKNOWN`, as suggested. The derivation is now `landedBytesFromFilePointer()` in `transaction_log_file.h`, compiled unconditionally so GoogleTest covers the zero, positive and negative-delta branches on every platform rather than only on a Windows runner.

— KrAIs (Claude Opus 5)

AGENTS.md
kriszyp: Scoped in 48c650a. Invariant 5 now says the physical extent tracks `size` on POSIX only, and spells out the Windows shape: pre-extended to `maxFileSize` with `SetEndOfFile` so it can be mapped, physical size stays there for the segment's life, end-of-entries found via the zero-timestamp convention, appends seek to `size` first. The erase-or-retire rule is stated as common to both.

— KrAIs (Claude Opus 5)

src/binding/transaction_log/transaction_log_file_posix.cpp
kriszyp: Added in 48c650a: `truncateFile()` DEBUG_LOGs a `newSize` below `lastOverlaySize` (naming the SIGBUS consequence), and lowers `lastOverlaySize` after a successful truncate so a later append re-overlays the reused range instead of assuming it is covered. Kept as a log rather than a refusal — `recoverTail()` legitimately truncates below an overlay established by the open-time index scan, before any mapping reaches a reader.

— KrAIs (Claude Opus 5)

src/binding/transaction_log/transaction_log_store.cpp
```diff
@@ -408,6 +413,32 @@ std::vector<TransactionLogBackupEntry> TransactionLogStore::snapshotForBackup()
...
+			// so a backup of a store with many rotated segments does not accumulate
+			// fds (or, on Windows, mappings) for the rest of the process's life.
+			// close() leaves `size` set, so this walk costs its syscalls once per
+			// process, not once per backup.
+			//
+			// Only a definite absence skips the open: a stat that *errors* leaves us
+			// unable to tell, and omitting a segment from a backup is the failure this
+			// is here to prevent, while the worst case of opening a since-deleted path
+			// is a 13-byte header stub that the next startup rescan purges.
+			if (file->size.load(std::memory_order_relaxed) == 0 && !file->isOpen()) {
```
cb1kenobi: **Medium: `doPurge()` is the next consumer of the same `size == 0` state, and its failure mode is deleting entries that never reached RocksDB**

The rebuttal on seeding at `registerLogFile()` holds up — I traced it. On Windows `openFile()` seeds `size` from `GetFileSizeEx` and then runs the index scan (`transaction_log_file_windows.cpp:197`), which corrects `size` down to the real end-of-entries via `hasAppendedSinceOpen` (`transaction_log_file.cpp:642-643`). A bare `file_size()` stat has no equivalent, so it really would have been right on POSIX and wrong on Windows. Borrowing the handle is the correct way to get a trustworthy extent.

That settles *how* to resolve the extent, though, not *where*. Resolving it per consumer leaves the next one uncovered, and there is one — `doPurge()`'s flushed-position guard (`:659-664`):

```cpp
if (sequenceNumber > lastFlushedPosition.logSequenceNumber ||
	(sequenceNumber == lastFlushedPosition.logSequenceNumber &&
		logFile->size > lastFlushedPosition.positionInLogFile)
) {
	continue;
}
```

The equality branch is the "this file still has unflushed entries, do not delete it" guard. A registered-but-never-opened file reads `size == 0`, so `0 > positionInLogFile` is false, the `continue` is skipped, and `removeFile()` runs.

Reachability is the same shape this hunk fixes. `registerLogFile()` only opens a file when `sequenceNumber >= currentSequenceNumber` (`:748-751`), and `load()`'s boundary scan opens at most the current file plus one predecessor and `break`s on the current file as soon as it has a complete transaction (`:1140-1160`) — so on a descending directory iteration every older segment stays lazy with `size == 0`. If `getLastFlushedPosition()` names one of those (RocksDB flush behind rotation) and it is past `retentionMs` or `before`, the JS-facing purge (`transaction_log_store_registry.cpp:340`) deletes a segment whose tail was never flushed. Silent — no error, no warning.

`collectStats()` mirrors the same guard at `:583-584` off the same field, so the gauge would have reported that file as `purgeableFiles` rather than `retainedUnflushedFiles` — the symptom that shows up just before the loss. And `:672`'s `removedSize` is the benign version again: `bytesPurged` under-reports for those files.

Narrow — it needs the flush frontier to sit in a retention-expired segment — but the outcome is data loss rather than a skipped backup, which is why it seems worth closing rather than documenting. Found by reading; I have no test that reaches it.

Suggested fix: lift these five lines into a small private `ensureExtent(file)` (precondition: `dataSetsMutex` held) and call it from `doPurge()`'s loop as well, so the next consumer inherits it instead of rediscovering it.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
```diff
@@ -408,6 +413,32 @@ std::vector<TransactionLogBackupEntry> TransactionLogStore::snapshotForBackup()
...
+				std::error_code existsEc;
+				if (std::filesystem::exists(file->path, existsEc) || existsEc) {
+					file->open(this->latestTimestamp);
+					file->close();
```
cb1kenobi: **Low: the borrowed handle is pinned, not released, when `open()` throws**

`open()` can throw after `openFile()` has already succeeded — bad token, unsupported version, "file is too small", a failed header read (`transaction_log_file.cpp:124-160`). On those paths `file->close()` is skipped, so the handle stays open for the life of the process, and on Windows the mapping `openFile()`'s index scan created (`transaction_log_file_windows.cpp:197` → `getMemoryMapLocked`) is pinned with it. That is exactly the "no fd or mapping is pinned" property the commit is named for, lost on the one path where the file is already known to be bad.

Letting the throw propagate is the right call (agreed with the reasoning on the earlier thread) — it should just propagate with the handle released.

Suggested fix:
```suggestion
					try {
						file->open(this->latestTimestamp);
					} catch (...) {
						file->close();
						throw;
					}
					file->close();
```

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_file.h
```diff
@@ -40,6 +40,34 @@
...
+ * pointer ended up relative to the offset the batch started writing at.
+ *
+ * A pointer BEHIND the origin is not "nothing landed" — it is nonsense for a
+ * write, and reporting 0 would tell the caller there is nothing to erase and
+ * nothing to retire, leaving the file appendable over a partial entry. Report
+ * it as unknown so the caller retires the file instead.
+ *
+ * Used by the Windows append path (`WriteFile` does not promise to set
+ * `lpNumberOfBytesWritten` on failure); defined here, unconditionally, so the
+ * branch is covered by the GoogleTest suite on every platform.
```
cb1kenobi: **Low: the new tests do not run on the one platform that calls this**

The function is compiled unconditionally, but its tests are not: `test/native/transaction_log_writev_test.cc` wraps the whole translation unit in `#ifndef _WIN32` (`:11`, closed at `:537`), so both `LandedBytesFromFilePointer` cases are compiled out on Windows. The Windows native C++ job on this head reports `105 tests from 14 test suites` — byte-identical to the count on `de43892c`, confirming the two new tests added nothing there.

The coverage goal is still met — the derivation is platform-independent and macOS/Linux exercise the zero, positive and negative branches — so this is the comment being wrong rather than a gap. Worth fixing because it is the kind of claim a later reader takes at face value. Moving the two `TEST()`s to a file that compiles everywhere (`transaction_log_validation_test.cc`) would make it true instead.

Suggested fix:
```suggestion
 * branch is covered by the POSIX GoogleTest suite (transaction_log_writev_test.cc
 * is `#ifndef _WIN32`) rather than only on a Windows runner.
```

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
```diff
@@ -408,6 +443,26 @@ std::vector<TransactionLogBackupEntry> TransactionLogStore::snapshotForBackup()
...
 		for (const auto& [seq, file] : this->sequenceFiles) {
+			// Resolve the written extent of a lazily-registered segment, or the
+			// `byteLimit == 0` skip below silently omits it from the backup
+			// (directory iteration order decides which segments are lazy, so ext4
+			// hits those orders and APFS hides them). close() leaves `size` set, so
+			// this walk costs its syscalls once per process, not once per backup.
+			try {
+				this->ensureExtent(file);
+			} catch (const std::exception& e) {
+				// A segment whose header will not open has no readable entries, so
```
cb1kenobi: **Medium: this catch cannot tell "the header is corrupt" from "I could not open it right now", and silently drops a healthy segment from the backup**

The comment states the premise — *a segment whose header will not open has no readable entries* — but `ensureExtent()` -> `open()` throws a bare `DBException` for two very different classes of failure, and `DBException` is `final` with no code or type to separate them (`core/exception.h:23`):

**Genuinely unreadable** (the premise holds): bad token, unsupported version, "File is too small" — `transaction_log_file.cpp:137,142,163`.

**Transient / environmental** (the premise does *not* hold — the segment may be intact and full of committed transactions):
- `"Failed to open sequence file for read/write"` — `::open` failed: **EMFILE/ENFILE**, EACCES, EIO (`transaction_log_file_posix.cpp:92`)
- `"Failed to get file size"` — `fstat` failed (`:99`)
- `"Failed to read version from file"` / `"Failed to read file timestamp from file"` — `readFromFile()` returned `-1`, i.e. a `pread` EIO on a bad sector (`transaction_log_file.cpp:147,157,168`)
- `"Failed to create parent directory"` (`transaction_log_file_posix.cpp:81`)

fd exhaustion is the one to worry about: RocksDB holds a lot of fds, and this loop opens *every* lazily-registered segment in one pass. One `EMFILE` and the backup drops that segment, returns success, and the gap is invisible until a restore needs it — which is the exact failure mode argued against on the earlier thread ("an unopenable segment hard-fails the backup rather than silently omitting itself"). `log.warn` is out-of-band; nothing in the return value tells the caller the backup is incomplete.

Note `doPurge()` gets this right for free — an unresolvable extent there means "refuse to purge", so both classes fail safe. It is only the backup where the two need separating.

Suggested fix: make the distinction at the throw site rather than at the catch, since that is the only place that still knows which it was — e.g. a `bool malformed` on `DBException` (or a small `TransactionLogFormatException`) set only by the token/version/too-small throws. Then `snapshotForBackup()` keeps the `continue` for a malformed header and rethrows everything else, preserving the previous hard-fail for the case where the segment might be fine.

Secondary: this catch is `catch (const std::exception&)` only, while the matching one in `doPurge()` (`:697`) also has a `catch (...)`. Worth matching them.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log.cpp
```diff
@@ -257,7 +257,15 @@ napi_value TransactionLog::FindPosition(napi_env env, napi_callback_info info) {
 	UNWRAP_TRANSACTION_LOG_HANDLE("FindPosition");
 	double timestamp = 0;
 	NAPI_STATUS_THROWS(::napi_get_value_double(env, argv[0], &timestamp));
-	LogPosition position = (*txnLogHandle)->findPosition(timestamp);
+	// findPosition() opens lazily-registered segments, and open() throws on a
+	// bad header — a C++ exception escaping an N-API callback aborts the process.
```
cb1kenobi, line 273: **Medium: the same unguarded `open()` throw escapes two neighbouring N-API callbacks in this file**

The reasoning in this comment is right, and it applies verbatim two functions up. Both of these call store methods that open lazily-registered segments with no `try`/`catch` anywhere on the path:

- `GetLogFileSize` (`:174`) -> `TransactionLogStore::getLogFileSize()` (`transaction_log_store.cpp:251` and `:260`) — opens the named file, or **every** file when no sequence number is given.
- `GetMemoryMapOfFile` (`:221`) -> `TransactionLogStore::getMemoryMap()` (`transaction_log_store.cpp:228`).

Same `open()`, same `DBException` on a bad token/version/too-small/short-read, same escape out of a `napi_callback`. `GetLogFileSize` is the more exposed of the two: `transaction-log-reader.ts` calls it from six sites including the tailing loop (`:250`, `:272`, `:283`, `:409`, `:471`), and `:471` asks for `nextLogId` — a segment that may well still be lazy.

Pre-existing rather than introduced here, but `1229e855` is specifically "keep a bad log segment from aborting FindPosition", and one corrupt `.txnlog` still aborts the process through either sibling. Wrapping both the same way would finish the commit's own goal.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
cb1kenobi: Confirmed fixed at `9873eb99`, and this one is now covered by a test that actually reaches it.

I built `80a1c9b4` in a second worktree with only `test/transaction-log.test.ts` taken from `ea72583f`, and `should not purge a log file whose tail is past the flushed position` **fails on base** — `expected [ Array(1) ] to deeply equal []`, i.e. base really does delete segment 1 — and passes on `ea72583f`.

Worth noting the test is stronger than its own comment claims. It says whether a segment is left lazy "depends on directory iteration order, which a test cannot pin down", but with three segments it is deterministic and does not depend on iteration order at all: the boundary scan in `load()` walks `sequenceFiles` in reverse, resolves the extent of the current file (3) and exactly one predecessor (2, whose `close()` leaves `size` set), then breaks on `scannedOlderFile` (`transaction_log_store.cpp:1186-1188`). Segment 1 is never opened, so it reads `size == 0` on every platform — and the fixture points the flushed position straight at it. That is why it fails on base every run rather than only on ext4.

On the other half of the question: the guard does not now over-retain. An unresolvable extent only blocks the one segment equal to the flushed sequence number; once the flush frontier advances past it, `sequenceNumber < lastFlushedPosition.logSequenceNumber` skips the extent check entirely and it purges normally.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
cb1kenobi: Confirmed fixed at `23baf6eb`, and fixed one level better than suggested — putting the release inside `open()` itself covers every speculative caller instead of just this one, and `closeLocked()` correctly avoids re-entering `fileMutex` that `open()` already holds. `closeLocked()` is also safe on the paths where `openFile()` itself threw: both platform bodies are guarded (`if (this->memoryMap)`, `if (this->fd >= 0)`), so a never-opened handle is a no-op.

Separately, `ea72583f` is the more interesting commit in this round, and it says my earlier clearance of the borrow was incomplete — so, for the record on why:

I verified the borrow does not **leak** a handle. I did not verify it cannot **steal** one. Those are different properties, and only the first is a lifetime question. `purge()` takes `writeMutex` *and* `dataSetsMutex` (`transaction_log_store.cpp:628-630`), so `ensureExtent()` there is fully serialized against the write path. `snapshotForBackup()` takes only `dataSetsMutex` — and `writeBatch()` never takes it while opening or appending to the active segment; it holds `writeMutex` and calls `logFile->isOpen()` / `logFile->open()` at `:840-844`. Nothing orders the two. So the interleaving is real: the backup thread reads `!isOpen()`, `open()`s, the writer then opens (a no-op, fd already set) and appends under `fileMutex`, and the backup thread's `close()` then pulls the fd out from under a live append — plus, on Windows, the mapping, and `lastOverlaySize` reset to 0. Excluding the active segment is the right cut because it is the only file the write path touches outside `dataSetsMutex`.

I also checked the exclusion is complete and does not reintroduce the original problem:

- The `currentSequenceNumber` load is relaxed, but it is not racy here: `rotateToNextSequence()` bumps it under `dataSetsMutex` (`:209`), the same lock `ensureExtent()` runs under, so it cannot observe a stale value that would let it borrow the *new* active segment.
- `size == 0 && !isOpen()` for the active segment really is the truth, so skipping it costs nothing. `registerLogFile()` opens eagerly for `seq >= currentSequenceNumber` (`:793-796`), and the only other way a file becomes current is `rotateToNextSequence()` -> `getLogFile()` (`:1080-1091`), which creates a `TransactionLogFile` for a sequence with no file on disk yet. `close()` never resets `size`, so a previously-opened segment can never fall back into this branch.
- `doPurge()` is unaffected by the exclusion for the same reason — reaching `removeFile()` on the active segment would need it to be both retention-expired and never appended to.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_file.h
cb1kenobi: Confirmed fixed at `1229e855` — the two `TEST()`s **moved**, they were not dropped. They are now in `test/native/transaction_log_validation_test.cc:509-517`, which has no `_WIN32` guard, and the `-18` on `transaction_log_writev_test.cc` is exactly the block that left.

Verified from the CI job log rather than the diff, since the count is the only thing that can distinguish "moved" from "deleted": the Windows native C++ job on `ea72583f` reports [107 tests from 15 test suites, all passed](https://github.com/HarperFast/rocksdb-js/actions/runs/32231063563/job/96000733330), up from 105 / 14 on `de43892c`. Exactly +2 tests and +1 suite — the new `LandedBytesFromFilePointer` suite compiling on Windows for the first time. macOS is unchanged at 127 tests / 19 suites (124 passed, 3 Linux-only madvise skips), which is the expected null result there since both translation units already compiled.

The header comment now says "so it is testable off Windows", which is true but understates it — the tests run on Windows now too.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
kriszyp: Addressed in 6ba78715 and 49cc9588. Format-invalid headers are now classified separately and skipped with a warning, while transient/environmental open failures propagate through backup collection as a failed `rocksdb::Status`. A rejected speculative open also clears its unvalidated cached extent, so preflight and copy cannot disagree on a malformed segment. — KrAIs (GPT-5 Codex)

src/binding/transaction_log/transaction_log.cpp
kriszyp: Addressed in 6ba78715. `GetLogFileSize` and `GetMemoryMapOfFile` now catch standard C++ exceptions at the N-API boundary and convert them to JavaScript errors, matching `FindPosition`; the regression test exercises both lazy-open failure paths. — KrAIs (GPT-5 Codex)

src/binding/transaction_log/transaction_log_store.cpp
```diff
@@ -408,6 +443,25 @@ std::vector<TransactionLogBackupEntry> TransactionLogStore::snapshotForBackup()
...
 		files.reserve(this->sequenceFiles.size());
 		for (const auto& [seq, file] : this->sequenceFiles) {
+			// Resolve the written extent of a lazily-registered segment, or the
+			// `byteLimit == 0` skip below silently omits it from the backup
+			// (directory iteration order decides which segments are lazy, so ext4
+			// hits those orders and APFS hides them). close() leaves `size` set, so
+			// this walk costs its syscalls once per process, not once per backup.
+			try {
+				this->ensureExtent(file);
+			} catch (const TransactionLogFormatException& e) {
```
cb1kenobi, line 455: **Low: the comment five lines up is now stale, and the malformed case re-warns on every backup**

The narrowing itself is right — I confirmed it by mutation: with this catch widened back to `catch (const std::exception&)` and everything else on this head unchanged, both new `rejects a transaction log snapshot when a segment cannot be opened` cases fail (`promise resolved "1"` / `"undefined"` instead of rejecting), i.e. the backup silently succeeds short. The narrow catch is load-bearing.

What drifted is the cost note at `:449-450` — *"close() leaves `size` set, so this walk costs its syscalls once per process, not once per backup"*. That is true for a segment that opens, but `49cc9588` now zeroes `size` in `TransactionLogFile::open()`'s catch, so a malformed segment stays `size == 0` forever and `ensureExtent()` re-opens it on every pass. Measured on this head with a store holding one malformed lazy segment, subscribing to `log.warn`:

```
backup #1 warnings: 2
backup #2 warnings: 2
```

Two per backup because `checkDiskSpace` makes `runCreateBackup` call `collectTransactionLogBackupEntries()` for the preflight estimate and again through `backupTransactionLogsToDir` — and it repeats for the life of the database, since nothing ever resolves the extent. A legacy bad segment therefore emits two identical warnings per backup forever.

Suggested fix: qualify the comment (the retry-per-backup cost applies to the failure path), and consider a sticky per-file "already reported malformed" flag so the warning is emitted once rather than on every snapshot.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/database/backup_transaction_logs.h
```diff
@@ -25,10 +25,13 @@ struct NamedTransactionLogBackupEntry final {
 /**
  * Collects the backup entries for every transaction log store owned by the
  * database at `descriptor` (via the process-global store registry). Each
- * store's files are snapshotted under that store's `dataSetsMutex`. Intended to
- * run on a backup worker thread while the database stays pinned.
+ * store's files are snapshotted under that store's `dataSetsMutex`. Returns an
+ * I/O status instead of letting a native exception escape the backup worker.
  */
-std::vector<NamedTransactionLogBackupEntry> collectTransactionLogBackupEntries(DBDescriptor* descriptor);
+rocksdb::Status collectTransactionLogBackupEntries(
```
cb1kenobi, line 33: **Low: the malformed-vs-environmental contract is now undocumented anywhere**

`c1fa8f78` ("Keep agent guidance out of the review diff") removed exactly the paragraph that wrote this rule down — AGENTS.md invariant 8 no longer says that a malformed header is excluded with a warning while an environmental open/read failure must reject the backup, nor that `collectTransactionLogBackupEntries()` converts those failures to a `rocksdb::Status` so nothing escapes the worker. I checked the commit: it touches only AGENTS.md and drops nothing else, and the text was not moved elsewhere.

The result is that the only record of a subtle data-integrity rule is the four-line comment at `transaction_log_store.cpp:454`. Widening that catch back to `std::exception` compiles, passes a casual reading ("a segment that will not open has nothing to back up"), and silently reintroduces short backups — which is exactly what the mutation above demonstrates. The rest of this PR adds 39 lines to AGENTS.md, so diff hygiene does not really argue for dropping this one.

Suggested fix: restore the invariant-8 sentences, or fold the same two sentences into this doc comment — it already explains the `rocksdb::Status` half but not *why* only malformed headers may be skipped.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log_store.cpp
cb1kenobi: Confirmed fixed at `6ba78715`, and fixed the way that actually closes it — **at the throw site, not the catch site**. The three format throws in `openLocked()` now raise a new `TransactionLogFormatException` (`transaction_log_file.h:53`): "File is too small" (`:140`), bad token (`:153`), unsupported version (`:166`). Everything environmental stays `DBException` — `::open` (EMFILE/ENFILE/EACCES/EIO), `fstat`, all three `readFromFile() < 0` short-read paths, and "Failed to create parent directory". No message matching, no widened catch, no retry, so an `EMFILE` can no longer be mistaken for a corrupt header.

`snapshotForBackup()` catches only the format type; everything else propagates, `collectTransactionLogBackupEntries()` turns it into `rocksdb::Status::IOError("Failed to snapshot transaction log metadata", ...)`, and all three call sites bail on it — `runCreateBackup` (preflight, before the directory lock is taken, so no lock leak), `backupTransactionLogsToDir` (which rolls the RocksDB backup back via `DeleteBackup`), and `doBackupStream`. A segment that cannot be opened now **fails the backup** instead of producing a silently short one.

Verified rather than read:

- **Mutation.** I reverted just this line to `catch (const std::exception& e)`, rebuilt the one TU, and both new `rejects a transaction log snapshot when a segment cannot be opened` cases fail — `promise resolved "1"` (directory) and `"undefined"` (stream). So the tests really do pin the fail-closed behaviour, and base really does swallow it.
- **Native.** `pnpm test:native` on macOS: 128 tests / 20 suites, 125 passed, 3 Linux-only madvise skips — up from 127/19/124 last round. The extra one is `TransactionLogFileOpen.ClassifiesMalformedHeadersSeparatelyFromIoFailures`, which asserts the type split directly (`EXPECT_THROW(..., TransactionLogFormatException)` for the three malformed shapes, `DBException` for an EISDIR open).
- **JS.** `test/backup-transaction-logs.test.ts` + `test/transaction-log.test.ts`: 88 passed / 3 skipped, `tsc --noEmit` and `oxlint` clean.

I also checked that the new exception type cannot be missed elsewhere: there is no `catch (const DBException&)` anywhere in `src/binding`, so nothing that used to catch these throws stops catching them.

On the secondary note — the `catch (...)` suggestion is moot now that the policy is "propagate everything that is not a format error"; a non-`std::exception` throw correctly falls through to `collectTransactionLogBackupEntries()`'s `catch (...)`. Resolving. Nice fix.

—
_Generated by Barber AI (Claude Opus 5)_

src/binding/transaction_log/transaction_log.cpp
cb1kenobi: Confirmed fixed at `6ba78715` — both siblings now carry the same guard `FindPosition` got, in the same shape: a `try` around the store call, `::napi_throw_error(env, nullptr, e.what())`, `return nullptr`. `GetLogFileSize` (`transaction_log.cpp:174`) and `GetMemoryMapOfFile` (`:227`) can no longer let a `DBException` cross the `napi_callback` boundary.

I walked the rest of the file to make sure nothing was left: `AddEntry` was already guarded, `GetLastCommittedPosition`/`GetLastFlushed`/`GetName`/`GetPath`/`GetStats` never reach `open()`, and `Constructor` -> `resolveTransactionLogStore()` -> `TransactionLogStore::load()` swallows a throwing `registerLogFile()` inside the directory-scan loop's own `catch (const std::exception&)` (`transaction_log_store.cpp:1139`), so the discovery path was never an abort route. That leaves no unguarded `open()` escape into N-API that I can find.

Covered by a test that actually reaches it, too: `reports lazy-segment open failures as JavaScript errors` (`test/transaction-log.test.ts:707`) builds a genuinely lazy segment, corrupts its header after `db.open()`, and asserts both `getLogFileSize(seq)` and `_getMemoryMapOfFile(seq)` throw `Invalid transaction log file` instead of aborting. It passes here on macOS. Worth noting the second assertion only works because `49cc9588` re-closes the file and zeroes `size` on a failed open, so the second call re-attempts the open rather than short-circuiting.

Resolving. Good to see the sibling paths swept rather than just the one that was reported.

—
_Generated by Barber AI (Claude Opus 5)_

test/backup-transaction-logs.test.ts
```diff
@@ -33,6 +36,29 @@ async function writeLog(db: RocksDatabase, name: string, count = 5, fill = 'x'):
...
 }
 
+function writeDiscoveredSegments(dbPath: string, name: string): string {
+	const logDirectory = join(dbPath, 'transaction_logs', name);
+	mkdirSync(logDirectory, { recursive: true });
+	const header = Buffer.alloc(TRANSACTION_LOG_FILE_HEADER_SIZE);
+	header.writeUInt32BE(TRANSACTION_LOG_TOKEN, 0);
+	header.writeUInt8(1, 4);
+	header.writeDoubleBE(Date.now(), 5);
+	for (let sequence = 64; sequence >= 1; sequence--) {
```
cb1kenobi: **High: after the rebase these three tests fail on every platform — the fixture no longer produces a lazily-registered segment**

CI on `e61ffb76` is red on all 15 JS jobs (macOS, Linux, Windows), and it is these three, reproduced locally on macOS:

- `rejects a transaction log snapshot when a segment cannot be opened (directory)` — got `Backup failed: Corruption: Transaction log file shorter than recorded size: .../unopenable/2.txnlog`
- `... (stream)` — got `Backup stream failed: IO error: While reading file sequentially: .../unopenable/2.txnlog: Is a directory`
- `consistently excludes a malformed lazy segment across backup preflight and copy` — the segment is present in the snapshot (`existsSync` true)

Root cause is base drift, not this diff. `2869eedc` on `kris/txnlog-committed-position-recovery` replaced load()'s "inspect only the predecessor" bound with a walk that continues until a boundary is found and only breaks at the flushed file. This fixture writes 64 header-only segments and no `txn.state`, so `flushedPosition.logSequenceNumber == 0`, both `break` guards are dead, and `TransactionLogStore::load()` opens and closes every segment looking for a boundary. `close()` leaves `size` set, so `ensureExtent()` hits its `size != 0` early return for all 64 — `snapshotForBackup()` never opens the replaced path, and the failures above come from the copy stage instead of the snapshot.

`e61ffb76`'s creation-order reversal cannot fix this: readdir order is not creation order. Measured on APFS with both loop directions — identical readdir order (`1,10,11,...,19,2,...`) and the same picked segment (`2`) either way.

Suggested fix: stop deriving laziness from discovery order and give the *highest* segment one complete transaction, so load()'s boundary walk stops at the current file and every lower segment stays unopened. Writing `header + [timestamp, length=4, flags=1] + 4 bytes` to `64.txnlog` makes all 14 tests in this file pass locally. Confirmed the underlying fix is still sound with that fixture in place: reverting `transaction_log_store.cpp:453` to `catch (const std::exception&)` re-fails exactly the two `rejects ...` cases with `promise resolved "1"` and `promise resolved "undefined"`, so it is the fixture that regressed, not the format/environmental split.

Worth a separate look on #723 as well: with no `txn.state`, load() now opens and fully reads every segment at startup, which is the unbounded full-history scan the removed comment was guarding against.

—
_Generated by Barber AI_

test/transaction-log.test.ts
```diff
@@ -704,6 +704,41 @@ describe('Transaction Log', () => {
...
 
+		it('reports lazy-segment open failures as JavaScript errors', () =>
+			dbRunner({ skipOpen: true }, async ({ db, dbPath }) => {
+				const logDirectory = join(dbPath, 'transaction_logs', 'open-error');
+				await mkdir(logDirectory, { recursive: true });
+				const header = Buffer.alloc(TRANSACTION_LOG_FILE_HEADER_SIZE);
+				header.writeUInt32BE(TRANSACTION_LOG_TOKEN, 0);
+				header.writeUInt8(1, 4);
+				header.writeDoubleBE(Date.now(), 5);
+				for (let sequence = 64; sequence >= 1; sequence--) {
```
cb1kenobi: **Low: the same reversal here is a no-op, and this test no longer establishes that the segment is lazy**

`readdir` order is independent of creation order on APFS/ext4/NTFS (measured: identical order and identical picked sequence with the loop in either direction), so flipping this loop changes nothing. This test still passes only because `getLogFileSize()` / `_getMemoryMapOfFile()` open the segment regardless of whether load() already resolved it — so the `lazySequence` it computes is no longer necessarily lazy, and the name over-claims what is covered.

Suggested fix: share one fixture helper with `test/backup-transaction-logs.test.ts` that makes the highest segment carry a complete transaction (so load()'s boundary walk stops there and lower segments really do stay unopened), and use it in both places.

—
_Generated by Barber AI_

src/binding/transaction_log/transaction_log_store.cpp
kriszyp: Addressed in 4d7b54e4. The snapshot comment now distinguishes successful one-time extent resolution from malformed retries, and a per-file atomic gate emits the malformed-segment warning only once. The regression test observes one warning even though disk-space preflight and copy both collect the snapshot.

src/binding/database/backup_transaction_logs.h
kriszyp: Addressed in 81e7dda5. The helper contract now explicitly documents that malformed headers are warned about and omitted because they contain no readable entries, while every other open/read failure rejects the incomplete snapshot.

test/backup-transaction-logs.test.ts
kriszyp: Addressed in 4d7b54e4. The shared fixture now puts a complete transaction in the highest segment, stopping the startup boundary walk there and leaving lower segments genuinely lazy. Both backup failure cases and the malformed-segment consistency case pass in the full 797-test suite.

test/transaction-log.test.ts
kriszyp: Addressed in 4d7b54e4. This test and the backup tests now share writeLazyTransactionLogSegments(), whose highest segment contains a complete transaction so the selected lower segment is demonstrably unopened rather than dependent on creation/readdir order.

cb1kenobi (approved):

PR #756, opened 2026-08-06
Honor dbWriteBufferSize open option

kriszyp:
Human-Review-Need: 4 @ 8ca8136556c9
`dbWriteBufferSize` was declared in `src/store.ts`, documented in `db_options.h`, and parsed into the handle options in `database.cpp` — then never read. `db_descriptor.cpp` set `db_write_buffer_size = 32 << 20` unconditionally, with a comment noting it might want to be configurable. A caller setting the option got 32 MiB regardless.

This honors it, validates the incoming value (non-negative safe integer), and **changes the default from 32 MiB to 0** — the global trigger is now off unless a caller asks for it.

## Why default to 0 rather than preserving 32 MiB

`db_write_buffer_size` is **one budget divided among every column family**, so any fixed default has a throughput cliff at some family count — it just depends where. Measured with a round-robin ingest across N column families on durable storage:

At a 16 MiB per-CF `writeBufferSize` (128 MiB ingest):

| column families | 32 MiB | 256 MiB | disabled |
|---|---|---|---|
| 8 | 211.9 MiB/s, 120 flushes, no stall | 232.6 | 229.5 |
| 16 | **1.0 MiB/s**, 57,504 flushes, 106 s stall | 228.3 | 232.3 |
| 32 | **0.7 MiB/s**, 87,220 flushes, 142 s stall | 216.8 | 232.2 |

At a 64 MiB per-CF `writeBufferSize` (1 GiB ingest, two independent runs), the same cliff appears at 16 families: ~515,000 flushes, 0.1 MiB/s, and roughly 85 minutes of accumulated write stall against ~220 MiB/s for every other arm.

The failure mode is thrashing: with the budget oversubscribed, RocksDB repeatedly flushes "the largest memtable" and ends up writing ~2 KB per flush. The rule that fits both runs is that throughput collapses once `familyCount × writeBufferSize` oversubscribes the budget by roughly 4×.

`disabled` was fastest or tied-fastest in essentially every configuration measured, and it is the only value with no cliff at any family count.

## The tradeoff this accepts

Defaulting to 0 removes an implicit bound on total memtable memory for any consumer that does **not** configure a WriteBufferManager. Without either, memtable memory is bounded only per family, so it scales with family count.

That is the right place for the bound anyway: a WriteBufferManager bounds memtable memory across databases rather than dividing one fixed budget among families, and it does not care how many families there are. Callers who want a cap should set one; callers who genuinely want the global trigger can now actually set it, which they could not before this PR.

**Release note worth carrying:** a database with many column families and no WriteBufferManager will hold more memtable memory after this change than before — bounded by `writeBufferSize × maxWriteBufferNumber × familyCount` rather than by the old 32 MiB ceiling. That is the intended trade (the old ceiling is what caused the cliff above), but it could surprise a memory-constrained deployment on upgrade.

Also worth knowing, surfaced in review: like every other database-wide option here, this binds at the **first** open of a path. The descriptor is process-global, so a later open of the same path — including from another worker thread — silently keeps the first opener's value. Documented in both the TS and native option comments rather than newly reconciled, since the same is already true of `maxOpenFiles`, `maxWriteBufferNumber` and the WriteBufferManager.

## Where to look

- `src/binding/options/db_options.h` — the default and the reasoning.
- `src/binding/database/db_descriptor.cpp` — one line: the hardcode becomes `options.dbWriteBufferSize`.
- `src/binding/database/database.cpp` — validation of the JS value (rejects NaN/Infinity/negative/non-integer/beyond-2^53).

## Verification

`pnpm test` full suite: 726 passed / 5 skipped / 54 files (macOS, Node 24).

`test/db-options.test.ts` covers the default (global trigger off — 96 MiB written under a 256 MiB per-CF buffer produces no SST), an explicit `0`, an explicit non-zero value taking effect, and the rejected-input cases. The default test is the inverse of what it asserted before this change, which is the intended behavior change made visible.

Benchmark harness used for the numbers above lives on `bench/db-write-buffer-size-multicf` and is not part of this PR.

Generated by Claude Opus 5.


kriszyp:
## Benchmark data behind the `0` default

Posting the full measurements, including a re-run under real compaction that wasn't available when the description was written.

All runs: round-robin ingest of 1 KB inline values across N column families, per-CF `writeBufferSize` 16 MiB (Harper's configured value), on durable NVMe storage, on an otherwise idle box.

### Under real compaction (scattered keys, W-Amp ~4.5), 256 MiB ingest

| column families | `dbWriteBufferSize: 32 MiB` | `256 MiB` | `disabled (0)` |
|---|---|---|---|
| 8 | 191.0 MiB/s, 256 flushes, no stall | 198.7 MiB/s | 200.4 MiB/s |
| 16 | **0.9 MiB/s**, 122,756 flushes, 242 s stall | 207.5 MiB/s | 201.9 MiB/s |

### Without compaction (sequential keys, W-Amp 1.0), 128 MiB ingest

| column families | `32 MiB` | `256 MiB` | `disabled (0)` |
|---|---|---|---|
| 8 | 211.9 MiB/s, 120 flushes, no stall | 232.6 MiB/s | 229.5 MiB/s |
| 16 | **1.0 MiB/s**, 57,504 flushes, 106 s stall | 228.3 MiB/s | 232.3 MiB/s |
| 32 | **0.7 MiB/s**, 87,220 flushes, 142 s stall | 216.8 MiB/s | 232.2 MiB/s |

Compaction costs roughly 10% across every arm and changes nothing else — the cliff sits in the same place with the same magnitude either way.

An earlier run at a 64 MiB per-CF `writeBufferSize` (1 GiB ingest, two independent repetitions) put the same cliff at 16 families: ~515,000 flushes, 0.1 MiB/s, and about 85 minutes of accumulated write stall, against ~220 MiB/s for every other arm.

### Reading of the data

The collapse happens when `familyCount × writeBufferSize` oversubscribes the budget by roughly 4× or more. 8 families × 16 MiB = 128 MiB against a 32 MiB budget (4×) still holds; 16 families = 256 MiB (8×) is over the edge. The failure mode is thrashing — with the budget oversubscribed, RocksDB repeatedly flushes "the largest memtable" and ends up writing ~2 KB per flush.

Any fixed default has this cliff at *some* family count, because the budget is divided among families and the family count isn't knowable when the value is chosen. `0` is the only setting with no cliff, and it was fastest or tied-fastest in every configuration measured.

### Methodology notes

Several of these bit me before the numbers were right, so they're worth recording for anyone re-running:

- **Keep values under 2048 bytes.** At or above `min_blob_size` values go to blob files rather than SSTs, and the SST/flush/compaction accounting stops reflecting the data volume — a 4 KB-value run showed 640 KB of SST for 205 MB of data.
- **Values must be genuinely incompressible.** Default LZ4 compresses a repeated-character value roughly 300:1. A *small* random pool isn't enough either, since block compression exploits cross-record redundancy when slices overlap; use a large pool with widely spaced offsets.
- **Key order decides whether compaction runs at all.** Strictly ascending keys give each flushed memtable a disjoint range, so RocksDB places each file directly at the bottommost level and never compacts (`compact.write.bytes` = 0, W-Amp 1.0 — truthfully). Scattered keys produce real work: 744 MB of compaction writes, W-Amp 4.5.
- **`rocksdb.num-files-at-level<N>` is a string property.** `getDBIntProperty()` returns `undefined` for it; `getDBProperty()` returns `"21"`. The benchmark's `levelFiles` column currently reads it through the int getter, so it prints `none` regardless of the real layout — worth fixing before this lands, since a column that can only say one thing is worse than no column.
- **Benchmark on durable storage.** `/tmp` is a tmpfs on this host, where `fsync` is a no-op; an earlier version of these numbers was measured there and had to be redone.

Flush counts, flush latency, ingest throughput, stall time and total SST size are trustworthy in the tables above. Level distribution is not, for the reason noted.

🤖 Measured and written by Claude Opus 5

cb1kenobi (approved):

Issue #754, opened 2026-08-05
EventListener reaches the DBDescriptor via a late-bound, racy weak_ptr — data race, lost-during-open background errors, and a purge-skip leak


cb1kenobi:
## Summary

The shared RocksDB `EventListener` (`TransactionLogEventListener` in
`src/binding/database/db_descriptor.cpp`) reaches the `DBDescriptor` through a
`std::shared_ptr<std::weak_ptr<DBDescriptor>>` that is **bound after `DB::Open`
returns** (`*descriptorWeakPtr = descriptor`), while RocksDB background threads
can invoke the listener's callbacks *during and after* open. This is a
pre-existing pattern (it predates the background-error work); a Codex
cross-model review of the `#730` branch surfaced three concrete hazards in it.
Filing as a scoped follow-up so `#730` can land without expanding into a
teardown-lifecycle refactor.

The listener currently overrides `OnFlushBegin` / `OnFlushCompleted` and (on the
`#730` branch) `OnBackgroundError` / `OnErrorRecoveryEnd`.

### 1. Data race on the shared `weak_ptr` (UB) — blocker

The opening thread executes `*descriptorWeakPtr = descriptor` while a background
flush/compaction thread may be calling `descriptorPtr->lock()` on the *same*
`weak_ptr` object. `std::weak_ptr` thread-safety only covers *distinct*
smart-pointer objects sharing a control block — concurrent read + write of one
`weak_ptr` object is a data race and undefined behavior (potential corruption /
crash). Affects the flush callbacks today; the `#730` background-error callbacks
inherit it.

### 2. A background error latched during `DB::Open` is lost (`#730`-specific) — significant

If RocksDB schedules recovery flush/compaction during `DB::Open` and it fails
*before* the weak pointer is bound, `descriptorPtr->lock()` returns null and the
callback returns without recording anything. Open can then succeed with RocksDB
internally latched read-only while `db.backgroundError` reports `null`.

### 3. A callback pin can permanently skip the descriptor purge (leak) — significant

Both callback bodies promote the descriptor to a `shared_ptr` for their
duration. If the final JS handle closes while a callback holds that reference,
`DBRegistry::PurgeIfUnreferenced()` observes `use_count > 1` and skips teardown;
when the callback returns and drops its ref, nothing re-runs the purge. The
registry entry and the open RocksDB instance leak. This is exactly the
skipped-purge / retry-after-release hazard documented in `AGENTS.md` invariant
[...]

PR #758, opened 2026-08-06
feat(database): observe and recover from latched RocksDB background errors (#730)

### LINKED ISSUE #730: No in-process recovery from a latched background error; only a restart clears it
## Summary

When a write fails at the filesystem level, RocksDB latches a background error and refuses all further writes. There is currently **no way for a consumer to recover in-process** — no exposed `DB::Resume()`, and no way to observe the latched state. The only remedy is restarting the process, so a transient, self-correcting condition becomes an indefinite outage.

## Observed

An instance exhausted its filesystem quota. RocksDB recorded exactly one background error and stopped accepting writes:

```
rocksdb.error.handler.bg.error.count                COUNT : 1
rocksdb.error.handler.bg.io.error.count             COUNT : 1
rocksdb.error.handler.bg.retryable.io.error.count   COUNT : 0
rocksdb.error.handler.autoresume.count              COUNT : 0
rocksdb.error.handler.autoresume.success.count      COUNT : 0
```

Note `bg.retryable.io.error.count: 0` — the error was classified **non-retryable**, so RocksDB's own auto-resume path was never eligible. That is arguably correct classification; the problem is what follows.

Every subsequent write failed with a **stored, stale** message naming a WAL file that no longer existed on disk (the newest WAL present was several generations older). This repeated **19,149 times over 7h45m**. Freeing space did not help — the quota was more than doubled and writes kept failing with the identical message. Only reopening the database recovered it.

Two consequences worth separating:

1. **No in-process recovery.** Once latched, the state is unreachable from JS. The caller cannot retry, cannot call `Resume()`, and cannot even detect that it is latched rather than merely failing.
2. **The latch blocks the engine's own cleanup.** While in the error state, RocksDB could not delete obsolete files. Reopening the database immediately freed **~644 MB** of files it had been holding. So a database that fails because storage is full then *keeps itself* full — the failure is self-sustaining rather than self-correcting.

## Why this hits self-hosted consumers hardest

A full disk is routine outside managed environments. The current behavior converts "disk filled up briefly, then I freed space" into "the database is read-only until someone notices and restarts it," with no diagnostic that says so — the error message actively misleads, since it names a file that does not exist.

## Suggested direction

- Expose `DB::Resume()` so a consumer can attempt recovery after the underlying condition clears.
- Expose the latched background-error state (and ideally its retryable/non-retryable classification) so a caller can distinguish "this write failed" from "this database is now read-only."
- Consider whether the error message surfaced to callers should be regenerated rather than replayed from the stored error, since the stored text becomes stale and misdirects diagnosis.
- Worth considering whether obsolete-file cleanup can proceed while in a write-latched state, given that it is the action most likely to clear the cause.

---

cb1kenobi:
## Summary

Exposes observability and in-process recovery for RocksDB background errors ([#730](https://github.com/HarperFast/rocksdb-js/issues/730)). When a write fails at the filesystem level (a full disk / exhausted quota), RocksDB records a background error and — for a hard-or-worse severity — stops accepting writes. Previously the only remedy was a process restart, with no way to even observe the state.

## API

- **`'error'` event** (per-database, `db.on('error', ...)`) — emitted with a `BackgroundError` when RocksDB reports a background error. The **push** surface.
- **`db.getLastError(): BackgroundError | null`** — the most recent background error, or `null`. The **pull** surface: for on-demand checks (a health probe) or to catch an error that fired before a listener was attached. Purely historical — **not** cleared by `resume()`.
- **`db.resume(): void`** — calls RocksDB's `DB::Resume()`; on success writes are accepted again (and RocksDB resumes the obsolete-file cleanup it was blocking), on failure it throws with the RocksDB error and the database stays read-only. A no-op on a healthy database.
- **`BackgroundError`** — a real `Error` subclass, exported from the package, so `err instanceof BackgroundError` (and `instanceof Error`) both hold. Fields: `message`, `name` (`'BackgroundError'`), `type` (`'background'`), `severity`, `severityName`, `writesDisabled`, `reason?`, `reasonName?`.

`writesDisabled` (severity ≥ hard) is the "RocksDB has stopped accepting writes" signal — named to keep it distinct from opening a database in read-only mode. Only then is `resume()` warranted; a soft error auto-recovers.

## How it works

`EventListener::OnBackgroundError` runs on a RocksDB background thread. It serializes the error to a JSON string via the Node-free `backgroundErrorToJson` (`core/background_error`) and stores it on the process-global `DBDescriptor` as a plain `std::string` under a mutex — **no `napi_env`/N-API state crosses the thread boundary**. That removes the whole class of cross-thread/env hazards the earlier persisted-mirror design had to guard against (no generation counter, no `OnErrorRecoveryEnd` reconciliation).

Both surfaces reconstruct one identical `BackgroundError` from that string on the JS thread: `getLastError()` (a `Database` method) and the `'error'`-event trampoline in the generic `EventEmitter`. The `BackgroundError` class is defined natively (`napi/background_error.cpp`), wired as a genuine `Error` subclass (prototype chain to `Error`, `Error.captureStackTrace`), and its constructor is stashed per-env via `napi_set_instance_data` so either path can build an instance holding only an `napi_env`.

## Tests

- **Native GoogleTests (9):** severity/reason name mapping, `writesDisabled`-by-severity, and `backgroundErrorToJson` field/escape/omit-reason coverage (Node-free).
- **Vitest (3):** healthy no-op (`getLastError()` is `null`, no event fires); a real read-only-directory write failure emits a `writesDisabled` `BackgroundError` (asserts `instanceof BackgroundError`), `getLastError()` returns the same error, and writes then throw; `resume()` reports failure honestly rather than silently reporting success. The read-only-directory reproduction skips on Windows and as root (matching the existing file-lock permission tests); CI runners are non-root.
- Full suite: **759 passing**.

Docs: README (`'error'` event, `db.getLastError()`, `db.resume()`) + the exported `BackgroundError` type/value.

Closes #730

---

> **Design note:** this supersedes the original `db.backgroundError` getter backed by a persisted, thread-safe `DBDescriptor` mirror. Because nothing env-bound is now persisted (only a JSON string), the mirror's race surface is gone by construction — the three items deferred to [#754](https://github.com/HarperFast/rocksdb-js/issues/754) that concerned the mirror (`weak_ptr` publication race, error-lost-during-`DB::Open`) no longer apply here; the callback-pin purge-skip leak is unrelated and remains tracked.


src/binding/database/db_descriptor.cpp
```diff
@@ -199,6 +199,47 @@ class TransactionLogEventListener : public rocksdb::EventListener {
...
+	// latched yet; do NOT overwrite with the recovery-attempt status.
+	void OnErrorRecoveryEnd(const rocksdb::BackgroundErrorRecoveryInfo& info) override {
+		if (!descriptorPtr) {
+			return;
+		}
+		auto desc = descriptorPtr->lock();
+		if (!desc) {
+			return;
+		}
+		if (info.new_bg_error.ok()) {
```
kriszyp: Recovery reconciliation is not generation-safe here, despite `resume()` explicitly preserving that invariant. These callbacks run on RocksDB background threads: error B can latch after recovery A begins, then A's successful callback unconditionally clears B. The failed branch has a similar check-then-act race: it can observe no error, allow B to latch, and then overwrite B with A's `old_bg_error`. Could this be one atomic descriptor operation under `backgroundErrorMutex`, conditioned on the error/generation being recovered? A deterministic interleaving test would protect both paths.

— KrAIs (Codex)

src/binding/database/db_descriptor.cpp
```diff
@@ -199,6 +199,47 @@ class TransactionLogEventListener : public rocksdb::EventListener {
 		}
 	}
 
+	// Mirrors RocksDB's latched background error onto the descriptor so a consumer
+	// can observe it and recover in-process (HarperFast/rocksdb-js#730). We do NOT
+	// suppress the error (leaving *bg_error untouched) — the point is to surface
+	// the read-only latch, not hide it. Runs on flush/compaction/write threads, so
+	// it must stay cheap: a mutex-guarded copy, nothing blocking.
+	void OnBackgroundError(rocksdb::BackgroundErrorReason reason, rocksdb::Status* bg_error) override {
```
kriszyp: This drops callbacks until `*descriptorWeakPtr = descriptor` at line 1091. The listener is installed before `DB::Open`, and RocksDB can start background work once open returns; the subsequent column-family setup widens the interval in which a flush/compaction error sees an empty weak pointer. The database can therefore become write-latched while `backgroundError` permanently reports `null`. Could the listener retain synchronized pending error/recovery state before attachment and transfer it atomically when the descriptor is ready? A callback-before-attachment test seam would cover this startup path.

— KrAIs (Codex)

src/database.ts
```diff
@@ -269,6 +270,30 @@ export class RocksDatabase extends DBI<DBITransactional> {
 		return this.store.db.columns;
 	}
 
+	/**
+	 * The RocksDB background error currently latched on this database, or `null`
+	 * when the database is healthy. When a write fails at the filesystem level
+	 * (e.g. a full disk or exhausted quota), RocksDB latches a background error
+	 * and refuses all further writes until recovery, so a non-null value means the
```
kriszyp: This overstates what a non-null value means. RocksDB soft severity (`severity === 1`) is auto-recoverable and does not represent the hard read-only latch; this documentation later identifies `severity >= 2` as the write-stopping signal. Consumers following the example may report a writable DB as read-only and invoke synchronous `resume()` for a transient soft error. Could we describe non-null as an observed background error, reserve read-only/manual recovery guidance for `severity >= 2`, and update the README accordingly? An `isReadOnly` field could also keep callers from duplicating the numeric threshold.

— KrAIs (Codex)

kriszyp (commented):
This looks like an extremely valuable addition for robustness, nice!
Is there, or should there, be a way for JS to notified of a background error? I assume we don't want to just poll `db.backgroundError`? Or would an error surface as a commit failure? (I'd have guessed not if it is a "background" error).
Anyway, codex some comments as well.

src/binding/database/db_descriptor.cpp
cb1kenobi: Good catch — this lost-during-`DB::Open` window is fixed in the stacked follow-up [#760](https://github.com/HarperFast/rocksdb-js/pull/760): the listener now writes to a `DBEventListenerState` created *before* `DB::Open` (not a weak pointer bound afterward), so an error latched mid-open is captured and transferred to the descriptor once it exists. I scoped it there to keep this PR focused on the observe/recover surface, but happy to fold it back here if you'd prefer it in one PR.

Generated by Barber AI 🤖

src/binding/core/background_error.cpp
```diff
@@ -0,0 +1,113 @@
...
+	const std::string& recoveredMessage,
+	int recoveredSeverity,
+	bool recovered
+) {
+	std::lock_guard<std::mutex> lock(this->mutex_);
+	if (recovered) {
+		// Clear only if the currently-latched error IS the one that was
+		// recovered. A newer error latched on another thread between recovery
+		// start and this callback has a different message and must survive.
+		if (this->info_.latched && this->info_.message == recoveredMessage) {
```
kriszyp: The mutex makes reconciliation atomic, but message equality does not establish that this is still the same latch. Error A can begin recovery, then a newer error B can latch with the same RocksDB message (for example, repeated `No space left on device` from the same path); A's successful callback then matches B's message and clears the mirror even though B remains active. The added test only covers B having a different message. Could recovery carry the mirror generation captured when that specific recovery begins and clear with `clearIfUnchanged(generation)` (for example via the corresponding recovery-begin callback/token), then add an identical-message A/B test?

— KrAIs (Codex)

src/binding/database/db_descriptor.cpp
```diff
@@ -235,6 +235,46 @@ class TransactionLogEventListener : public rocksdb::EventListener {
...
```
kriszyp: This drops callbacks until `*descriptorWeakPtr = descriptor` at line 1134. The listener is installed before `DB::Open`, and RocksDB can start background work once open returns; the subsequent column-family setup widens the interval in which a flush/compaction error sees an empty weak pointer. The database can therefore become write-latched while `backgroundError` permanently reports `null`. Could the listener retain synchronized pending error/recovery state before attachment and transfer it atomically when the descriptor is ready? A callback-before-attachment test seam would cover this startup path.

— KrAIs (Codex)

kriszyp (commented):
I am still not sure how to listen for background errors.

---
Proposed inline comments (anchors failed):
- `src/binding/database/db_settings.cpp:209`: This runtime mutation bypasses the new open-time history safeguard. A database can attach a WBM while `allowStall` is false, so its column families retain RocksDB's derived `max_write_buffer_size_to_maintain`; changing `writeBufferManagerAllowStall` to true here immediately enables stalling on that same manager without revisiting those existing targets. With a budget below the retained history, writes can still stall permanently—the root failure this change is meant to remove. The size-0 path has a related mismatch for late-created families: existing DBs retain the manager, while `resolveMaxWriteBufferSizeToMaintain()` consults the global size and may derive history again. Could we enforce this invariant against the manager actually attached to each descriptor, either by making `allowStall` immutable after attachment or safely updating/rejecting incompatible CF targets before enabling it? Please add coverage for enabling stall after open and for creating a CF after global size is reset to zero.

— KrAIs (Codex)
- `test/write-buffer-manager-stall.test.ts:41`: `db.put()` is declared async, but it calls `store.putSync()` before returning its promise. If RocksDB stalls inside this native call, the JS worker thread is blocked and Vitest's 60-second timeout cannot execute; the regression can hang until the outer CI job timeout rather than fail this test. Could this scenario run in a worker/child process that the parent can terminate on a deadline, or use a genuinely asynchronous native operation so the test runner remains able to enforce the timeout?

— KrAIs (Codex)

src/binding/database/db_descriptor.cpp
```diff
@@ -199,6 +199,47 @@ class TransactionLogEventListener : public rocksdb::EventListener {
...
+		if (!descriptorPtr) {
+			return;
+		}
+		auto desc = descriptorPtr->lock();
+		if (!desc) {
+			return;
+		}
+		if (info.new_bg_error.ok()) {
+			desc->clearBackgroundError();
+		} else {
```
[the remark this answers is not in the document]
cb1kenobi: This is no longer relevant.

cb1kenobi:
> I am still not sure how to listen for background errors.

Good question — it's exactly what drove a redesign of this surface. The polling `db.backgroundError` getter is gone; background errors are now **pushed** as an event, with a pull fallback:

```js
import { BackgroundError } from '@harperfast/rocksdb-js';

db.on('error', (err) => {          // err instanceof BackgroundError (extends Error)
  if (err.writesDisabled) {        // hard+ severity: RocksDB has stopped accepting writes
    // ...free disk space, then:
    db.resume();
    db.setLastError(null);         // reset the historical error once recovered
  }
});
```

- **Push:** `db.on('error', …)` fires with a `BackgroundError` (`message`, `severity`, `severityName`, `writesDisabled`, `reason?`).
- **Pull:** `db.getLastError()` returns the most recent one (or `null`) — for a health probe, or to catch one that fired before a listener was attached.
- On your other question: a background error is asynchronous (a failed flush/compaction), so it doesn't map to a specific commit — that's why the event exists. But once RocksDB has stopped writes, subsequent writes *do* throw synchronously, so you also find out at the next write.

The mirror/generation machinery your Codex comments flagged is gone too — nothing is persisted as env-bound state, so those races no longer exist. PR description updated to match.

Generated by Barber AI 🤖

src/binding/database/db_descriptor.cpp
```diff
@@ -235,6 +237,34 @@ class TransactionLogEventListener : public rocksdb::EventListener {
...
 
+	// Surfaces a RocksDB background error to JS (HarperFast/rocksdb-js#730).
+	// Serializes it to a JSON string and hands it to `setLastError`, which stores
+	// it (readable on demand via `db.getLastError()`) and emits the `'error'`
+	// event — both reconstruct the same `BackgroundError` from this string on the
+	// JS thread, so nothing N-API/env-bound is touched here. We do NOT suppress
+	// the error (leaving *bgError untouched) — the point is to surface it, not
+	// hide it. Runs on flush/compaction/write threads; storing a string and the
+	// thread-safe, asynchronous emit keep this cheap and non-blocking.
+	void OnBackgroundError(rocksdb::BackgroundErrorReason reason, rocksdb::Status* bgError) override {
```
kriszyp, line 307: This still drops every callback until `*descriptorWeakPtr = descriptor` after `DB::Open()` and column-family setup. An open-time background error therefore never reaches either the event or `getLastError()`, even though the latter is documented as catching errors that occurred before listener attachment. Please retain synchronized pending listener state and transfer it when the descriptor is attached, or give the listener stable storage before opening.

— KrAIs (Codex)

src/database.ts
```diff
@@ -418,6 +420,75 @@ export class RocksDatabase extends DBI<DBITransactional> {
...
+	 * {@link BackgroundError}. `type` defaults to `'background'`.
+	 *
+	 * @example
+	 * ```typescript
+	 * db.resume();
+	 * db.setLastError(null); // recovered — clear the historical error
+	 * ```
+	 */
+	setLastError(error?: BackgroundErrorOptions | null): void {
+		return this.store.db.setLastError(error == null ? null : { type: 'background', ...error });
```
kriszyp: `BackgroundErrorOptions` makes `severity`, `severityName`, and `writesDisabled` optional, but this forwards the object without supplying defaults, and the native constructor copies only present properties. Thus the explicitly valid `db.setLastError({ message: 'boom' })` makes `getLastError()` return an object whose required `BackgroundError` fields are all `undefined`. Please either require those fields in the input, make them optional on the returned type, or normalize deterministic defaults here; add a minimal-object round-trip test as well.

— KrAIs (Codex)

kriszyp (commented):
This is great, love the API!
Codex has some mechanical suggestions.
---
Proposed inline comments (anchors failed):
- `src/binding/database/db_settings.cpp:209`: This runtime mutation can enable `allowStall` after column families were opened with the derived retained-history target. That bypasses the open-time safeguard which zeroes history only when a stalling manager already exists, recreating the permanent stall when retained history exceeds the manager budget. Please enforce the invariant when reconfiguring the manager—either reject this transition while affected databases are open or update their history targets before enabling stalls—and cover the false-at-open/true-after-open sequence.

— KrAIs (Codex)
- `src/binding/transaction/transaction.cpp:607`: Capturing this `shared_ptr` in work queued on `descriptor->commitWorker`/`logWorker` creates `descriptor -> worker queue -> lambda -> descriptor` ownership. If worker-env teardown closes the transaction and final DB handle while the task is queued, `CloseDB()` sees this extra reference and `PurgeIfUnreferenced()` skips. When the task later releases the capture, nothing retries the purge, so the registry entry and open RocksDB instance remain indefinitely. This is the same skipped-purge invariant handled explicitly by backup/checkpoint state destructors. Please arrange a purge retry after releasing the task’s descriptor pin without allowing descriptor destruction on its own worker, and add a final-handle teardown regression test.

— KrAIs (Codex)
- `test/write-buffer-manager-stall.test.ts:41`: This timeout cannot reliably detect the intended regression. `db.put()` enters the synchronous native write before returning its promise; if RocksDB stalls there, the Vitest worker’s event loop is blocked and its timeout callback cannot run. Please execute the scenario in a child process or worker that the parent can terminate on a real wall-clock deadline, or use a genuinely non-blocking native path.

— KrAIs (Codex)

src/database.ts
cb1kenobi: Fixed.

src/binding/database/db_descriptor.cpp
kriszyp: This still drops every callback until `*descriptorWeakPtr = descriptor` after `DB::Open()` and column-family setup. An open-time background error therefore never reaches either the event or `getLastError()`, even though the latter is documented as catching errors that occurred before listener attachment. Please retain synchronized pending listener state and transfer it when the descriptor is attached, or give the listener stable storage before opening.

— KrAIs (Codex)

cb1kenobi:
> `src/binding/transaction/transaction.cpp:607`: Capturing this `shared_ptr` in work queued on `descriptor->commitWorker`/`logWorker`… `PurgeIfUnreferenced()` skips… nothing retries the purge…

Good catch. This is the same purge-skip class as the EventListener callback pin, just at a different site (the commit/log worker rather than the flush listener), and it predates #730. Rather than expand this PR, I've folded it into #754 ([comment](https://github.com/HarperFast/rocksdb-js/issues/754#issuecomment-5365420835)) — it'll be fixed alongside the listener-lifecycle rework in #760 (`fix/730-listener-lifecycle`), where the retry-after-pin-release discipline already lives.

The other two mechanical items from your review are tracked as well: the WriteBufferManager `allowStall` runtime-mutation and stall-test concerns in #781, and the open-time drop in #754 / #760.

Generated by Barber AI 🤖

src/binding/database/db_descriptor.cpp
cb1kenobi: The doc overclaim is addressed here in #758 — `getLastError()` no longer promises to catch an error from *during* the initial open; it now notes that narrow startup window is handled separately. The actual capture fix — retaining synchronized pending listener state before `DB::Open()` and transferring it once the descriptor is attached (exactly as you suggest) — is #760 (fixing #754 hazard #2), which replaces the late-bound `weak_ptr` with a pre-open listener-state object.

Generated by Barber AI 🤖

src/binding/database/db_descriptor.cpp
```diff
@@ -149,24 +151,85 @@ struct JobTracker final {
...
+	// descriptor (after) — never lost.
+	void publishDescriptor(std::shared_ptr<DBDescriptor> descriptor) {
+		std::string pending;
+		{
+			std::lock_guard<std::mutex> lock(this->mutex_);
+			this->descriptor_ = descriptor;
+			pending.swap(this->pendingError_);
+		}
+		if (!pending.empty()) {
+			descriptor->setLastError(std::move(pending));
```
kriszyp, line 205: Publishing the descriptor before draining `pendingError_` allows a newer callback to overtake the pending error. For example, error A is stashed during `DB::Open`; publication installs `descriptor_` and unlocks; error B then calls `setLastError(B)`; finally this line calls `setLastError(A)`. `getLastError()` now reports stale A, and notifications can arrive B then A—especially unsafe if B has a higher severity or disables writes. Please keep post-publication errors queued until the pending error has been stored, or otherwise serialize the store/emit ordering, and add a deterministic publish/error race test.

— KrAIs (Codex)

kriszyp (approved):
Excellent

PR #761, opened 2026-08-06
docs: document write-buffer and compression open options

cb1kenobi:
## Summary

While reviewing the `RocksDatabase.open()` options, we noticed that several options the native layer accepts and parses (`src/binding/database/database.cpp`) were never listed in the README's options documentation. This PR documents them.

Starting point was `maxWriteBufferSizeToMaintain` (flagged as missing); auditing the native option-parsing block against the README surfaced the full set.

## Options added to the docs

All are valid `open()` options (part of `StoreOptions`, parsed in `database.cpp`) but absent from the enumerated list:

| Option | Default | Notes |
| --- | --- | --- |
| `writeBufferSize` | 16 MB | Per-CF memtable size at which the memtable is sealed and flushed. |
| `maxWriteBufferNumber` | `16` | Max memtables queued per CF before writes stall. |
| `maxWriteBufferSizeToMaintain` | `-1` (derived) | Bytes of memtable history kept for transaction conflict checking. |
| `dbWriteBufferSize` | `0` (disabled) | Global memtable budget across all CFs. |
| `compressionForAllColumnFamilies` | `false` | Already covered in the Compression section, but was missing from the options list — added for discoverability. |

## Notes

- Descriptions and defaults are taken directly from `src/binding/options/db_options.h` and the existing per-field TSDoc in `src/store.ts`.
- Entries are inserted alphabetically to match the existing ordering, and line-wrapped to the file's ~100-char convention.
- Docs-only change — no code touched.


kriszyp:
Nice audit catch — several of these options (`writeBufferSize`, `maxWriteBufferNumber`, `compressionForAllColumnFamilies`) were genuinely missing from the docs.

One accuracy note on the `maxWriteBufferSizeToMaintain` entry, given [#755 "Fix permanent write stall when the derived memtable-history target exceeds the WriteBufferManager budget"](https://github.com/HarperFast/rocksdb-js/pull/755) merged just ahead of this PR: the `-1` derivation description doesn't yet reflect that change. `resolveMaxWriteBufferSizeToMaintain` (`db_descriptor.cpp:83-92`) now resolves the derived default to `0` whenever a stalling `WriteBufferManager` is configured (`writeBufferManagerSize > 0 && allowStall`) — which, per #755's own description, is Harper's default configuration. So `-1` doesn't always land on `maxWriteBufferNumber * writeBufferSize`; it depends on the WriteBufferManager config too.

Suggested wording:

> `-1` (the default) derives from `maxWriteBufferNumber * writeBufferSize` — except when a stalling `WriteBufferManager` is configured (`writeBufferManagerSize > 0` with `allowStall`), in which case it resolves to `0` to avoid retaining memtable history the manager will never release. An explicit non-negative value is always honored as-is.

The `dbWriteBufferSize` entry looks correct against [#756 "Honor dbWriteBufferSize open option"](https://github.com/HarperFast/rocksdb-js/pull/756)'s new `0` default.

— Claude Sonnet 5

kriszyp (commented):
Awesome, great addition! I think it just needs to be updated for #755. I think the comment has a suggestion; its an incredibly confusing property, so I dunno how to word this any better 🤷 .

cb1kenobi:
Thanks Kris — good catch on the #755 interaction. Updated the `maxWriteBufferSizeToMaintain` entry in a9d42cf7 to reflect `resolveMaxWriteBufferSizeToMaintain`:

> `-1` (the default) derives the value from `maxWriteBufferNumber * writeBufferSize` (the RocksDB-recommended default for optimistic transactions) — except when a stalling `writeBufferManager` is configured (`writeBufferManagerSize > 0` with `writeBufferManagerAllowStall`), in which case it resolves to `0` to avoid retaining memtable history the manager will never release. An explicit non-negative value is always honored as-is.

I kept it to the two branches the code actually takes (`>= 0` honored untouched; `-1` derived, with the stalling-manager exception) so it stays accurate without over-explaining a genuinely confusing option. The two Gemini nits (`specified by name`, `16777216` default) were already in the branch.

*Generated by Barber AI 🤖*

README.md
```diff
@@ -53,6 +53,14 @@ Creates a new database instance.
     when linked, else no compression). See [Compression](#compression). Throws if the algorithm is
     not compiled into the native build — check [`supportedCompression`](#supportedcompression) for
     the available list.
+  - `compressionForAllColumnFamilies: boolean` When `true`, applies `compression` to every column
+    family opened for the database rather than only the column family specified by `name`. Requires an explicit
+    `compression`. Defaults to `false`. See [Compression](#compression).
+  - `dbWriteBufferSize: number` The total memtable memory budget in bytes shared across all of the
```
kriszyp, line 59: Could we include the first-open behavior here? `dbWriteBufferSize` is database-wide and is applied only when the process-global descriptor is created; a later open of the same path, including from another worker, silently retains the first opener’s value. Without that note, callers can reasonably expect this option to reconfigure or reject a conflicting open. The `Store.dbWriteBufferSize` API documentation already describes the behavior, so mirroring that sentence here would keep the public README accurate.

— KrAIs (GPT-5)

kriszyp (approved):
Looks good, thank you! One small suggestion.

PR #766, opened 2026-08-08
fix(vt): never vouch for a version the producer marked non-unique

kriszyp:
## What

The VerificationTable answers `FRESH_VERSION_FLAG` from version equality alone. That is only sound
while a value's version identifies it — and a producer that stores two distinct values under one
version has no way to tell us, so the consumer holding the superseded copy is the one we confirm.

This adds one flag to the value-header contract the VT already reads, and stops answering FRESH or
publishing a slot for a value that carries it.

## Why now

Harper hits this. A resequenced (out-of-order) write keeps `Math.max(txnTime, existingVersion)`, so
the merged record is stored under the version it merged onto: same version, different value. The
worker holding the pre-merge value is told it is current, and an `addTo` folding onto that stale
value drops the increment the merge had already applied — the lost counter increments
`ttl-rate-limiter-concurrent.test.ts` "(5) multi-worker stress" has been catching in Harper CI.

Captured on Harper with a detector on the cache-hit path — same key, same version, two values:

```
[QA431stale] {"id":"it5-mw8","cachedVersion":1786209857971.8462,
              "storeVersion":1786209857971.8462,"cached":{"hits":26},"store":{"hits":27}}
```

Harper cannot defend itself against this from JS, which is why the fix belongs here. Parking a
sentinel in the slot only suppresses the pre-read fast path; the post-read equality check does not
consult the slot at all. Verified against the pinned build before writing any of this:

```
post-merge version 1786218151061  count 2   sameVersion true
slot holds sentinel: true                   vouches old V: false
native getSync(expectedVersion=preMergeV) === FRESH_VERSION_FLAG: true
```

## The contract

`extractVersionFromValue()` already documents the header this reads: an 8-byte big-endian version at
offset 0. A producer that has flags to declare writes a 4-byte big-endian metadata word after it —
top byte `VERSION_HEADER_TAG`, low 24 bits producer flags. `VERSION_NOT_UNIQUE_FLAG` (exported as
`constants.VERSION_NOT_UNIQUE_FLAG`, so producers set the bit this reads) is the one flag this
library interprets.

A value carrying it is never answered FRESH and never published to a slot, on every read path —
sync, async, and both transactional ones. Because publication is refused, the pre-read fast path
needs no change of its own: such a version cannot enter a slot, and the write that made it non-unique
already cleared whatever was there.

The flag is only read for a column family that opted into the verification table, i.e. one whose
producer is already writing versions into these bytes. A value with no header, or one whose word is
not tagged, answers "unique" and keeps its existing behaviour — and a misread in the other direction
costs caching, never correctness.

## Latest-value check

FRESH speaks about the consumer's cached copy, not about the value in hand, so a read that may be
behind the latest cannot answer it from what it read: the value that made the version ambiguous can
have been committed after the snapshot, and the consumer may have taken its copy from that newer
value. Every FRESH-answering path therefore consults the latest value when its read may be behind it
— free when the read is provably the latest (no snapshot, or nothing committed since), and on the
branch where it is not free, `vtPopulateIfSettled` was already paying for the same `Get`, which the
check now hands its result to instead of repeating.

On the async paths that check runs in the **worker**, where the database and the caller's column
family are still pinned; the completion only reads a bool. It has to: the completion runs after
teardown may have released both.

## Test notes

- `test/verification-table.test.ts` — a `VERSION_NOT_UNIQUE_FLAG` block: no FRESH when the version
  matches, no publication (via either `POPULATE_VERSION_FLAG` or `expectedVersion`), caching resumes
  once a later write gives the value a version of its own, the flag is read only from a tagged word,
  and the sync / transactional / async / behind-a-snapshot paths each covered.
- `test/native/verification_table_test.cc` — the predicate itself: tagged-and-marked, tagged-and-not,
  other flags in the same word, the same bit in an untagged word, and both too-short shapes.
- `pnpm test` 754 passed / 2 skipped, native GoogleTests 104 passed, `tsc --noEmit` / `oxlint` /
  `oxfmt --check` clean.

## For the human reviewer

Start at `src/binding/core/verification_table.h` — the contract and the predicate are the whole idea,
and the predicate is deliberately permissive in one direction only. Then the four decision sites
(`database.cpp` sync, `database.h` async, `transaction.cpp`, `transaction_handle.cpp`), which are the
same shape each time; the thing to check hardest is that no FRESH-answering path is left ungated, and
that the argument for leaving the pre-read fast path alone holds — it rests entirely on publication
being refused everywhere.

Decisions a reviewer might reasonably question, and why they went this way: the library now parses
one producer flag rather than staying agnostic about value bytes (it already parses the version from
those same bytes, and the alternative — keying the oracle on a per-write token such as the RocksDB
sequence number — is a much wider API change worth doing on its own terms, not as a bug fix); the
tag is 8 bits, so a header-less producer *on a VT-enabled column family* could in principle collide,
which costs caching and not correctness; and `populateVersion()` is left alone, since it never sees a
value and by design trusts its caller.

Cross-model review ran three rounds (codex + gemini + harper-domain) and found real defects in the
first two — the snapshot-value hole and then a use-after-free in my first async attempt. Both are
fixed here; the review's remaining open items are the decisions above.

<sub>Human-Review-Need: 4 @ 6f66027ce5c2</sub>


src/binding/database/database.h
```diff
@@ -429,7 +484,8 @@ void resolveGetResult(
 				state->callResolve(freshResult);
 				return;
 			}
-			if ((state->wantsPopulate || state->hasExpectedVersion) && extracted != 0) {
+			if ((state->wantsPopulate || state->hasExpectedVersion) && extracted != 0
+					&& !state->vtLatest.notUnique) {
```
cb1kenobi: **Medium: async populate/FRESH seeds the snapshot-read version, not the latest — the sync paths' stale-value fix isn't mirrored here**

This is the async counterpart of the decision sites you fixed. The three sync sites publish `populateVersion = latest.read ? latest.latestVersion : extracted`, honoring the stated invariant "the version published is always the LATEST committed version, never a value read at an older snapshot." `resolveGetResult` still publishes `extracted` (populate branch, next line) and `state->expectedVersion` (FRESH branch above) — the value read at the caller's snapshot — even though the worker has already computed `state->vtLatest.latestVersion`.

The non-unique goal is fully covered on this path via `vtLatest.notUnique`; this is the adjacent *unique-but-stale* case. On the transactional async fallthrough (`kReadAllTier`, snapshot behind a newer unique write), the read returns the pre-write value at version V while latest is V'. `vtObserved` is captured at get-issue time — *after* that newer write settled the slot — so the `populateVersionIfUnchanged` CAS-from-observed can still succeed (no write cycle after issue) and seed the stale V into the slot. A later fast-path `verifyVersion(key, V)` would then vouch a superseded version. The CAS guards writes *between read and populate*, but for a snapshotted read the value is already older than `vtObserved`'s world, so that guard doesn't cover this window.

Pre-existing (the old async code published `extracted` too), but it sits on the lines you're changing and the fix is the value you already carry:

```cpp
const uint64_t populateVersion = state->vtLatest.read ? state->vtLatest.latestVersion : extracted;
// FRESH branch and populate branch both publish populateVersion
```

Worth a test on the async txn behind-a-snapshot path — the current behind-a-snapshot test exercises `getBinarySync`, not the async fallthrough.

—
_Generated by Barber AI_

src/binding/database/database.h
```diff
@@ -415,22 +470,24 @@ void resolveGetResult(
...
 
 	if (state->status.IsNotFound() || state->status.ok()) {
-		if (state->status.ok() && state->vtSlot) {
-			// VT check and populate for the async read result.
-			rocksdb::Slice valueSlice(state->value.data(), state->value.size());
+		rocksdb::Slice valueSlice(state->value.data(), state->value.size());
+		if (state->status.ok() && state->vtSlot && !VerificationTable::valueVersionIsNotUnique(valueSlice)) {
 			uint64_t extracted = VerificationTable::extractVersionFromValue(valueSlice);
-			if (state->hasExpectedVersion && extracted != 0 && extracted == state->expectedVersion) {
+			const uint64_t populateVersion = state->vtLatest.read ? state->vtLatest.latestVersion : extracted;
```
cb1kenobi, line 470: **Medium: the async path publishes the latest version without `vtPopulateIfSettled`'s snapshot gates — and it now does so exactly when a snapshot behind that version is known to be open**

This fixes the stale-version publish (thanks), but it lands the *right* version through the *wrong* door. The three sync sites publish via `vtPopulateIfSettled`, which applies the two-gate settled check before touching the slot; `resolveGetResult` calls `populateVersionIfUnchanged` directly, so no gate runs.

That asymmetry was previously benign-ish because the async path published `extracted`. It isn't now: `vtLatest.read == true` is, by `vtCheckLatest`'s own definition, `readSnapshot != nullptr && snapshotSeq < GetLatestSequenceNumber()`. So on every read where this line changes behaviour, the caller's own snapshot is strictly behind the latest write — which is precisely Gate 2's blocking condition (`oldestSnapshotSeq < latestSeq`). Whenever the new value is used, the sync path would have refused to publish at all.

The consequence is the torn view `Transaction::GetBinarySync` documents at transaction.cpp:920-931. Snapshot `S` pinned at seq 100; a unique write `V'` commits at 101; this completion publishes `V'` into the slot with no write cycle and no lock/settle transition; a transaction still reading at `S` then hits the pre-read fast path with `expectedVersion == V'`, gets FRESH, and serves a cached copy its snapshot cannot see. The "pin the snapshot, then load the slot" ordering defends against the write cycle racing the load — it does not defend against an ungated publication landing between the pin and the load.

The new test encodes the divergence rather than catching it: `expect(db.verifyVersion(key, latestVersion)).toBe(true)` at test/verification-table.test.ts:698 asserts the slot is published while `txn`'s behind-the-latest snapshot is still open (the `abort()` is in the `finally`). Run the same scenario through `getBinarySync` and Gate 2 suppresses publication, so `verifyVersion` returns false.

Suggested fix: run the gate in the worker, where the DB and CF are still pinned, and carry only the decision to the completion — e.g. have the worker compute `vtLatest` as it does now plus a `settled` bool from the same two gates (`rocksdb.oldest-snapshot-time` / `rocksdb.oldest-snapshot-sequence`), and have `resolveGetResult` publish only when `settled`. FRESH itself can still be answered from `vtLatest`; it's only the slot write that needs the gate. Then flip the test's expectation to `false` while the snapshot is open, which is what the sync sites already do.

—
_Generated by Barber AI_

test/verification-table.test.ts
```diff
@@ -256,6 +269,30 @@ describe('Verification Table', () => {
 				expect(db.verifyVersion(key, backdatedVersion)).toBe(true);
 			}));
 
+		it('suppresses async seeding while an older snapshot is open', () =>
```
cb1kenobi, line 289: **Low: nothing asserts the async paths still publish when the version *is* settled**

The new gate is a bool computed in the worker and defaulted to `false`, and both async publish sites in `resolveGetResult` are now `&& state->vtVersionSettled`. Every async assertion in this suite is negative — this test, `does not publish from an async transactional snapshot read while its snapshot is behind`, and the two `does not answer FRESH ... async` tests all end in `verifyVersion(...) === false`. Every positive seeding test (`a transactional read seeds the VT when the key is settled`, `seeds the slot when POPULATE_VERSION_FLAG is set`) goes through `getSync`/`getBinarySync`.

So if `vtVersionSettled` were left permanently false — an early `return` added to `vtCheckAsyncGet`, the gate evaluated against the wrong column family, or `vtLatest.read` flipping `populateVersion` to a `0` latest — async cache seeding would stop entirely and the suite would stay green. That is a silent caching regression, not a correctness one, but it is exactly the direction the current tests cannot see.

Suggested fix: add the async analogue of `a transactional read seeds the VT when the key is settled` — with no snapshot open, `await db.getBinary(key, { expectedVersion: version })` on a `noBlockCache` store (assert the result is a `Promise` so it really takes the async path), then `expect(db.verifyVersion(key, version)).toBe(true)`.

—
_Generated by Barber AI_

kriszyp: Addressed in 6f66027c: added the positive async analogue with `noBlockCache`, asserted `getBinary()` returns a Promise, confirmed `FRESH_VERSION_FLAG`, and verified the settled version was published to the VT slot. The focused suite and full Vitest suite pass. — GPT-5 Codex

cb1kenobi:
Reviewed `6f66027` for the incremental update from `f03f044` — no issues found.

The added `noBlockCache`/Promise assertion covers the async completion path, and the final `verifyVersion` assertion catches the silent regression where `vtVersionSettled` never permits async VT publication.

_Generated by Barber AI_

PR #767, opened 2026-08-08
Add storage tiering: SST paths and blob file placement across volumes

kriszyp:
Adds database-wide SST placement through `paths`, per-column-family blob placement through `blobs.dir`, and a separate process-wide blob cache for large values. The final validation pass [fails closed around persisted blob directories and derives directory creation permission from the same relocation decision](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1634), allows a harmless leftover acknowledgement on a read-only no-op, and makes [warm `paths` reuse distinguish omission from an explicit empty list](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R473): omission inherits and `[]` [rejects a tiered live database](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R502), while the same `[]` [matches an already-untiered one](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R515).

Directory isolation remains caller-managed by prior ruling: [two databases must never share a `paths` or `blobs.dir` directory](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-a7eebabb761b90e9e1c65f2a8e2e451f1c42b217ecaae8d053615fdd4c98450fR109). The branch keeps the conflict-free `origin/main` merge requested by @kriszyp; no history was rewritten.

One unrelated but blocking fix rode along: CI's Node/Bun/Deno test lanes went uniformly red after the rebase, not from this branch's own code — pnpm 12 (installed via this repo's `version: latest` action-setup pin) changed `pnpm rebuild` to no longer fall through to a same-named `package.json` script the way pnpm 11 did, so [`pnpm build`'s bare `pnpm rebuild`](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R38) silently stopped compiling the native addon in CI and every lane fell back to whatever stale prebuild happened to be in `node_modules` (missing e.g. `catchUpWithPrimary`, the new drop-latch test hooks). Confirmed by reproducing locally: `pnpm@12 run build` from a clean `build/` produces a real compile with `pnpm rebuild`, silently produces none. [`pnpm run rebuild`](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R38) is unambiguous on every pnpm version; the same fix landed in the [ASan benchmark workflow](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-b423d7ba07b5148425388ab8be38b573d0b2969e954b8fd63f9108c1835f455bR70) and the [developer-facing docs](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R16) that tell someone to type the command directly.

A successful non-default column-family drop now [removes that family from every live and retained destroy layout](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R394) before [unregistering its name](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1923). This prevents a later `destroy()` through an old handle from deleting a replacement database that reused the dropped family's blob directory, while preserving same-name recreation and directories still shared by another family; [five regressions cover synchronous, asynchronous, live-descriptor, stale-handle, and shared-directory cases](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R1311).

That drop is also atomic against `DBRegistry::OpenDB`. It [holds `databasesMutex` across `DB::DropColumnFamily` itself, both layout erasures, and the `columns` erase](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R354) — the same mutex a warm open takes before `columnsMutex` — so an open sees the family wholly present or wholly gone rather than being handed a `ColumnFamilyDescriptor` for a family RocksDB has already dropped. The verification-table sweep deliberately [stays outside that section](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R511): the VT writer mutex is process-global and must never be taken under the registry mutex. A [test-only two-phase latch](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R370) parks the drop mid-section until a worker's warm open has reached the registry mutex, keyed by database path because the counters are process-global and Vitest's thread pool shares them, so the [regression asserts the interleaving happened](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-522bb7de846f02893e17e7a83b8ece2b98b60f4da4deae466546af5b8146b5ceR314) rather than hoping for it.

The retained destroy layout is now canonical rather than last-writer-wins. Registry entries are keyed by path AND read-only, so one database can have two live descriptors, and a read-only handle opens with a `paths` shorter than the writer's — or none — whenever the files it needs still sit at path index 0. `db_paths` is serialized nowhere, so recording that shorter list erased the only trace of the writer's other volumes and `destroy()` reported success with their SST files still on disk (reported by @cb1kenobi). [`RecordLayout` now takes a new list only when it EXTENDS the retained one](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R359), and [`DestroyDB` applies the record first and the live descriptor second](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R248), so the entry the registry loop happens to pick can add volumes but never remove them. Both halves of that rule are load-bearing in opposite directions, and [invariant 18 carries the contract](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R696): a shorter list must not shorten the record, and a divergent one must not extend it either — `destroy()` deletes every SST it finds in each recorded directory, so one mistyped `paths` naming a neighbouring database's volume would otherwise take that database's files down with this one. [Three regressions](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R471) cover the reader that omits `paths`, the reader that passes a shorter list and stays live across `destroy()`, and [a reader naming a neighbour's volume](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R520) — each mutation-checked against the rule it exists for.

Keeping that record is only half of it, and @cb1kenobi found the other half: a writable open the record refused still succeeded, warned about in the info LOG and otherwise ignored. RocksDB then used the requested list while `destroy()` used the retained one — compaction writing SSTs to a volume `destroy()` never sweeps, and `destroy()` still sweeping one this open disowned, which another database may since have been given. Neither is repairable afterwards, so [such an open now fails](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R351), and it fails [before `DB::Open`](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1543) — past that point a single compaction placing a single SST is already the divergence. [`RecordLayout` throws on the same condition](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R391) rather than returning a bool nobody acted on, now purely as the assertion that the record did not move underneath an open database.

The dangerous shape is the quiet one: while every file still sits at path index 0, RocksDB opens happily under a shorter or divergent list, so [the regression uses exactly that setup](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R608) and asserts that no new OPTIONS file appears — RocksDB writes one on every open, so an unchanged set is what proves the refusal preceded the open rather than being unwound after one. Removing the pre-open call makes that assertion fail. One consequence: the check preempts `explainOpenFailure`'s removed/reordered guidance whenever the same process opened the database earlier, so coverage for that message moved to [a spawned child](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-d19fd7eb428ce10baa83a87acaf084fc5d5ed8b230558abede87f4ed9a576469R1), where the record is gone and RocksDB reports the MANIFEST as corrupt — which is what a restart actually produces.

## For the human reviewer

1. **Require patched-build coverage before merge.** The selected policy is to wait for a prebuild carrying HarperFast/rocksdb-prebuilds#13, then run the `ROCKSDB_JS_REQUIRE_BLOB_DIR=1` suite and specifically confirm external-blob cleanup through `DestroyDB`; the alternative is merging native storage code that is compiled out by every current prebuild. Waiting is reversible as soon as the prebuild exists, while saying no accepts an unexecuted data-integrity path. One change on this branch sits entirely inside that gap: `destroy()` now prefers the retained record's blob directory over a live descriptor's, so a read-only handle opened before a relocation can no longer make the sweep visit the directory the blob files left. It is argued from the ordering rather than demonstrated, because every case that would exercise it is capability-skipped.
2. **Keep `allowDirChange` database-scoped even though it is nested under per-family `blobs`.** A relocation moves every family sharing the source directory, so [the decision deliberately reaches those families together](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-85ce2e3dbb3be9c557f8fc54e1f2e80929eb32da486dcffbd9f32306f3508a8fR70); the alternative is splitting or renaming the acknowledgement before release. Changing this later is breaking, while saying no now costs an API redesign and its migration tests.
3. **Retain tier layouts in the process registry, not a persistent marker.** This makes `close()` followed by `destroy()` work because [the registry keeps a path-keyed layout copy](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R314), but a fresh process cannot recover `db_paths` and may orphan external SSTs. A rocksdb-js-owned marker would cover fresh-process destroy and directory identity at the cost of a new durable format; adding one later is possible but cannot retroactively describe existing tiered databases. Independent review raised the same gap from its other end this round — a process whose first open is read-only seeds an EMPTY path record, so `close()` + `destroy()` there reports success with external SSTs still on disk. Declined as the same open question rather than a separate defect: the empty record suppresses nothing, because the fallback it bypasses is `LoadLatestOptions`, and `db_paths` sits in RocksDB's "not yet supported" serialization block, so that path recovers exactly as little. Trusting a reader's list instead is the cross-database deletion the retained-record rule exists to prevent. Only a durable marker closes it.
4. **Leave the read-only descriptor's column entry registered on a drop.** Independent review raised this three rounds running as a leak: a warm read-only descriptor keeps serving the dropped family. It is a separate `DB::OpenForReadOnly` instance whose handles a writable drop does not invalidate and which by design never observes later changes, so unregistering there would make the next read-only open throw "cannot create column family in read-only mode" for a family its own frozen manifest view still contains; the [rationale is now recorded at the call site](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R420). The alternative is treating a read-only view as invalidated by a writable drop, which is a broader semantic change than this PR should make.
5. **Keep omitted `blobs.dir` strict on warm opens.** Omitting it means “alongside the SSTs,” so a plain second open of a family already using an external directory rejects even though the live handle cannot move. The alternative is an explicitness latch limited to warm reuse; loosening this later is compatible, while saying no now costs that additional state and test surface.

## Verification

- `pnpm build` — production TypeScript bundle and native binding built successfully.
- `pnpm check` — type-check, lint, and formatting checks passed for 191 files.
- `pnpm test` — 864 passed, 31 skipped, 0 failed on the pushed head; `pnpm check` (type-check, lint, format over 192 files) passed.
- `pnpm test:native` — 197/197 passed; the focused `BlobRelocation`/`BlobDirScanner` run passed 31/31, including [destination-creation ownership and no-op acknowledgement cases](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-7c8a747fed3eb2f3e59d0138492b5bce67e88ba3197609df7e2a6058ed50011cR271).
- The three destroy-layout regressions were mutation-checked one rule at a time: without the `RecordLayout` guard the omitted-`paths` reader orphans the writer's second volume, without the `DestroyDB` merge the shorter-list reader does, and with either the size-only rule or the previous live-first ordering the neighbour's SST files are deleted.
- The drop/warm-open regression was mutation-checked: restoring the two-lock split makes the worker receive the dropped family and its write is silently discarded, and a second latch use in the same process still works from the baselined counters.
- `nm -C -u build/Release/rocksdb-js.node | rg 'rocksdb_js::'` and `git diff --check` produced no output.
- On the `origin/main` merge (one conflict, the include block in `db_descriptor.cpp`, both sides kept): `pnpm build`, `pnpm test` (875 passed / 32 skipped, 65 files), `pnpm test:native` (209/209), `pnpm check`, and `nm -C -u` clean of undefined `rocksdb_js::` symbols — the check that caught the last rebase silently dropping main's `ParkTimeoutRegistry` hunks.
- The pre-open refusal is mutation-checked in the direction that matters: disabling only `AssertDbPathsExtendRetained` leaves the post-open throw in place, the message assertion still passes, and the test fails on the OPTIONS-file assertion (`OPTIONS-000007, OPTIONS-000015` vs `OPTIONS-000007`) — i.e. the open had happened.
- The new spawned fixture was run under the repo's actual Deno command (`deno run --allow-all --sloppy-imports` on Deno 2.7.5): both storage-path regressions pass. Independent review flagged it as needing its own permission flags; measured against `pnpm test:deno` rather than a bare `deno run`, the child inherits, as the existing `blob-cache` fixture already relies on.
- The 28-case [`blobs.dir` integration suite remains capability-skipped](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R735) because the current RocksDB prebuild lacks the downstream patch; this includes the new [read-only relocation/no-op distinction](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R1043) and [dropped-layout regressions](https://github.com/HarperFast/rocksdb-js/pull/767/changes?w=1#diff-c85251cd2a05f72f780896f207b1ac34e1c14110bf733dadfc32bd0b48badbc7R1311).
- The pnpm-rebuild fix was verified against the actual failure: with `build/` removed, `npx pnpm@12 rebuild` (bare, pre-fix spelling) leaves `build/Release` empty; `npx pnpm@12 run build` (post-fix) produces a real `CXX`/`SOLINK_MODULE` compile and a `build/Release/rocksdb-js.node` exporting `catchUpWithPrimary` and the drop-latch test hooks. The full suite then passes clean: `pnpm test` — 937 passed, 38 skipped, 0 failed.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=gemini,codex; adjudicated=domain; declined=cursor-grok,cursor-composer; rounds=31 @ 8ed73dababec</sub>

<sub>Human-Review-Need: 3 (decisions: blob-cache-derived-budget, allow-dir-change-reaches-other-families, flatten-means-omitted-dir, db-paths-retained-in-process-only, destroy-sweeps-shared-dirs-unchecked, drop-serialized-under-registry-mutex, test-latch-in-production-binary) @ 8ed73dababec</sub>


src/binding/database/db_registry.cpp
```diff
@@ -345,6 +345,24 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
 			}
 		}
 
+		// db_paths is fixed for the life of the open database, so a second open
+		// asking for different volumes cannot take effect on the reused handle.
+		// Reject rather than let the caller believe SST files are being tiered.
+		if (!options.paths.empty()) {
+			const auto& currentPaths = entry.descriptor->db->GetDBOptions().db_paths;
```
cb1kenobi: **High: `destroy()` leaves every SST on a `paths` volume (and every blob file in `blobs.dir`) on disk**

`DBRegistry::DestroyDB` (line 229 of this file) calls `rocksdb::DestroyDB(path, rocksdb::Options())`. A default `Options` has an empty `db_paths` and an empty `blob_dir`, so RocksDB only scans the database directory; the following `std::filesystem::remove_all(path)` does the same. Before this PR that was complete, because everything lived under `path`. It no longer is.

Reproduced on this branch (macOS, unpatched prebuild, so `paths` only — `blobs.dir` follows by the same mechanism since the patched `DestroyDB` collects `blob_dir` from the `Options` it is handed):

```
sst files on tiered volume before destroy(): 1
sst files on tiered volume after  destroy(): 1  ['000009.sst']
database directory still exists: false
```

Control without `paths`: 1 -> 0. So `db.destroy()` now silently orphans the bulk of a tiered database — exactly the volume the user moved the data to because it was large.

Suggested fix: `DestroyDB` already holds the live descriptor before it resets it. Capture the real layout there and pass it down instead of a default `Options` — `destroyOptions.db_paths = descriptor->db->GetDBOptions().db_paths;` plus the per-CF `blob_dir` from `GetOptions(cf)` — so RocksDB deletes the files it actually created.

—
_Generated by Barber AI_

src/binding/database/db_descriptor.cpp
```diff
@@ -104,9 +108,23 @@ rocksdb::ColumnFamilyOptions buildColumnFamilyOptions(
 		tableOptions.block_cache = DBSettings::getInstance().getBlockCache();
 	}
 
-	cfOptions.enable_blob_files = true;
-	cfOptions.min_blob_size = 2048;
-	cfOptions.enable_blob_garbage_collection = true;
+	cfOptions.enable_blob_files = options.blobs.enabled;
+	cfOptions.min_blob_size = options.blobs.minSize;
```
cb1kenobi: **High: a cold open restamps every other column family's blob settings**

These five per-CF blob options are taken from the opening handle and applied to *every* column family in `cfDescriptors`, and `loadPersistedCFOptions` restores only `compression`/`blobCompression`/`compressionOpts` (and `blobDir`, for comparison). RocksDB does not restore per-CF options on its own — which is precisely why the comment block above the cold-open loop says opening one column family must not change another's algorithm. Blob placement now violates that rule.

Before this PR the three blob fields were hardcoded constants, so every CF got identical values and divergence was impossible. Now it is silent. Reproduced on this branch:

```
// two named CFs, both created with the default minSize 2048
open(dbPath, { name: 't1', blobs: { minSize: 1<<20 } })   // cold open
open(dbPath, { name: 't2' })                              // warm
// write 30 x 8KB values to t2, flush
blob files before/after: 2 -> 2      // t2 stopped extracting blobs
control (t1 opened with no blobs option): 2 -> 3
```

Harper maps every table to a named column family, so which table happens to open the database first decides `min_blob_size`/`enable_blob_files`/GC for all of them, and it flips on restart. `enabled: false` on one table turns blob separation off for the whole database.

Suggested fix: mirror the compression handling — carry these fields in `PersistedCFOptions`, restore them for non-target families, and apply the request only to `options.name` (with an explicit all-families opt-in if that is wanted, as `compressionForAllColumnFamilies` does). Anything less makes `blobs` a database-wide option in practice while presenting as per-open.

—
_Generated by Barber AI_

src/binding/database/db_descriptor.cpp
```diff
@@ -1027,6 +1050,14 @@ std::shared_ptr<DBDescriptor> DBDescriptor::open(const std::string& path, const
 		? deriveMaxOpenFiles(getEffectiveOpenFileLimit())
 		: options.maxOpenFiles;
 	dbOptions.keep_log_file_num = 5; // these are informational log files that clutter up the database directory
+	// Spread SST files across volumes. Left on `db_paths` rather than `cf_paths`
+	// so one tiering policy covers every column family in the database (RocksDB
+	// falls back to db_paths when cf_paths is empty). A file's path index is
+	// recorded in the MANIFEST, so entries may only ever be appended — see the
+	// note on DBOptions::paths.
+	for (const auto& storagePath : options.paths) {
+		dbOptions.db_paths.emplace_back(storagePath.path, storagePath.targetSize);
```
cb1kenobi, line 1566: **High: adding `paths` to an existing untiered database makes it unopenable, with a corruption error**

When `db_paths` is empty RocksDB sanitizes it to `[{dbname, max}]`, so existing SST files carry path index 0 meaning *the database directory*. Supplying `paths` for the first time redefines index 0 as `paths[0]`, and every existing file is looked for on the wrong volume. Reproduced on this branch:

```
open(dbPath)                       // 500 keys, flush, close  -> 1 sst in dbPath
open(dbPath, { paths: [{ path: fast, targetSize: 1<<30 }] })
=> Corruption: IO error: No such file or directory: .../000009.sst
   The file .../MANIFEST-000005 may be corrupted.
```

Two problems. The database is bricked by what reads like a supported configuration change, and the error blames the MANIFEST — an operator seeing that will reach for backup restore rather than reverting one config line. `docs/tiered-storage.md` frames the append-only rule as "existing files keep path index 0 and stay put", which is true only when a list already exists; the zero-to-one transition is the migration people will actually attempt, and it is neither documented nor tested (`should still read data written before a path was appended` starts from a one-entry list).

Suggested fix: document that adding tiering to an existing database requires the database directory itself as `paths[0]` (`paths: [{ path: dbPath, targetSize: ... }, { path: slow, ... }]`), and say so in the README option description too. A cheap guard is also available here: if `paths` is non-empty and the database already exists (the `ListColumnFamilies` result just below tells you), and `paths[0].path` is not the database directory, consider failing with a message that names the real cause instead of letting RocksDB report corruption.

—
_Generated by Barber AI_

src/binding/database/database.cpp
```diff
@@ -1794,6 +1794,174 @@ napi_value Database::ListLogs(napi_env env, napi_callback_info info) {
...
+		const std::string targetSizeLabel = "paths[" + std::to_string(i) + "].targetSize";
+		if (!getByteSizeProperty(
+				env, entry, "targetSize", storagePath.targetSize, true, targetSizeLabel.c_str())
+		) {
+			return false;
+		}
+		// Resolve now: RocksDB interprets a relative path against the process
+		// working directory at open time, so storing it raw would let the same
+		// string resolve to a different volume on a later open.
+		storagePath.path = std::filesystem::absolute(storagePath.path).lexically_normal().string();
```
cb1kenobi: **Medium: `std::filesystem::absolute` can abort the process here (static analysis, not reproduced)**

`parseStoragePaths` and `parseBlobOptions` are called at line 2168, *outside* the `try`/`catch` that wraps `(*dbHandle)->open(...)` at line 2184. The throwing overload of `std::filesystem::absolute` raises `std::filesystem_error` (it consults the process working directory for a relative path, and fails on Windows for an unresolvable one). An exception escaping an N-API callback is `std::terminate`, not a JS throw — a malformed config string takes the whole process down instead of rejecting the open.

Suggested fix: use the `error_code` overload and convert the failure into a JS error like the other validation in these two functions:

```cpp
std::error_code ec;
auto resolved = std::filesystem::absolute(storagePath.path, ec);
if (ec) {
	::napi_throw_error(env, nullptr,
		("paths[" + std::to_string(i) + "].path could not be resolved: " + ec.message()).c_str());
	return false;
}
storagePath.path = resolved.lexically_normal().string();
```

The same applies to the `blobs.dir` call a few lines below.

—
_Generated by Barber AI_

src/binding/database/db_settings.cpp
```diff
@@ -151,6 +178,50 @@ napi_value DBSettings::Config(napi_env env, napi_callback_info info) {
...
+				env,
+				nullptr,
+				"Blob cache size must be a non-negative integer no larger than Number.MAX_SAFE_INTEGER"
+			);
+			return nullptr;
+		}
+		configuredBlobCacheSize = static_cast<size_t>(requested);
+		updateBlobCacheSize = true;
+	} else if (blockCacheSizeProvided) {
+		configuredBlobCacheSize = static_cast<size_t>(blockCacheSize) / 10;
```
cb1kenobi: **Medium: every existing `config({ blockCacheSize })` caller silently gains a blob cache**

Two behaviours fall out of this branch that existing callers did not ask for.

1. The blob cache is *additional* capacity, not carved out of the block cache. A deployment that today calls `config({ blockCacheSize: 8 * 1024 ** 3 })` because that is its memory budget now has a ceiling of 8.8 GB. Nothing in the call changed, so this lands on upgrade, not on opt-in.
2. It is unconditional on `hasBlobCacheSize` in *this* call, so a later block-cache-only call discards an explicitly configured blob cache. `config({ blobCacheSize: 8MB }); config({ blockCacheSize: 10MB })` ends at 1 MB — the added test locks that in, so it is deliberate, but it means explicit intent has no way to survive an unrelated later call.

Suggested fix: latch explicit intent — record whether `blobCacheSize` was ever set explicitly and skip the derivation once it was. If the derived-by-default behaviour is kept, the README entry should say plainly that upgrading raises the process memory ceiling by 10% of `blockCacheSize` with no code change; right now that has to be inferred from "additional to the block-cache capacity".

—
_Generated by Barber AI_

test/tiered-storage.test.ts
```diff
@@ -0,0 +1,448 @@
...
+}
+
+/** A value large enough to land in a blob file at the default 2 KB threshold. */
+function largeValue(seed: number): string {
+	return `${seed}:`.padEnd(8192, 'x');
+}
+
+afterEach(() => {
+	while (tempPaths.length) {
+		rmSync(tempPaths.pop()!, { recursive: true, force: true });
```
cb1kenobi, line 107: **Low: a failed assertion leaves a database open while `afterEach` deletes its directory**

Every test here ends with a bare `db.close()` after the `expect`s. When an assertion fails the close never runs, and this hook then `rmSync`s the database directory, the `paths` volumes and the blob directory out from under a live RocksDB with background compaction running. In a repo that already routes background errors through a native listener, that turns one clean assertion failure into follow-on noise or a worker crash, and the real failure gets harder to read.

Suggested fix: open in a `try`/`finally` with `db.close()` in the `finally` (the `should reject a warm reopen` test already does this), or track the open databases in a module-level array and close them in this hook before the `rmSync` loop.

—
_Generated by Barber AI_

src/binding/database/db_registry.cpp
```diff
@@ -174,6 +206,32 @@ void DBRegistry::DestroyDB(const std::string& path) {
 		}
 	}
 
+	// Nothing live to read the layout from. `blob_dir` is persisted per column
+	// family, so it can still be recovered; `db_paths` is not written to the
+	// OPTIONS file at all (RocksDB serializes it in its "not yet supported"
+	// block), so a `paths` database that is not open in this process cannot have
+	// its tiered SST files removed. `Database::Destroy` requires an open handle,
+	// so that is not reachable from the public API.
```
cb1kenobi: **High: `close()` then `destroy()` still orphans every SST on a `paths` volume**

The comment above says `Database::Destroy` requires an open handle "so that is not reachable from the public API", but `destroy()` only requires a live *handle*, not a live *descriptor*. Once the last handle closes, the registry entry is gone, `capturedLayout` stays false, and this fallback cannot recover `db_paths` — so `rocksdb::DestroyDB` runs with an empty `db_paths` and the tiered SST files survive while the database directory is removed. `destroy()` reports success.

Measured on this head (local build, RocksDB 11.8.1, darwin/arm64), one SST on the tiered volume in each case:

```
destroy while open ......... tier sst after destroy = 0   (dir removed)  ok
close() then destroy() ..... tier sst after destroy = 1   ORPHANED
second handle still open ... tier sst after destroy = 0   ok
control, untiered .......... clean
```

This is the same data-orphaning the `layoutDbPaths` snapshot was added to fix, reached by an ordinary call order. `test/tiered-storage.test.ts` only covers `destroy()` on an open database, so nothing catches it.

Suggested fix: keep the layout snapshot reachable after the descriptor leaves the registry — either have `Database::Destroy` capture `layoutDbPaths` from its own `dbHandle`'s descriptor and pass it to `DBRegistry::DestroyDB`, or refuse `destroy()` on a closed handle rather than silently doing a partial job. Add a `close(); destroy();` case to the tiered `destroy()` test either way.

—
_Generated by Barber AI_

src/binding/database/db_descriptor.cpp
```diff
@@ -995,6 +1159,105 @@ void DBDescriptor::lockReleaseByOwner(DBHandle* owner) {
...
+ * open would break. An already-tiered database has no SST files in its own
+ * directory, and one that lists its own directory as `paths[0]` finds them
+ * there, so both pass — including when the two spellings differ (trailing
+ * separator, symlink).
+ *
+ * This does NOT catch reordering or removing an entry of an existing list; the
+ * files are elsewhere and nothing on disk records which index they came from.
+ * That stays a documented append-only rule (see `DBOptions::paths`).
+ */
+static void assertStoragePathsUsable(
```
cb1kenobi, line 1412: **Medium: removing `paths` bricks the database with the same error this guard exists to preempt**

`assertStoragePathsUsable` catches the zero-to-one transition, but the one-to-zero transition — deleting the `paths` line from a config, no code change — is unguarded and produces the identical failure. Reordering an existing list does too. Reproduced on this head:

```
tiered db (1 sst in tier volume, 0 in db dir)
  drop `paths`  -> Corruption: IO error: No such file or directory:
                   .../db/000009.sst ... MANIFEST-... may be corrupted
  reorder list  -> Corruption: ... .../tier2/000015.sst ...
  restore paths -> opens fine, key-0 = value-0
```

So it is recoverable rather than data loss, but the message names the MANIFEST, which is exactly the "sends an operator to backup restore rather than to the one config line they changed" outcome the doc comment above calls out. The PR body raises this as open item 3; flagging it so it gets an explicit ruling rather than shipping by default.

Suggested fix: nothing on disk records the old list, so it cannot be detected up front — but it can be translated. Catch a `Corruption` from `DB::Open` whose message names a missing `.sst` and, when `options.paths` is empty or shorter than what the files need, rethrow with "if this database was previously opened with `paths`, that list must be supplied again and may only be appended to". That turns the same brick into a one-line config fix.

—
_Generated by Barber AI_

src/binding/database/db_descriptor.cpp
```diff
@@ -104,30 +194,93 @@ rocksdb::ColumnFamilyOptions buildColumnFamilyOptions(
...
+	cfOptions.enable_blob_garbage_collection =
+		options.blobs.garbageCollection.value_or(kDefaultBlobGarbageCollection);
+	cfOptions.blob_garbage_collection_age_cutoff =
+		options.blobs.garbageCollectionAgeCutoff.value_or(kDefaultBlobGarbageCollectionAgeCutoff);
+	cfOptions.blob_garbage_collection_force_threshold =
+		options.blobs.garbageCollectionForceThreshold.value_or(
+			kDefaultBlobGarbageCollectionForceThreshold);
+	if (options.blobs.prepopulateCache.value_or(kDefaultBlobPrepopulateCache)) {
+		cfOptions.prepopulate_blob_cache = rocksdb::PrepopulateBlobCache::kFlushOnly;
+	}
```
cb1kenobi: **Medium: `prepopulate_blob_cache` leaks from the database's first opener into every warm-created column family**

Every other blob field here assigns unconditionally via `value_or(...)`, so a column family created on an already-open database gets the documented creation default. This one only assigns in the `true` branch, and the warm path passes the descriptor's own `cfOptions` as the base (`db_registry.cpp:447`, `buildColumnFamilyOptions(options, entry.descriptor->cfOptions)`) — which carries `kFlushOnly` when the database's *first* opener asked for prepopulation. The new family inherits it, and it is persisted to the OPTIONS file, so it survives restarts.

Measured on this head: database opened with `blobs: { prepopulateCache: true, minSize: 1MB, garbageCollectionAgeCutoff: 0.9 }`, then a family `warm-cf` created with no `blobs` at all:

```
warm-cf min_blob_size                      = 2048       (default cf = 1048576)  ok
warm-cf blob_garbage_collection_age_cutoff = 0.250000   (default cf = 0.900000) ok
warm-cf prepopulate_blob_cache             = kFlushOnly (default cf = kFlushOnly) LEAKED
```

In the Harper shape where every table is a column family, a table that never opted in warms the shared blob cache on every flush. Same class as the restamping this PR fixed, and it contradicts `applyBlobCreationDefaults`, which sets `kDisable` explicitly on the cold-open equivalent.

Suggested fix:
```suggestion
	cfOptions.prepopulate_blob_cache =
		options.blobs.prepopulateCache.value_or(kDefaultBlobPrepopulateCache)
			? rocksdb::PrepopulateBlobCache::kFlushOnly
			: rocksdb::PrepopulateBlobCache::kDisable;
```

—
_Generated by Barber AI_

src/binding/database/db_settings.cpp
```diff
@@ -138,19 +165,95 @@ napi_value DBSettings::Config(napi_env env, napi_callback_info info) {
...
+		std::lock_guard<std::mutex> lock(settings.blobCacheMutex);
+		bool update = blobCacheSizeProvided;
+		size_t configuredBlobCacheSize = requestedBlobCacheSize;
+		if (blobCacheSizeProvided) {
+			settings.blobCacheSizeExplicit.store(true, std::memory_order_relaxed);
+		} else if (!settings.blobCacheSizeExplicit.load(std::memory_order_relaxed)) {
+			// Derive only while the caller has never stated a budget: an unrelated
+			// later `config({ blockCacheSize })` says nothing about blobs and must
+			// not discard one.
+			configuredBlobCacheSize = static_cast<size_t>(blockCacheSize) / 10;
```
cb1kenobi, line 244: **Medium: the derived blob cache is still additional capacity, so the memory ceiling still rises on upgrade**

The `blobCacheSizeExplicit` latch fixes the second half of the earlier finding — a later block-cache-only `config()` no longer discards a stated budget, and `test/blob-cache.test.ts` covers it. The first half is unchanged and now documented rather than resolved: a process that has never set `blobCacheSize` and calls `config({ blockCacheSize })` — every existing caller — gains `blockCacheSize / 10` of capacity **on top of** the block cache. Confirmed on this head via `test/fixtures/blob-cache-derivation.mts`: 10 MB block cache yields a 1 MB blob cache, total 11 MB.

That is a 10% process memory-ceiling increase that lands on upgrade with no code change, in a library whose callers size the block cache against a container limit. Raising it as a ruling, not a defect — the PR body lists it as deliberate.

Suggested fix: if the 10% default is worth keeping, carve it out of `blockCacheSize` rather than adding to it (block cache gets 90%, blob cache 10%), so the documented budget stays the total. If it must be additional, default it to `0` and let callers opt in — the derivation buys nothing they cannot ask for.

—
_Generated by Barber AI_

test/tiered-storage.test.ts
```diff
@@ -0,0 +1,947 @@
...
+		expect(() =>
+			RocksDatabase.open(tempPath(), { blobs: { garbageCollectionForceThreshold: -0.1 } })
+		).toThrow(/garbageCollectionForceThreshold must be between 0.0 and 1.0/);
+	});
+
+	// The blob CACHE tests live in test/blob-cache.test.ts: the "explicitly set"
+	// flag they exercise is process-global and latched, so they need a process
+	// that has never configured one.
+
+	it('should keep each column family on its own blob settings', () => {
```
cb1kenobi, line 849: **Low: this regression test passes with the fix removed**

`t2` is created *after* `t1`'s cold open, so when `t1` opens, `t2` is not in the OPTIONS file and there is nothing for `restorePersistedBlobOptions` to restore — `t2` then picks up its settings as a freshly created family. The restamping this test is named for needs `t2` to already exist on disk.

Verified by mutation: deleting the `restorePersistedBlobOptions(cfo, it->second);` call from `db_descriptor.cpp` (marker string confirmed in `db_descriptor.o` and `rocksdb-js.node`) fails two other cases in this file, but **this one still passes**. With `t2` seeded first, the mutant stops extracting 8 KB values into blob files (1 blob file before, 1 after; control 1 -> 2).

Suggested fix: create and close `t2` with the default threshold before the cold open of `t1`, then reopen `t2` and assert it still extracts.

—
_Generated by Barber AI_

kriszyp:
Thanks — four of the five were real and are fixed at d725c32a; the fifth is in the PR body for a ruling.

- **destroy() after close()** — fixed, and the root cause turned out to be wider than a missing capture: any per-handle snapshot goes stale the moment another open appends a storage path. The process now remembers each database's layout by path (`DBRegistry::knownLayouts`), so `close()` → reopen with `[A, B]` → close → `destroy()` on the first handle deletes both volumes. Covered by two cases, both mutation-checked.
- **prepopulate_blob_cache leak** — fixed as suggested; `blob_cache` was inheriting the same way past a `noBlockCache` request and got the same treatment.
- **The regression test passing with the fix removed** — confirmed by mutation, and seeding `t2` first was not enough on its own: the later plain reopen asks for the 2 KB default anyway and writes the correct value back, so it still passed. It now asserts against the OPTIONS file, which is where the restamp actually lands.
- **Removing/reordering `paths`** — `explainOpenFailure` now appends your suggested guidance to any `Corruption` naming a `.sst`. Appended and conditional rather than substituted, since genuine corruption gives the same status.
- **The derived blob cache** — left for @kriszyp with the two alternatives (carve it out of `blockCacheSize`, or default to 0 and require opt-in) written up in the PR body.

Chasing the flat-family case in your third comment also turned up a bigger one: `allowDirChange` skipped its own "checked, not trusted" guard whenever the persisted `blob_dir` was empty — which is every family of an untiered database, so the flat-to-external move, the only migration those databases can make, was the one form taken on faith. Both sides now resolve an empty directory to the database directory before comparing.

— Claude Opus

src/binding/database/db_settings.cpp
kriszyp: Ruled by @kriszyp: the derivation stays **additive** as it is. Neither alternative was taken — not carving the 10% out of `blockCacheSize` (which would silently shrink every existing caller's block cache), and not defaulting to `0` (which would leave blob values uncached, the thing the cache exists to fix on a tiered database).

So this ships as a 10% process memory-ceiling increase for any caller that sets `blockCacheSize` and never sets `blobCacheSize`. It is documented as such in three places — the README `blobCacheSize` entry, the blob-cache section of `docs/tiered-storage.md`, and the option's own doc comment — each stating explicitly that the capacity is additional and that setting `blobCacheSize` (including to `0`) latches and stops the derivation. It is also carried in the PR body under the deliberate decisions, flagged as needing a release-note line in whatever version carries it, since the ceiling change is what upgraders will actually feel.

The `blobCacheSizeExplicit` latch you confirmed in `test/blob-cache.test.ts` is unchanged.

— Claude Opus

cb1kenobi:
Re-reviewed `d725c32ab9882d7665039f183b336543dd9b0068` (round 12). Every prior finding was re-measured on a local build rather than read for the absence of the old code — **0 Critical, 0 High, 1 Medium open.**

### Confirmed fixed, by measurement

**High — `close()` then `destroy()` orphaned every tiered SST.** All three sequences now clean up, where the middle one was the bug:

| sequence | round 11 | `d725c32a` |
|---|---|---|
| destroy-while-open | 1 → 0 | 1 → 0 |
| close-then-destroy | 1 → **1** | **1 → 0** |
| second-handle-open | 1 → 0 | 1 → 0 |

Proved load-bearing rather than incidental: disabling only the `knownLayouts` fallback in `DestroyDB` and rebuilding reproduces the old result exactly (close-then-destroy back to 1 → 1, the other two unaffected). The cache lifetime holds up too — it survives `shutdown()` (1 → 0), and reopening a path with a different list overwrites it, so it cannot go stale. Reverting the `PurgeAll` clear in `a5fbf362` was the right call; the measurement agrees with the reasoning in that commit message.

**Medium — `prepopulate_blob_cache` leaking into warm-created families.** Fixed, and the new test at `tiered-storage.test.ts:620` is load-bearing: restoring the one-sided assignment fails it with `expected 'kFlushOnly' to be 'kDisable'`, i.e. the leak persisted into the OPTIONS file, reproduced exactly.

**Low — the regression test that passed with its own fix removed.** Now proves its claim. Making `RecordLayout` first-writer-wins (a per-handle layout) fails it with 8 orphaned SSTs on the appended volume.

**Medium — removing or reordering `paths`.** Removal and reorder still fail the open, which is inherent since `db_paths` is never persisted, but the corruption status now carries the explanation and the recovery, and restoring the list does recover the database. Verified all four cases directly. Appending the note rather than substituting it, and phrasing it conditionally, is the right call given a genuinely corrupt database produces the same status.

### Still open

**Medium — the derived blob cache is capacity *on top of* the block cache** (`db_settings.cpp:240`, already commented inline). Re-measured: `config({ blockCacheSize: 100 MB })` with `blobCacheSize` never set gives a 100 MB block cache **plus** a separate 10 MB blob cache — total ceiling 110 MB, +10% on upgrade with no code change. The `blobCacheSizeExplicit` latch is correct (an explicit `blobCacheSize: 0` leaves no blob cache attached), so only the never-set case is affected — which is every existing caller. `load-binding.ts:631-638` now documents this accurately, but the thread was resolved without a reply, so it reads as unanswered rather than decided. If the +10% is intended it is a release-note item and can be closed as accepted — worth saying so explicitly rather than leaving a silent resolve to stand in for the decision.

### Coverage is still the headline risk

Local run is green — 62 files, **846 passed / 22 skipped / 0 failed**, plus native **163 passed / 3 skipped** (15 `OptionsFile` cases). But skips have grown **9 → 19 → 21**, and all 21 are the `blobs.dir` block gated on the `blob_dir` patch. Of the seven tests this round adds, five run and two do not:

- skipped: `should refuse to relocate before the blob files have moved`
- skipped: `should refuse to relocate a flat family before its blob files have moved`

So the relocation guards added by `648678d2` / `1f17124a` / `9d94879e` / `cdcea6be` are **static-only on every build without the patch**, which is every build until [rocksdb-prebuilds#13 — *Add RocksDB blob_dir patch so blob files can live on a separate volume*](https://github.com/HarperFast/rocksdb-prebuilds/pull/13) lands. That PR is still open and is now waiting on @cb1kenobi. The `ROCKSDB_JS_REQUIRE_BLOB_DIR=1` gate is the right guard so a release cannot silently ship without the feature, but it does not make these tests run in the meantime — green CI is not evidence for the `blobs.dir` half.

—
_Generated by Barber AI_

kriszyp:
Addressed the round-12 executable-coverage gap in ccdbcc3e. The patched-only relocation decision now lives in Node-free `core/blob_relocation.cpp`, with its filesystem predicates injected, and the unpatched native test binary executes 18 cases covering the two skipped guards plus the full relocation blast radius, flattening, stale acknowledgements, and missing-directory failures. No relocation behavior changed. Local verification: `pnpm build`; `pnpm test:native` (184/184, including all 18 new cases); `pnpm test` (844 passed / 24 skipped); `pnpm check`.

src/binding/database/db_settings.cpp
cb1kenobi: Re-reviewed at `ccdbcc3e`. This thread shows resolved but nothing changed here: `db_settings.cpp` is untouched since the last review (confirmed via `git diff --stat d725c32a..ccdbcc3e -- src/binding/database/db_settings.cpp`, empty), and the 110MB ceiling still measures the same way — `configuredBlobCacheSize = blockCacheSize / 10` still runs whenever `blobCacheSizeExplicit` is false. `load-binding.ts:631-638` documents it as intentional, so this may be a deliberate call rather than an oversight — flagging so the resolution reflects a decision, not an accidental close.

—
_Generated by Barber AI_

src/binding/database/db_descriptor.cpp
```diff
@@ -576,178 +735,6 @@ void DBDescriptor::releaseCommitCompletionsByEnv(napi_env env) {
 	}
 }
```
cb1kenobi, line 782: **Critical: the rebase dropped main's `ParkTimeoutRegistry` implementation — the binding no longer loads on Linux/Windows and segfaults on `close()` on macOS**

This push rebased `da0ef135..ccdbcc3e` onto `06961b5f`. The conflict resolution deleted the whole `ParkTimeoutRegistry` implementation that main's #744 (`fd4e0017`) added to this file, but left the declaration in `db_descriptor.h` and the seven uses in `transaction/transaction.cpp` in place.

Measured, not inferred:

- `ParkTimeoutRegistry::` in `db_descriptor.cpp` — `06961b5f` (base): 12, `a2965c0b` (head): 0. `db_descriptor.h` still declares the class and `DBDescriptor` still holds `parkTimeouts = std::make_shared<ParkTimeoutRegistry>()`.
- Locally built `.node` at this head: `nm -u` reports 4 unresolved symbols — `~ParkTimeoutRegistry()`, `schedule(...)`, `fire(uint64_t)`, `releaseByEnv(napi_env)` — and 0 definitions.
- CI on this exact head (run `32880698422`, job `97909195169`): `undefined symbol: _ZN10rocksdb_js19ParkTimeoutRegistryD1Ev`. 15 jobs fail — every Linux and Windows job, Bun, Deno, benchmarks and stress.
- macOS links anyway (`-undefined dynamic_lookup`), so it fails later instead: a plain `open()` / `putSync()` / `close()` script exits **139 (SIGSEGV)** on the `close()`.
- Restoring the deleted block and rebuilding (`make BUILDTYPE=Release`) takes the unresolved count to 0, the same script to exit 0, and the full suite to `855 passed | 22 skipped`, 0 failed.

Two smaller things went with it in the same hunks and should come back together: `#include <system_error>`, and both `this->parkTimeouts->shutdown()` calls (the destructor safety net and the one in `finishClose()` that exists because a park can sit on a foreign-`dbId` tracker, so `cancelForDB()` alone does not wake it). The `finishClose()` comment recording why the waiting `flush()` default is deliberate was deleted too, though the `flush(bool allowWriteStall = false)` signature itself is intact after `a2965c0b`.

Suggested fix: re-apply main's `db_descriptor.cpp` content for those four hunks — `git checkout 06961b5f -- src/binding/database/db_descriptor.cpp` is not right (it would drop this PR's changes to the file), so restore just the deleted regions, then re-run CI.

Nothing else in the rebase regressed: the `range-diff` over separately-computed merge bases (`da0ef135..ccdbcc3e` vs `06961b5f..a2965c0b`) is otherwise prose and mechanical adaptation, and `transaction.cpp` has no net PR diff against the new base.

—
_Generated by Barber AI_

src/binding/database/db_settings.cpp
cb1kenobi: Re-reviewed at `a2965c0b` (rebase onto `06961b5f`). Still open, and this time re-measured rather than read off the constant.

A fresh process that has never set `blobCacheSize`, calling `config({ blockCacheSize: 100 MB })` and then opening a database, reports:

- `rocksdb.block-cache-capacity` = 100 MB
- `rocksdb.blob-cache-capacity` = 10 MB
- ceiling = **110 MB, +10.0%**

`db_settings.cpp` is byte-identical between `ccdbcc3e` and `a2965c0b`, and so is the `blobCacheSize` doc block in `load-binding.ts`, which describes the additive behaviour accurately. So this push changed neither the behaviour nor the docs — this remains a decision awaiting ratification (the release-note line the PR body asks for), not an oversight. The `blobCacheSizeExplicit` latch half is correct and verified.

Flagging the thread state again: this thread reads resolved and has no code change behind it on either push.

—
_Generated by Barber AI_

src/binding/database/db_descriptor.cpp
kriszyp: Fixed in `66cc2eee`, exactly as diagnosed. All four hunks restored from main: the `<system_error>` include, the `ParkTimeoutRegistry` method definitions, both `parkTimeouts->shutdown()` calls (destructor safety net + the `finishClose()` one for the foreign-`dbId` tracker case), and the `finishClose()` comment on the deliberate waiting `flush()` default. Applied as `git diff da0ef135 origin/main -- src/binding/database/db_descriptor.cpp | git apply -3`, so the restored regions are byte-identical to main and no tiered-storage change was touched.

Verified locally: `pnpm build` links clean, `pnpm test` 853 passed / 24 skipped, `pnpm test:native` 184/184, `pnpm check` clean — including `test/lock-tracker.test.ts`'s two bounded-park cases. I also re-audited the whole rebase: `git diff origin/main HEAD` now has no deletion anywhere that reverts a main-side change, so nothing else from `da0ef135..06961b5f` is missing.

— Claude Opus 5

cb1kenobi: Re-reviewed at `66cc2eeead76c18eac3c0e345e424ac1f911fa7c`. **Resolved — verified by measurement, not by thread state.**

**The restoration is faithful, not a re-implementation.** I extracted every top-level `ParkTimeoutRegistry::` definition from the merge base (`06961b5f`) and from this head and diffed them: 8 definitions on each side, **byte-identical**. Marker counts match main exactly — `ParkTimeoutRegistry` refs 12/12, `parkTimeouts->shutdown()` 2/2, `#include <system_error>` 1/1.

**Symbol-level, on a genuine local build.** `make BUILDTYPE=Release` (confirmed a real `CXX(target) db_descriptor.o` + `SOLINK_MODULE`, `.o` mtime advanced — not a `node-gyp` no-op). Per-object closure over all 22 objects: 548 undefined, 1696 defined, **0 unresolved `rocksdb_js` symbols**. The four that were unresolved before now all bind:

| symbol | referenced from | now defined in |
| --- | --- | --- |
| `ParkTimeoutRegistry::schedule` | `transaction.o` | `db_descriptor.o` |
| `ParkTimeoutRegistry::fire` | `transaction.o` | `db_descriptor.o` |
| `ParkTimeoutRegistry::releaseByEnv` | `db_registry.o` | `db_descriptor.o` |
| `DBRegistry::ReleaseParkTimeoutsByEnv` | `binding.o` | `db_registry.o` |

Rebuilding this same worktree with the `a2965c0b` version of `db_descriptor.cpp` reproduces **exactly 4** unresolved symbols and the crash: open + 500 puts + `close()` dies right after the read-back, **exit 139 (SIGSEGV)**. With the restored file the same script prints `CLOSED CLEANLY` and **exits 0**. Full suite at this head: 63 files, 855 passed, 22 skipped, 0 failed.

**Nothing else from main was dropped.** I reconstructed the mechanical result — `git merge-tree --write-tree` of the pre-rebase tip `d725c32a` with main — and diffed it against this head. `db_descriptor.cpp` **auto-merges cleanly**, so the deletion was never forced by an unavoidable conflict. Against the pinned base, only six files delete anything at all (35 lines total); five of them are byte-identical to the faithful mechanical merge, and every deleted line in `db_descriptor.cpp` is branch-authored blob-relocation code that moved into `core/blob_relocation.cpp`. AGENTS.md: 77 added, 0 deleted. No conflict markers anywhere.

---

**One correction to the record, since it points future auditing at the wrong commit.** The message says the rebase "resolved the `db_descriptor.cpp` conflict by taking the branch side wholesale". Measured per commit, that is not where it went:

| commit | `ParkTimeoutRegistry` | `shutdown()` | `<system_error>` |
| --- | --- | --- | --- |
| `06961b5f` (base) | 12 | 2 | 1 |
| `e4659bd9` | 12 | 2 | 1 |
| `76df853d` | **0** | **0** | **0** |
| `a2965c0b` | 0 | 0 | 0 |
| `66cc2eee` | 12 | 2 | 1 |

The whole stack already sits directly on the base (`880aea9b^ == 06961b5f`), and `e4659bd9` still had all of it — so the rebase preserved main's #744 work. The deletion happened in **`76df853d` "Make the blob relocation rules executable on an unpatched build"**, whose `db_descriptor.cpp` numstat is 30/278: roughly 94 of those deletions are the inline blob rules it meant to extract, and the other ~184 are the `ParkTimeoutRegistry` block it took with them. `a2965c0b` "Rebase tiered storage onto main" is a 3-line commit that only changed `flush()` → `flush(bool allowWriteStall)` plus one Markdown indent; it performed no rebase and did not cause this.

Worth naming because the mechanism is different and still unexamined: not a conflict resolution, but a refactor commit silently deleting 184 unrelated lines next to the ~94 it was extracting. A rebase conflict at least announces itself.

—
_Generated by Barber AI_

src/binding/database/db_settings.cpp
cb1kenobi: Re-reviewed at `66cc2eeead76c18eac3c0e345e424ac1f911fa7c`. **Still open** — and this thread is now marked resolved for the third time, still with no reply and no code change behind it.

This push touches exactly one file, `src/binding/database/db_descriptor.cpp`. It changes **neither the behaviour nor the docs** for this finding.

Re-measured rather than read off the constant. Fresh process, the exact caller shape in question — `config({ blockCacheSize: 100 MB })`, `blobCacheSize` never set:

```
rocksdb.block-cache-capacity = 104857600  (100.00 MB)
rocksdb.blob-cache-capacity  =  10485760  ( 10.00 MB)   <- a second, separate cache
TOTAL                        = 115343360  (110.00 MB)
```

So the +10% ceiling on upgrade is unchanged and real.

The latch half remains correct: `config({ blobCacheSize: 0 })` followed by `config({ blockCacheSize: 100 MB })` leaves `rocksdb.blob-cache-capacity` absent entirely — no blob cache attached, ceiling stays 100 MB. `blobCacheSizeExplicit` holds.

The documentation is accurate about this — `RocksDatabaseConfig.blobCacheSize` in `src/load-binding.ts` states the derived capacity is "**additional** to the block-cache capacity" and that it "raises the process memory ceiling by 10% with no code change". Nothing here is an oversight; this is a deliberate default awaiting @cb1kenobi's ratification. Flagging only so it is ratified rather than absorbed by a resolve click.

—
_Generated by Barber AI_

AGENTS.md
```diff
@@ -581,6 +581,83 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     is the descriptor's single `CommitWorker` thread (see "Commit execution" above), which dispatches
     every `Transaction.commit()` in order — so opting a flush into a stall queues up every commit
     behind it, including ones from callers that never touched flush.
+17. **File placement is recorded two different ways, and only one survives a config change**: `paths`
```
cb1kenobi, line 633: **Low: this invariant number collides with main's new invariant 17, and the merge is clean so nothing warns**

Main moved while this PR sat: [#780](https://github.com/HarperFast/rocksdb-js/pull/780) merged as `6d2c4cd8` and added its own **17. "An env's pending transactions are reaped by its cleanup hook — never by `DBHandle::close()`"**. This branch adds 17 (`File placement...`) and 18 (`Every per-column-family option...`) on a base that stopped at 16.

`git merge-tree --write-tree 66cc2eee 6d2c4cd8` merges **cleanly** (exit 0), and the resulting `AGENTS.md` contains two invariant 17s:

```
15. A callback-style native method owes its caller exactly one settled callback...
16. FlushOptions::allow_write_stall defaults to the waiting behavior...
17. File placement is recorded two different ways...          <- this branch
18. Every per-column-family option belongs in buildColumnFamilyOptions...  <- this branch
17. An env's pending transactions are reaped by its cleanup hook...  <- main
```

This exact collision has already cost this repo two commits: `49242df5` "docs: fix duplicate invariant number that broke `pnpm check`" and, today, `46096ac3` "docs: fix invariant numbering and require `pnpm fmt` before commit" — which added the warning that oxfmt covers TS/JS/JSON only, so a mis-numbered Markdown invariant slips past `fmt:check` untouched.

Suggested fix: on the next rebase onto main, renumber this branch's two invariants to **18** and **19** so main's 17 keeps its number. Nothing cites either by number yet — I checked `src/`, `test/` and `docs/` on both sides, and the only numeric citations are to invariants 5, 14 and 15 — so the change is confined to `AGENTS.md`.

—
_Generated by Barber AI_

kriszyp:
Two things landed since `ccdbcc3e`.

**The rebase onto main dropped `ParkTimeoutRegistry`.** `db_descriptor.cpp` conflicted and the resolution took the branch side wholesale, deleting every hunk #744 had added to that file while leaving the class declared in `db_descriptor.h` and used seven times in `transaction.cpp` — four unresolved symbols, so `dlopen` failed outright on Linux and Windows and macOS linked under `-undefined dynamic_lookup` and crashed on `close()`. `66cc2eee` re-applies main's four hunks verbatim. Verified after the fact: `git diff origin/main HEAD` has no park-related delta at all, every one of its 35 deletions is accounted for by this feature, `nm` reports the four symbols defined and none undefined, and the #741 park-timeout cases pass. CI on that commit is green on every platform that has reported.

**An unset `blob_dir` is not the database directory on a tiered database** (`6ae5562d`). RocksDB derives blob paths from `cf_paths.front()`, which falls back to `db_paths.front()`, and the downstream patch keeps that as the empty-value behavior — so a database that also sets `paths` keeps its flat blob files on `paths[0]`. `decideBlobRelocation` resolved an empty directory to the database directory unconditionally, so on a tiered database the `allowDirChange` guard read a directory the files were never in: no `.blob` files found, acknowledgement accepted, and the database opens with every value at or above `min_blob_size` unreadable. `BlobRelocationInput` now carries the resolved default; six native cases cover it and three fail against the old resolution. Same commit short-circuits `assertStoragePathsUsable` when `paths[0]` is spelled exactly as the database directory — the supported form of the zero-to-one migration, where it otherwise stat'ed every SST file of the database on every open.

Directory isolation stays caller-managed per @kriszyp's ruling; the reasoning is recorded in the description. One new question for the human reviewer is there too: a plain `RocksDatabase.open(path)` throws on an already-open family with an external `blob_dir`, because `BlobOptions::dir` is deliberately not optional.

Local verification: `pnpm build`; `pnpm test` (853 passed, 24 skipped); `pnpm test:native` (189); `pnpm check`.

Signed: Claude Opus

kriszyp:
CI note, and it needs a decision rather than a fix from me.

`Check` is red on `oxfmt --check AGENTS.md`, and it is **not** from anything in this branch's commits. GitHub builds a pull request from the *merge* of the branch into the current base, and `main` moved to v2.8.0 while this was in review — #780 added a new invariant numbered **17**, which is the number this branch already uses for tiered storage (and it also uses 18). In the merged file the ordered list reads 17, 18, 17, and oxfmt renumbers the last one to 19.

Reproduced both ways: `git merge-tree --write-tree origin/main <sha>` produces an `AGENTS.md` that oxfmt rejects for `66cc2eee` — the commit whose `Check` was green when it ran, against the older `main` — exactly as it does for `6ae5562d`. Each side's `AGENTS.md` passes on its own.

Nothing in the branch can fix it: my two invariants come first in the file, so the correct renumbering is on the incoming item, and that text only exists here after an integration. It clears when the branch is brought up to `main` and the invariants are renumbered — which is a force-push, so I have left it alone.

One thing worth carrying into that integration: the last rebase resolved `db_descriptor.cpp` by taking the branch side wholesale and silently deleted main's #744 hunks (see the previous comment). `main` has since landed #780, which touches `db_descriptor.cpp`, `transaction.cpp`, `db_registry.cpp`, `store.ts` and `load-binding.ts` — the same files this branch touches. Two checks catch that class of mistake cheaply afterwards: `nm -C -u build/Release/rocksdb-js.node` must report no undefined `rocksdb_js::` symbols, and every deletion in `git diff origin/main HEAD` should be attributable to this feature.

Everything else is green — 21 checks pass, two Windows test jobs still running.

Signed: Claude Opus

AGENTS.md
kriszyp: The collision is real, but the suggested direction is backwards for the merge GitHub actually
builds, and applying it now would make things worse rather than better. Measured against
`origin/pull/767/merge` (`b19fbfcf`, `6ae5562d` into `6879b823`) rather than reasoned about:

In that merge both sides insert at the same anchor — the end of invariant 16 — and this branch's
block lands **first**:

```
16. FlushOptions::allow_write_stall ...
17. File placement is recorded two different ways ...          <- this branch
18. Every per-column-family option belongs in ...              <- this branch
17. An env's pending transactions are reaped by its cleanup hook ...   <- main (#780)
```

So the item that is out of sequence is the incoming one, and `oxfmt 0.64.0` agrees: run on that
merged file its only edit is `17.` → `19.` on main's invariant. This branch's two numbers are
already the correct ones for that ordering.

Renumbering them to 18/19 here fails both ways, checked with the same oxfmt:

- **On the branch alone** the list reads 16, 18, 19 and `oxfmt --check AGENTS.md` fails — it
  rewrites them straight back to 17/18. The branch's `AGENTS.md` passes today; that change is what
  would break `pnpm check` locally.
- **In the merge** the list reads 18, 19, 17, which oxfmt renumbers to 17, 18, 19 — so `Check`
  stays red, just for a different reason.

The suggestion is right for the *other* integration path: a rebase replays this hunk onto a main
that already has 17, our two land after it, and 18/19 is then correct. That is the fix, and it can
only be made as part of the integration — which is a force-push, so it is not this task's to make
(the branch is checked out elsewhere). Leaving the thread open so it is picked up there. Whoever
does it: renumber to 18/19 *after* the rebase places the block below main's 17, and re-run
`pnpm fmt:check` on the result rather than on either side alone — each side passes on its own,
which is exactly why nothing warned.

— Claude Opus

kriszyp: Fixed in `de700365`, and your numbering is what landed — 18 and 19 — but it took the integration
to get there, so the earlier reply on this thread ("the incoming item is the one out of sequence")
described the *auto-merge* ordering rather than the ordering that shipped.

`origin/main` is now merged into this branch rather than rebased onto it. The 3-way merge is
conflict-free in all seven overlapping files, where a rebase would replay 22 commits across
`db_descriptor.{cpp,h}`, `db_registry.{cpp,h}`, `store.ts` and `load-binding.ts` — the operation
that silently dropped #744's hunks from this branch once already (`66cc2eee` had to restore them).
#780 merged `main` into its own branch three times before landing, so this is the prevailing shape
here.

Git's auto-merge put this branch's block first and would have renumbered #780's invariant to 19.
I overrode that and kept **main's invariant at 17**, moving this branch's two to **18** and **19**
— exactly what a rebase would have produced. That matters beyond taste: `AGENTS.md` is now a
*pure insertion* over `origin/main` (zero deletions in `git diff origin/main -- AGENTS.md`), so
the file's content no longer depends on which integration path this branch eventually takes, and
no invariant number that anything else references moves. oxfmt 0.64.0 also wanted a blank line
before the moved block, which is in.

Verified on the merge commit:

| check | result |
|---|---|
| `pnpm check` | pass, 191 files — the job this thread was blocking |
| `pnpm test` | 858 passed, 24 skipped, 0 failed |
| `pnpm test:native` | 190/190 |
| `nm -C -u build/Release/rocksdb-js.node` | no undefined `rocksdb_js::` symbols |
| dropped-hunk audit | `CloseTransactionsByEnv`, `closeTransactionsByEnv`, `ParkTimeoutRegistry`, `onWrapperCollected`, `wrapperCollected` all match their `origin/main` occurrence counts; `src/binding/transaction/`, `binding.cpp` and `napi/async.h` byte-identical to main |
| deletion audit | 35 deletions vs `origin/main`, every one attributable to this feature |
| #780's own repro | all three `test/lingering-txn-shutdown.test.ts` cases pass on the merged tree |

Signed: Claude Opus

src/binding/database/db_descriptor.cpp
```diff
@@ -2135,4 +2582,22 @@ rocksdb::Status DBDescriptor::compactRange(
...
+ *
+ * `layoutMutex` is released before that call: `DBRegistry::DestroyDB` takes it
+ * while holding `databasesMutex`, so recording must not reach back for a
+ * registry lock from under it.
+ */
+void DBDescriptor::recordColumnFamilyLayout(const std::string& name, const std::string& blobDir) {
+	DBFileLayout layout;
+	{
+		std::lock_guard<std::mutex> lock(this->layoutMutex);
+		this->layoutBlobDirs[name] = blobDir;
```
cb1kenobi, line 2883: **Stale dropped-family directories remain destroy targets — High**

`layoutBlobDirs` is only inserted or updated; the existing column-family drop path never erases the entry or refreshes `knownLayouts`. After a family is dropped, a later `destroy()` still passes its former `blobs.dir` to RocksDB, so reassigning that now-unused directory to another database allows the old database’s destroy to delete the new database’s live `.blob` files. Add a layout-removal method and invoke it after `DropColumnFamily` succeeds, erasing the family and recording the updated layout.

—
_Generated by Barber AI_

kriszyp: Addressed in 51b67183 and strengthened in 434ac008. A successful drop now prunes the family by name from every live descriptor layout for the database path and from the retained knownLayouts entry before unregistering the name, so a same-name recreation cannot be erased and a stale re-drop remains mutation-free. Regression coverage exercises sync/async drop, live and retained destroy layouts, directory reuse, same-name recreation, and shared blob directories. Local verification: pnpm build; pnpm test (860 passed, 30 skipped); pnpm test:native (197/197); pnpm check. The five new blobs.dir cases are capability-skipped on the current unpatched prebuild, matching the existing patched-build merge gate in the PR description.

src/binding/database/db_descriptor.cpp
```diff
@@ -1468,9 +1915,11 @@ uint32_t DBDescriptor::transactionGetNextId() {
  * creates a fresh column family instead of reusing the dangling dropped
  * handle. DBHandles still holding the descriptor keep it alive via their
  * shared_ptr and can continue reading until they close; only the by-name
- * lookup is removed.
+ * lookup is removed. The family is removed from the destroy layout first so a
+ * concurrent open cannot register a fresh same-name family before that cleanup.
  */
 void DBDescriptor::unregisterColumnFamily(const std::string& columnName) {
+	DBRegistry::RemoveColumnFamilyLayout(this->path, columnName);
```
cb1kenobi: **Concurrent reopen can reuse the dropped family — High**

This call releases `databasesMutex` before `columnsMutex` is acquired below. A warm `OpenDB` can enter during that gap, find the old `columns[columnName]`, and return a handle to the already-dropped RocksDB column family, so `open()` succeeds but subsequent operations fail. Make layout removal and erasing the column entry one registry-atomic operation, taking `databasesMutex` and then the descriptor locks in the established order.

—
_Generated by Barber AI_

kriszyp: Fixed in ba65fed0 — and the gap turned out to be wider than the one this comment names.

`DBRegistry::DropColumnFamily` now holds `databasesMutex` across `DB::DropColumnFamily` **itself**, both layout erasures, and the `columns` erase. Closing only the gap between the two cleanup steps would have left the older one: between `DropColumnFamily()` returning and the unregister starting, a warm open could already be handed the dropped family. Putting the RocksDB call inside the same section makes an open observe the family wholly present or wholly gone.

The verification-table sweep deliberately stays outside that section — the VT's writer mutex is process-global and must never be taken under the registry mutex.

Covered by a deterministic regression rather than a timing race: a test-only two-phase latch parks the drop mid-section until a worker's warm open has actually reached the registry mutex (keyed by database path, since the counters are process-global and Vitest's thread pool shares them), then holds on. Mutating the section back into two lock acquisitions fails it — the worker receives the dropped family and its write is silently discarded. Local verification: `pnpm build`; `pnpm test` (861 passed, 31 skipped); `pnpm test:native` (197/197); `pnpm check`.

One deliberate non-change: the drop unregisters the by-name column only from the descriptor that performed it. A read-only entry for the same path is a separate `DB::OpenForReadOnly` instance whose column-family handles a writable drop does not invalidate, and which by design never observes later changes; unregistering the name there would make the next read-only open throw "cannot create column family in read-only mode" for a family its own frozen manifest view still contains. That rationale is now recorded at the call site.

— Claude Opus 5

src/binding/database/db_registry.cpp
```diff
@@ -229,17 +297,137 @@ void DBRegistry::DestroyDB(const std::string& path) {
...
+	if (!instance) {
+		return;
+	}
+	std::lock_guard<std::mutex> lock(instance->knownLayoutsMutex);
+	const bool defaultLayout = layout.dbPaths.empty() &&
+		std::all_of(layout.blobDirs.begin(), layout.blobDirs.end(), [](const auto& entry) {
+			return entry.second.empty();
+		});
+	if (defaultLayout) {
+		instance->knownLayouts.erase(path);
```
cb1kenobi: **High: a secondary open can erase the tiered destroy layout**

Registry entries are separated by read-only/mode, but this path-keyed layout is replaced by whichever descriptor opens last. With a writer using `[dbPath, B]`, a concurrent read-only open that omits `paths` can succeed while all SSTs still use path index 0 and then erase the remembered layout here; if the writer subsequently compacts files onto `B`, closing both descriptors followed by `destroy()` cannot recover `db_paths`, so it reports success while leaving those SSTs behind.

Suggested fix: maintain one canonical append-only `db_paths` layout per database path—an omitted or shorter secondary open must inherit/preserve the longer live layout—and always use that canonical layout during destroy.

—
_Generated by Barber AI_

src/binding/database/db_registry.cpp
```diff
@@ -154,6 +177,39 @@ void DBRegistry::DestroyDB(const std::string& path) {
...
+	// Retained record first, then the live layout: paths only where they extend
+	// the record, blob directories only for families it does not already name.
+	std::unordered_map<std::string, std::string> destroyBlobDirs;
+	auto applyLayout = [&](const DBFileLayout& layout) {
+		if (extendsDbPaths(destroyOptions.db_paths, layout.dbPaths)) {
+			destroyOptions.db_paths = layout.dbPaths;
+		}
+		for (const auto& [cfName, blobDir] : layout.blobDirs) {
+			auto [it, inserted] = destroyBlobDirs.emplace(cfName, blobDir);
+			if (!inserted && it->second.empty()) {
```
cb1kenobi: **Empty retained blob directories lose to stale descriptors — High**

Severity: High. An empty `blob_dir` is a real placement, but this branch treats it as missing. After relocating blobs from an external directory back to the default while an older read-only descriptor remains live, `destroy()` replaces the retained empty value with the stale external directory and can delete a replacement database that reused it. Keep the retained value even when empty; an existing map key must always win.

—
_Generated by Barber AI_

src/binding/database/db_registry.cpp
```diff
@@ -229,17 +332,147 @@ void DBRegistry::DestroyDB(const std::string& path) {
...
+		layout.dbPaths = known->second.dbPaths;
+	}
+	const bool defaultLayout = layout.dbPaths.empty() &&
+		std::all_of(layout.blobDirs.begin(), layout.blobDirs.end(), [](const auto& entry) {
+			return entry.second.empty();
+		});
+	if (defaultLayout) {
+		instance->knownLayouts.erase(path);
+	} else {
+		instance->knownLayouts[path] = std::move(layout);
```
cb1kenobi: **A secondary reader can poison an untiered destroy layout — High**

Severity: High. Default layouts are erased above, so a read-only open with `paths: [dbPath, neighborVolume]` can succeed while every SST still uses index 0 and then record `neighborVolume` here. A later `destroy()` sweeps the neighboring database’s live SSTs. Retain an authoritative untiered-layout marker and prevent secondary/read-only descriptors from extending it; only a validated writable cold transition should establish appended paths.

—
_Generated by Barber AI_

src/binding/database/db_registry.cpp
kriszyp: Fixed in 2759485f (implementation in a8ace705). DestroyDB now treats the retained path-keyed layout as canonical and no longer merges any live descriptor snapshot, so an authoritative empty blob_dir cannot be overwritten by a stale read-only descriptor. The native merge tests cover empty replacement, and the patched-build integration case keeps replacement-database blobs intact.

src/binding/database/db_registry.cpp
kriszyp: Fixed in 2759485f (implementation in a8ace705). RecordLayout now keeps an authoritative default marker and permits only writable opens to establish or append db_paths; a read-only secondary cannot add a neighbor volume to destroy targets. The stock-build integration regression leaves the secondary open through destroy and verifies the neighbor SSTs remain intact.

src/binding/database/db_descriptor.cpp
```diff
@@ -1352,18 +1803,43 @@ std::shared_ptr<DBDescriptor> DBDescriptor::open(const std::string& path, const
...
+#ifdef ROCKSDB_HAS_CF_BLOB_DIR
+		layoutBlobDirs[options.name] = cfo.blob_dir;
+#endif
 	}
 
 	DEBUG_LOG("DBDescriptor::open Creating DBDescriptor for \"%s\"\n", path.c_str());
 	auto descriptor = std::shared_ptr<DBDescriptor>(new DBDescriptor(path, options, cfOptions, db, std::move(columns), dbOptions.statistics));
+	descriptor->layoutDbPaths = dbOptions.db_paths;
+	descriptor->layoutBlobDirs = std::move(layoutBlobDirs);
+	if (!DBRegistry::RecordLayout(path, descriptor->captureLayout(), !options.readOnly) &&
```
cb1kenobi: **Reject refused writable path changes — High**

A writable cold open with a shorter or divergent `paths` list succeeds even when `RecordLayout` returns false; only a warning is emitted. RocksDB then uses the requested list while `destroy()` uses the old retained list, so compaction can create SSTs that destroy silently orphans, or destroy can sweep an old volume after another database reuses it. Treat a false result for a writable open as an open failure, ideally by validating the list before opening.

—
_Generated by Barber AI_

kriszyp: Fixed at `22cecbfc`, and you were right that validating before the open is the better shape — that is where it now happens.

`DBRegistry::AssertDbPathsExtendRetained` runs right after `dbOptions.db_paths` is built and before `DB::Open`, so a writable open whose list neither matches nor appends to the retained record fails outright instead of warning. Past `DB::Open` a single compaction placing a single SST is already the divergence `destroy()` cannot repair, so unwinding afterwards would have been too late. `RecordLayout` throws on the same condition rather than returning a bool nobody acted on, but only as an assertion now — reaching it means the record moved underneath an open database.

The shape worth naming: the dangerous case is the quiet one. While every file still sits at path index 0, RocksDB opens happily under a shorter or divergent list, so nothing complains until a compaction spills onto a volume `destroy()` never sweeps. `test/tiered-storage.test.ts` uses exactly that setup, and asserts no new OPTIONS file appears — RocksDB writes one on every open, so an unchanged set is what proves the refusal happened before the open rather than being unwound after one. Removing the pre-open call makes that assertion fail (`OPTIONS-000007, OPTIONS-000015` vs `OPTIONS-000007`).

One consequence worth flagging: this preempts `explainOpenFailure`'s removed/reordered guidance whenever the same process opened the database earlier, which is what the old in-process test was asserting. That guidance still matters for the case it was written for — a restart, where the record is gone and RocksDB reports the MANIFEST as corrupt — so its coverage moved to a spawned child (`test/fixtures/paths-removed-list.mts`).

Also in this push: `origin/main` merged in (one conflict, the include block in `db_descriptor.cpp`, both sides kept).

Signed: Claude Opus

Issue #726, opened 2026-07-27
Pessimistic transactions still poison the environment on a drop-race commit


kriszyp:
## What

`DBHandle::transactionWriteOptions()` (`src/binding/database/db_handle.h`) deliberately
excludes `ignore_missing_column_families` — setting it would turn a transaction commit that
spans a live column family and a dropped one into a silent partial commit (live half applied,
dropped half discarded, reported as success). That would break the atomicity
`Transaction::Commit` promises, so PR #725 kept this off the transactional path.

The tradeoff: in **pessimistic** mode, a commit that races a column-family drop reaches
RocksDB's fatal path and latches a background error on the whole environment — the same
poisoning behaviour PR #725 fixes for non-transactional `putSync`/`removeSync`, still reachable
here. In **optimistic** mode (Harper's default, and the only mode Harper uses) this is already
safe: conflict validation rejects the commit early with an attributable
`Could not access column family <id>` error and nothing is applied.

This is a **pre-existing bug, unchanged by PR #725** — not a regression it introduces. It's
already documented at the call site (`db_handle.h`, see `transactionWriteOptions()`'s doc
comment) and pinned by a test (`test/drop.test.ts`, "should not partially apply a pessimistic
transaction spanning a dropped column family") that asserts only atomicity and explicitly notes
the poisoning is a separate issue.

## Why it needs real work, not a small patch

A transaction's touched column-family set isn't known at `BeginTransaction` — it's only
finalized as writes are staged. A correct fix needs:

- Per-column-family "in-flight touch" tracking, incremented on each transactional
  `Put`/`Delete` and decremented on commit/rollback.
- A "dropping" flag on `ColumnFamilyDescriptor` so a write racing an in-progress drop fails
  fast instead of proceeding.
- `Database::Drop`/`DropSync` (`src/binding/database/database.cpp`) checking/waiting on that
  count before calling `DropColumnFamily`, coordinated with the existing `columnsMutex` lock
  ordering and the async commit pipeline (`CommitWorker`) without introducing deadlocks against
  RocksDB's own lock timeouts.

This spans `db_descriptor.h/.cpp`, `db_handle.h/.cpp`, `database.cpp`'s transactional write
paths, and `transaction.cpp`/`transaction_handle.cpp`'s commit/rollback paths, plus new
concurrency tests — a cross-cutting hot-path change with real deadlock/perf risk, not
appropriate to rush into a review-response patch.

## Refs

[...]

PR #768, opened 2026-08-09
Release a transaction dropped without commit or abort, instead of leaking its read snapshot for the life of the process

kriszyp:
A transaction dropped without `commit()` or `abort()` used to live for the life of the process, holding a RocksDB read snapshot that stopped the database from ever discarding obsolete row versions. It now releases itself when V8 collects its JS wrapper.

Reported as [HarperFast/harper#2107 — RocksDB read snapshot leaks permanently](https://github.com/HarperFast/harper/issues/2107): on a production cluster a high-churn secondary index reached ~4 keys per live row in 5 days, `numberReseeksIteration` went 51 → 43,647, and bounded range scans degraded ~21x, with a process restart as the only recovery.

## The defect

`DBDescriptor::transactionAdd` stores a **strong** `shared_ptr` in the `transactions` map, while the `closables` entry written on the very next line is a `weak_ptr`. `TransactionHandle::close()` is the only path that calls `ClearSnapshot()`, and it is reached from `~TransactionHandle()` — but the destructor can never run while the registry holds its reference. The napi finalizer only did `(*txnHandle).reset()`, which drops the JS-side reference and nothing else. So the destructor was correct and unreachable, and any caller that dropped a transaction leaked it permanently and silently.

The finalizer now calls `onWrapperCollected()` first. Once V8 has collected the wrapper, no JS code can commit, abort, retry, or read through that handle again, so it is closed. A commit in flight is the one exception — `TransactionCommitState` still owns the handle, and closing there would cancel the commit mid-flight — so the commit-completion paths close it instead: success already closed unconditionally, and the failure paths, which deliberately reset to `Pending` so a caller can retry, now check `wrapperCollected` because there is no caller left.

**Why not make the registry reference `weak_ptr`.** That was the first suggestion in the issue and it is not safe here: an async `get` holds a raw `TransactionHandle*` (`AsyncGetState<TransactionHandle*>`) and depends on `close()` running `cancelAllAsyncWork()` / `waitForAsyncWorkCompletion()` before the transaction is destroyed. Letting the last `shared_ptr` drop destroy the handle would race that. Going through `close()` from the finalizer gets the same self-healing with that machinery intact.

## Observability

`registryStatus()` now reports `transactionDetails` — `id` and `ageMs` per live handle — and `rocksdb.num-snapshots` is exposed as a curated stat. A bare transaction count cannot distinguish a request in flight from a database that can never reclaim again; a nonzero snapshot count plus a handle age beyond any plausible request lifetime can. The `registryStatus()` read of the transactions map also now takes `txnsMutex`, which it was missing.

`transactionDetails` deliberately does **not** report `snapshotSet` or `state`. See the decision below.

## Verification

`pnpm test` — 752 passing, 1 skipped, on Node. `pnpm check` clean. `pnpm build:binding:debug` compiles.

New `test/transaction-orphan-gc.test.ts` drives the real native stack (no mocks): a dropped transaction, a dropped transaction after a rejected commit, a still-referenced transaction (control — must not be disturbed), a caller that drops its reference before the commit settles, and an orphan that never read. The first two fail on `main`; the controls pass on both, so they are not vacuous.

The reproducer behind the issue, against `main` before this change:

```
baseline           num-snapshots= 0  oldest-snapshot-time= 0           registry.transactions= 0
after read in txn  num-snapshots= 1  oldest-snapshot-time= 1786236991  registry.transactions= 1
after GC x8        num-snapshots= 1  oldest-snapshot-time= 1786236991  registry.transactions= 1
```

## Not fixed here

`if (state == Committing) state = Pending` in the commit-completion paths is a cross-thread compare-then-store that can overwrite an `Aborted` set concurrently by `close()`. Review surfaced it; it is pre-existing, independent of this change, and wants its own fix rather than being folded into a leak fix. Filed as #769.

## For the human reviewer

1. **Finalizer-driven `close()` rather than a weak registry reference.** Chosen because an async `get` holds a raw `TransactionHandle*` and relies on `close()`'s cancel-and-wait; a destructor triggered by the last `shared_ptr` drop would skip it. The alternative — weak registry ref plus changing `transactionRemove` to take a raw pointer (`close()` currently calls `shared_from_this()`, which would throw `bad_weak_ptr` from a destructor) — is a larger lifetime change that also brushes against the open #741 and #726. Reversible, but it would be a rewrite of this approach rather than a tweak. Say no if you'd rather take the lifetime change properly.

2. **`transactionDetails` reports only `id` and `ageMs`.** An earlier revision reported `snapshotSet` and `state` and made both `std::atomic` to do it safely. Review showed that was the wrong trade: it broke the debug build (`DEBUG_LOG` is variadic, `std::atomic` has no copy constructor), the implicit operators default to `seq_cst` so every state and snapshot check on the read/write paths grew a full barrier for an occasional diagnostic read, and it still didn't make the compare-then-store transitions atomic. `id` and `createdAt` are fixed before the handle is published to the registry, so they are race-free by construction. Cost: you cannot see per-handle "is this one holding a snapshot" — `rocksdb.num-snapshots` answers that per database instead. Cheap to revisit if the per-handle flag turns out to matter operationally.

3. **The `Committing` deferral branch is unreachable from JS.** A pending commit promise's executor still holds the wrapper, so V8 cannot collect it while the commit is in flight. The branch is kept as a correctness guard for any future path that can drop the wrapper mid-commit (a native-side caller, or a JS commit shape that doesn't retain), and the test that used to claim to exercise it has been renamed to what it actually proves. Alternative is deleting the branch and asserting the invariant instead; kept because the cost is three lines and the failure mode it guards is a cancelled in-flight commit.

**Where to look hardest:** `onWrapperCollected()` in `src/binding/transaction/transaction_handle.cpp` and the two `wrapperCollected` checks in `src/binding/transaction/transaction.cpp`. The question worth the most scrutiny is whether `state == Committing` is a sufficient test for "a commit still owns this handle" on every path that can reach the finalizer.

**What the tests do not prove:** cross-worker registry safety under concurrent envs, and finalization during an active commit (unreachable, above). The `pnpm test:bun` and `pnpm test:deno` routes are handled by the GC adapter in the new test file but were not run here.

**Coverage:** Codex and Gemini both ran on the final artifact; the Harper-domain leg ran on the previous revision. Findings acted on: the cross-environment field race, the debug-build break, the seq_cst hot-path regression, the Bun GC adapter, the overclaiming mid-commit test, and the missing never-read orphan case. Findings dropped after checking the code: Gemini's null-dereference on `state->handle` (the enclosing condition already requires a non-null handle) and the extra `steady_clock` read per transaction (a vDSO read next to `BeginTransaction()`, and `steady_clock` is the correct source for an age that must survive wall-clock adjustments).

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 4db2a0bca5df</sub>

<sub>Human-Review-Need: 4 @ 4db2a0bca5df</sub>


AGENTS.md
```diff
@@ -217,7 +217,15 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     the type. `tsc` enforces this.
   - **Fixture helpers must be `src`-free** only where they'd otherwise pull a heavier graph — e.g.
     `createWorkerBootstrapScript` lives in `test/lib/worker-bootstrap.ts` (no `src` import), separate
-    from `test/lib/util.ts` (which imports `src`); every call site imports it directly.
+      from `test/lib/util.ts` (which imports `src`); every call site imports it directly.
```
cb1kenobi: **High: this line fails `oxfmt --check`, and the `Check` CI job is red because of it**

The rebase commit `4db2a0bc` re-indented this continuation line from 4 to 6 spaces. It belongs to the `**Fixture helpers must be `src`-free**` bullet, which the base owns and this PR otherwise does not touch. `AGENTS.md` formats clean at the new base `276f8faf` and fails at this head — I reproduced it locally with the repo's own `oxfmt`, and the CI log for the failing `Check` run shows the same single file:

```
$ oxfmt --check
AGENTS.md (260ms)
Format issues found in above 1 files.
[ELIFECYCLE] Command failed with exit code 1.
```

Reverting the indent is the whole fix (`pnpm fmt` produces exactly this):

Suggested fix:
```suggestion
    from `test/lib/util.ts` (which imports `src`); every call site imports it directly.
```

—
_Generated by Barber AI_

src/binding/transaction/transaction_handle.cpp
```diff
@@ -253,6 +254,26 @@ void TransactionHandle::releaseIntent() {
...
+	this->wrapperCollected.store(true);
+
+	if (this->state == TransactionState::Committing) {
+		DEBUG_LOG("%p TransactionHandle::onWrapperCollected Commit in flight, deferring close (txnId=%u)\n", this, this->id);
+		return;
+	}
+
+	DEBUG_LOG("%p TransactionHandle::onWrapperCollected Closing orphaned transaction (txnId=%u, state=%d)\n",
+		this, this->id, static_cast<int>(this->state));
+	this->close();
```
cb1kenobi, line 298: **Medium: GC-time `close()` can free the handle out from under a still-running async `get`, because the wait it relies on is allowed to time out silently**

This is a new interaction with the rebased base rather than a change in this PR's own code — the diff is byte-identical to the previously reviewed revision, but the base moved underneath it.

The safety argument for calling `close()` from a finalizer is that `close()` cancels and waits for in-flight async work before destroying the transaction. That wait does not actually guarantee it:

```cpp
// src/binding/napi/async.h:162-179
void waitForAsyncWorkCompletion(std::chrono::milliseconds timeout = std::chrono::milliseconds(5000)) {
    ...
    if (elapsed >= timeout) {
        DEBUG_LOG(... "timeout waiting for async work completion, %u items remaining" ...);
        return;   // gives up, no signal to the caller
    }
```

`close()` calls it with the default and then proceeds unconditionally to `this->txn->ClearSnapshot(); delete this->txn;` (`transaction_handle.cpp:342-344`). It has already dropped the registry's strong reference via `transactionRemove` at the top, so when control returns here and the finalizer runs `(*txnHandle).reset()` (`transaction.cpp:117`), that is the last reference and the handle is freed.

What makes this newly reachable is the pairing of two independent changes:

1. **This PR** makes an orphaned `TransactionHandle` destroyable at all. Before it, the registry's strong reference kept the object alive forever, so a timed-out `close()` left a live-but-cancelled object and the execute handler's `state->handle->isCancelled()` guard read valid memory.
2. **The new base** added worker-thread dereferences inside the same execute handler:

```cpp
// transaction_handle.cpp:515-519 (base-added)
vtCheckAsyncGet(state, state->handle->dbHandle->descriptor->db.get(), ...);
```
plus `vtCheckLatest` reading `state->readOptions.snapshot` on the worker (`database.h:440` → `database.h:118` `readSnapshot->GetSequenceNumber()`) — a raw snapshot pointer captured at queue time, which `ClearSnapshot()` has now released.

`state->handle` is a bare `TransactionHandle*` (`AsyncGetState<TransactionHandle*>`), so past the timeout even the `isCancelled()` guard on line 503 is itself a use-after-free.

Reachability is narrow but not exotic: the async-work registration is claimed at queue time, so the 5 s budget includes libuv queueing delay (default `UV_THREADPOOL_SIZE=4`). It also needs a directly-constructed `Transaction` rather than `db.transaction()`, whose async-function context pins the wrapper for the whole body — but `Transaction` is public API (`src/index.ts:56`) and this PR's own test constructs one that way.

The invariant worth enforcing rather than working around is in `waitForAsyncWorkCompletion` itself: a wait that is permitted to return without the condition holding is not a wait. Having it report the timeout, and having `onWrapperCollected()` re-arm/defer instead of destroying when it did not drain, fixes this at the root. Note the same pattern is relied on by `DBHandle::close()` (`db_handle.cpp:132-135`), so the fix likely belongs in `async.h` rather than in this PR alone — reasonable to split it out if you'd rather not grow a leak fix.

—
_Generated by Barber AI_

AGENTS.md
kriszyp: Fixed in d1567e73. The continuation is back at the formatter-owned four-space indent, and `./node_modules/.bin/oxfmt --check AGENTS.md` now passes. Thanks for the precise reproduction.\n\n— GPT-5 Codex

src/binding/transaction/transaction_handle.cpp
```diff
@@ -253,6 +254,26 @@ void TransactionHandle::releaseIntent() {
  * from multiple threads concurrently (e.g. DBDescriptor::close() on env M's
  * JS thread racing the async commit's complete callback on env W's JS thread).
  */
+/**
```
cb1kenobi, line 250: This method seems to have been inserted between the description of the `TransactionHandle::close()` method and the actual `TransactionHandle::close()` code. Please move `TransactionHandle::close()`'s docblock next to it's implementation.

src/binding/transaction/transaction_handle.cpp
cb1kenobi: Agreed — `close()` from the finalizer is not safe while an async get still holds the handle, because `waitForAsyncWorkCompletion` can time out and then `delete this->txn` under the worker.

Stacked fix in https://github.com/HarperFast/rocksdb-js/pull/789:

- `onWrapperCollected` still sets `wrapperCollected` and still defers if `Committing`.
- For a pending txn it **cancels** in-flight work, and **does not `close()`** while `activeAsyncWorkCount > 0`.
- `AsyncGetState` now pins the handle with a `shared_ptr` (same shape as async `Database::Get`).
- The execute handler sees cancelled/collected and rejects with `Transaction is closed` instead of calling `txn->Get`.
- The complete callback `close()`s once execute has finished, so the snapshot is still released (harper#2107).

That way a get that races GC throws rather than UAFing. In-flight commit is unchanged (`completeCommitWork` still owns close).

cb1kenobi (changes requested):

AGENTS.md
```diff
@@ -402,6 +410,27 @@ sufficient (env teardown does not honor tsfn acquire counts); see
...
+    only recovery (HarperFast/harper#2107; `test/transaction-orphan-gc.test.ts`). Two constraints on
+    any redesign here: the registry reference cannot simply be made weak, because an async `get`
+    holds a raw `TransactionHandle*` (`AsyncGetState<TransactionHandle*>`) and relies on `close()`'s
+    `cancelAllAsyncWork()`/`waitForAsyncWorkCompletion()`, which a plain destructor race would skip;
+    and `registryStatus()` may only report handle fields that are fixed before the handle is
+    published to the registry (`id`, `createdAt`), because `txnsMutex` covers the map's membership
+    while the mutable fields' writers — the owning thread's read paths, the commit-completion
+    callback — hold no lock at all.
+
 12. **A recovered active transaction-log file ends on a transaction boundary when recovery can
```
cb1kenobi: **High: the rebase left two invariants both numbered `12.`, `oxfmt --check` fails on it, and the `Check` CI job is red**

This PR inserts its own invariant as item `12.` immediately above the base's item `12.` ("A recovered active transaction-log file..."), which the base owns and this PR otherwise does not touch. That leaves a duplicate ordinal in the list, and `oxfmt` renumbers ordered lists, so `pnpm fmt:check` fails.

`AGENTS.md` formats clean at the merge base `23f21726` and fails at this head — reproduced locally with the repo's own `oxfmt` 0.63.0, and the CI log for the failing `Check` run shows the same single file:

```
$ oxfmt --check
AGENTS.md (276ms)
Format issues found in above 1 files. Run without `--check` to fix.
[ELIFECYCLE] Command failed with exit code 1.
```

This is the same class of breakage as the `AGENTS.md` indentation issue fixed at `d1567e73` — the numbering collided when the branch was rebased onto a main that had added its own item `12.`. Renumbering the base's item is exactly what `pnpm fmt` produces:

Suggested fix:
```suggestion
13. **A recovered active transaction-log file ends on a transaction boundary when recovery can
```

Worth a second on the alternative: appending this PR's invariant *after* the base's rather than before it would keep the base-owned line untouched and make this PR's item `13.` instead, which leaves less for the next rebase to trip over.

—
_Generated by Barber AI_

cb1kenobi:
Re-reviewed `78e67752` — both prior findings are fixed, and no new issues. Nice job on this one.

**Genuine new work: 2 commits, 9 files, +260/−48.** `git range-diff` with merge bases computed separately per head (both `23f21726`) maps the 8 earlier commits 1:1 as unchanged; the new work is `19ace325` and `78e67752`. No merge commits in the range, so no evil-merge content to isolate.

**High — `AGENTS.md:434`, duplicate `12.` — fixed** by `78e67752`, which renumbers the base-owned transaction-log invariant to `13.`. Confirmed with the formatter rather than the badge: `oxfmt 0.63.0 --check` **fails** on `AGENTS.md` at `132dc8fc` and **passes** at this head, with all 180 files clean. The `Check` job is green.

**Medium — `transaction_handle.cpp:274`, GC-time destroy racing worker-thread derefs — fixed** by `19ace325`, and fixed properly rather than papered over. The async `get` state now owns a `shared_ptr<TransactionHandle>` instead of a raw pointer, `closeOrphanIfUnused()` defers the orphan close while `activeAsyncWorkCount > 0` or `activeIteratorCount > 0`, and the last dependent to release retries it — the async-get execute handler after `signalExecuteCompleted()`, or `unregisterIterator()` on the final iterator release. `DBIteratorHandle` now owns the handle and resets its RocksDB iterator before unregistering, and `keepAlive = shared_from_this()` guards the case where `transactionRemove()` drops the registry's last reference on a worker thread.

This was verified by building both heads, not by reading. Applying **only** the `ROCKSDB_JS_TXN_GET_DELAY_MS` test seam to `132dc8fc` (marker confirmed present in the built `.node`, `closeOrphanIfUnused` confirmed absent from the object file) and running the new `async-get` fixture:

| build | delay | result |
| --- | --- | --- |
| `132dc8fc` + seam | 5250 ms (> the 5000 ms wait timeout) | **SIGSEGV** (exit 139) |
| `132dc8fc` + seam | 1000 ms (< timeout) | exit 1, assertion only — no crash |
| `132dc8fc` + seam | 0 ms | exit 1, assertion only — no crash |
| `78e67752` | 5250 ms | **exit 0** |

So the crash is specific to exceeding `waitForAsyncWorkCompletion`'s timeout, it is not an artifact of the seam, and this head closes it. The `iterator` mode fails at the old head too (`transactionCount()` was not 1 — the transaction was closed out from under a live iterator) and passes here. `test/transaction-orphan-gc.test.ts` is 7/7 at this head, including the two new cases.

One residual, unchanged and out of scope: `close()` itself still runs `ClearSnapshot(); delete this->txn;` after a `waitForAsyncWorkCompletion()` that can time out silently. That path is now unreachable from the GC/orphan route, but remains reachable from explicit abort, `DBDescriptor::close()`, and commit completion — which is exactly what #789 and #784 cover. Nothing to do here.

Checked and clean: no `napi_cancel_async_work` anywhere in `src/`, so there is no cancellation window where `activeAsyncWorkCount` reaches zero without a retry point; commit-completion paths still close a collected wrapper via their own `wrapperCollected` checks; and `state->handle` cannot be empty for this state type, so the unguarded `closeOrphanIfUnused()` call at the end of the execute handler is safe.

Your comment at `transaction_handle.cpp:250` is addressed — the `close()` docblock now sits directly above `close()`. All seven review threads on this PR are resolved. Your `CHANGES_REQUESTED` review is still on record and only you can clear it.

—
_Generated by Barber AI_

cb1kenobi (approved):
This is great!

PR #775, opened 2026-08-12
Add FlushOptions.allowWriteStall, and settle flush()/compact() promises on read-only databases

### LINKED ISSUE #774: flush() and compact() never settle their promise on a read-only database
`Database.flush()` and `Database.compact()` return a promise that can **never settle** when the database was opened read-only. Neither the `resolve` nor the `reject` callback is invoked — the native method just returns `undefined`.

## Where

`Database::Flush` takes `resolve`/`reject` as `argv[0]`/`argv[1]` and returns before creating the async work:

```cpp
// src/binding/database/database.cpp:579
napi_value Database::Flush(napi_env env, napi_callback_info info) {
	NAPI_METHOD_ARGV(2);
	UNWRAP_DB_HANDLE_AND_OPEN();

	if ((*dbHandle)->descriptor->readOnly) {
		NAPI_RETURN_UNDEFINED();   // <-- resolve/reject never called
	}

	napi_value resolve = argv[0];
	napi_value reject = argv[1];
	...
```

`Database::Compact` has the identical shape at `src/binding/database/database.cpp:360`.

The JS wrapper has no read-only guard of its own, so the caller is handed a forever-pending promise:

```ts
// src/database.ts:388
flush(): Promise<void> {
	return new Promise((resolve, reject) => this.store.db.flush(resolve, reject));
}
```

## Why it looks intentional but isn't

The early return is correct for the **sync** siblings — `FlushSync` (`database.cpp:561`) and `CompactSync` (`database.cpp:360`) return `undefined` and a no-op is a fine answer for a read-only database. The async pair inherited the same guard, but for them "return undefined" is not a no-op: the promise contract makes it a permanent hang, and `await db.flush()` never resumes.

## Impact

Any `await store.flush()` on a read-only handle wedges its caller for the life of the process, with no error, no log line, and a fully live event loop. If that await sits in a shared path (a durability gate, an apply loop, a shutdown sequence), everything behind it wedges with it.

Not currently reachable from Harper's replication receive path — the only read-only opens in `harper` are in `integrationTests/qa-scratch/` — so this is a latent bug rather than a live incident. It surfaced while auditing unbounded native awaits behind [harper-pro#678](https://github.com/HarperFast/harper-pro/pull/678).

## Suggested fix

Resolve immediately instead of returning, so async matches the sync no-op semantics:

```cpp
if ((*dbHandle)->descriptor->readOnly) {
	napi_value undefined;
	NAPI_STATUS_THROWS(::napi_get_undefined(env, &undefined));
	napi_value result;
	NAPI_STATUS_THROWS(::napi_call_function(env, undefined, argv[0], 0, nullptr, &result));
	NAPI_RETURN_UNDEFINED();
}
```

Rejecting would also be defensible (`flush()` on a read-only database is arguably a caller error), but resolving keeps the existing "read-only flush is a no-op" contract that `flushSync` already establishes, so it can't break a caller that flushes opportunistically.

Either way the invariant worth enforcing is broader than these two methods: **every binding entry point that accepts `resolve`/`reject` must settle exactly one of them on every path.** A quick audit of the other `napi_create_async_work` call sites for the same early-return shape would be worthwhile.

## Related, lower priority

The `complete` callback drops both callbacks when N-API reports cancellation:

```cpp
// src/binding/database/database.cpp:615
if (status != napi_cancelled) {
	... callResolve / callReject ...
}
delete state;
```

Nothing in the binding calls `napi_cancel_async_work` today, so this is unreachable in practice — but it's the same "promise abandoned" shape and would be worth closing off alongside the fix.

---
Generated by Claude Opus 5.

### LINKED ISSUE #790: User shared buffer finalizer dereferences a freed callbackRef during shutdown (SIGSEGV)
## Summary

`userSharedBufferFinalize` reads the buffer's borrowed `callbackRef` via `napi_get_reference_value` during env/process teardown, but that `napi_ref`'s ownership was transferred to the listener's threadsafe function in `addListener`. Once the listener is torn down at shutdown (via `removeListener` / `removeListenersByOwner` / `removeListenersByEnv`), the tsfn finalizer `napi_delete_reference`s it — so the buffer finalizer dereferences a freed `napi_ref`: a use-after-free.

Reported twice on macOS when sending `ctrl-c` to quit Harper (SIGSEGV during shutdown).

## Crash

```
PID ##### received SIGSEGV for address: 0x1
2   rocksdb-js.node   rocksdb_js::userSharedBufferFinalize(napi_env__*, void*, void*)
4   node              napi_env__::CallIntoModule<...CallFinalizer...>
6   node              v8impl::BufferFinalizer::FinalizeBufferCallback(char*, void*)
7   node              node::Buffer::CallbackInfo::Clean()
8   node              node::Environment::RunCleanup()
9   node              node::FreeEnvironment(node::Environment*)
11  node              node::worker::Worker::Run()
```

Symbolicated locally (macOS `.ips`), the faulting frame is `napi_get_reference_value` called from `userSharedBufferFinalize` — `EXC_BAD_ACCESS (KERN_INVALID_ADDRESS)`.

## Root cause

`getUserSharedBuffer(key, buf, { callback })`:

1. `addListener` creates a `napi_ref` and **transfers its ownership to the listener's tsfn** (passed as `thread_finalize_data`; the tsfn finalizer `finalizeListenerCallback` deletes it). See `src/binding/napi/event_emitter.cpp`.
2. The same raw ref is *also* stashed in the buffer's `UserSharedBufferFinalizeData::callbackRef` as a borrowed pointer.
3. At shutdown, tearing the listeners down releases the tsfn → its finalizer deletes the ref.
4. `userSharedBufferFinalize` then runs for the still-live external `ArrayBuffer` and calls `napi_get_reference_value(env, finalizeData->callbackRef, …)` on the **already-freed ref** (`src/binding/database/db_descriptor.cpp`).

On a plain build the freed slot usually reads back benign, which is why it only bit intermittently in production. It needs no worker or GC — a plain main-thread `db.close()` + natural process exit reproduces it, because the ref lifetime is owned by the tsfn, not the finalizer.

## Reproduction

A macOS-only, Guard-Malloc-gated child-process test reproduces it deterministically (SIGSEGV, 5/5):

- `test/fixtures/fork-user-shared-buffer-shutdown.mts` — open a DB, get user-shared buffers **with callbacks**, `db.close()`, keep the buffers referenced, exit naturally so `Environment::RunCleanup` finalizes them.
- `test/user-shared-buffer-shutdown.test.ts` — runs the fixture under Apple's Guard Malloc (`libgmalloc`) so the UAF faults; asserts a clean exit.

## Fix direction

Stop the buffer finalizer from dereferencing the tsfn-owned `callbackRef`. Remove the associated listener by identity (a lifetime-tracked handle to the `ListenerCallback`) rather than re-resolving a borrowed `napi_ref`, so the finalizer never touches a ref it does not own.

## Related

- #786 (Audit DBHandle N-API references reachable from cross-environment shutdown)
- #741 (Worker-env teardown destroys transactions on the shared DBDescriptor)

kriszyp:
## What

`flush()` and `flushSync()` now take an optional `{ allowWriteStall }` bag, mapping to `rocksdb::FlushOptions::allow_write_stall`. The default is unchanged.

`DBDescriptor::flush()` constructed a default `rocksdb::FlushOptions`, and that was the only occurrence of the type in the binding, so every flush ran with `allow_write_stall = false` and no caller could say otherwise. False is the *waiting* behavior — despite how the name reads, it means the flush waits until it can run without causing a write stall — and that wait has no timeout at any layer. It is taken on the calling thread, which for the async `flush()` is a libuv worker, so a database sitting in a stall condition hands its caller a promise that never settles while the event loop stays alive, with no error and nothing logged. The stall conditions that produce it are ordinary: immutable-memtable backlog, the L0 stop trigger, `soft_pending_compaction_bytes_limit`, or a WriteBufferManager budget that cannot drain.

A caller whose flush is a durability gate it is already blocked on would rather stall writers for the duration than wait a stall out. That is the situation in [harper-pro#678](https://github.com/HarperFast/harper-pro/pull/678), where the base-copy cursor flush is awaited from the copy's `onCommit`.

This adds an option; it changes no existing behavior. `DBDescriptor::close()` is the one internal caller and **keeps the waiting default** — see the CI finding below for why that is deliberate rather than an oversight.

## Also fixes #774

Same class of defect one layer down, found while tracing the above. `Database::Flush` and `Database::Compact` returned without invoking **either** `resolve` or `reject` when the database was opened read-only, so `await db.flush()` there was an unconditional permanent hang — no stall required. They now resolve, matching the no-op `flushSync` already performed. Recorded as AGENTS invariant 12: a callback-style native method owes its caller exactly one settled callback on every path, and for a promise an early `return` is not a no-op.

Argument validation in all three methods now precedes the read-only short-circuit, so the same argument gets the same verdict in either mode. In `Compact` it also precedes the `AsyncCompactState` allocation, so a rejected key buffer can no longer throw past the `new` and leak it.

## Verification

Route **(b) new test file** plus **(a) extension of the suite that owns this area**.

- `test/flush-options.test.ts` (new) — the option is accepted and flushes durably on both entry points; no-args / `{}` / explicit `false` still work; a non-object bag or non-boolean value throws rather than being silently ignored, in both read-only and read-write mode; the two read-only settle cases from #774; and the compact reorder, exercised through the native handle since `RocksDatabase.compact` always materializes real `Buffer`s and can't reach the validation.
- **Fails-on-base, verified mechanically:** reverting only the two read-only settle blocks and rebuilding makes both read-only cases fail with `Test timed out in 10000ms`. A regression here is a hang rather than an assertion failure — the promise never settles and no in-test timer can surface it — so the per-test timeout *is* the assertion. Restored and rebuilt afterward.
- `test/write-buffer-manager-stall.test.ts` extended by one case, since it is the only file that builds a stalling `WriteBufferManager` (`allowStall` is fixed when the singleton is created, so it needs its own process). It is labelled plumbing-only on purpose — see decision 4.
- Full suite **762 passed / 2 skipped / 56 files**, Node 26.2.0 on Linux x64, RocksDB 11.8.1. `pnpm check` (type-check + oxlint + oxfmt) clean.
- CI at the pre-rebase head (0c71a3e7): all checks green, across Node 22/24/26, Bun, and Deno on Linux/macOS/Windows, plus the native C++ suites. The one earlier failure and its resolution are described above. See the rebase note below for CI at the current head.

## What CI caught, and what it means

An earlier revision of this branch had `close()` opt in (`flush(true)`), on the reasoning that a closing database has stopped accepting work so a stall costs it nothing. **CI refuted that**, and it is the most useful thing to know about this change.

`Test on Bun (windows-latest)` failed deterministically — both in-job retries, while the same job is green on nine recent runs from other branches. No test reported a failure; a vitest worker died. Diffing per-file test counts against the passing Bun/ubuntu job located it: `transaction-log.test.ts` ran 30 of 70 tests and died at *"should write to same log from multiple workers"*, which drives concurrent `worker_threads` writers against one shared DB path while looping `purgeLogs({ destroy: true })`.

Mechanism: `allow_write_stall = true` makes the flush switch memtables **immediately** rather than waiting for a stall-free moment, so it fires `OnFlushBegin`/`OnFlushCompleted` (`db_descriptor.cpp:189`) into `TransactionLogStore`s that the concurrent purge is destroying. Reverting only that line turned CI fully green, which is the confirmation.

Two things follow, and they matter more than the option itself:

- **The reasoning was too narrow, not just the code.** "Close has stopped accepting work" is true of the *closing handle*; the descriptor is process-global and shared across `worker_threads`, so other handles are still live. Any future attempt to stop close from wedging has to avoid flushing into that teardown race — recorded in AGENTS invariant 13.
- **The public option does not appear to inherit this.** I chased that down: `TransactionLogEventListener` is registered in `dbOptions.listeners` (`db_descriptor.cpp:969`), so those callbacks fire on *every* flush — including RocksDB's own background flushes, continuously, in normal operation. The option changes **when** a manual flush runs, not whether the callbacks happen, so it is not a new crash class. What made close special is that it flushed *during teardown*, when the stores are being destroyed. Moderate confidence: reasoned from registration and the listener's `weak_ptr` discipline, not from a constructed race.

## Open concern, not fixed here

`flushBeforeBackup` flushes inside RocksDB's `BackupEngine`, which builds its own default `FlushOptions` this layer cannot reach — the same unbounded wait, and it is taken **after** `runCreateBackup` acquires the exclusive `.backup.lock`. A stalled database therefore turns a backup into an indefinite hang that also blocks every other backup/`delete`/`purge` on that directory, cross-process and cross-container, until the process dies. Pre-existing and not introduced here; recorded in AGENTS invariant 13 as a known uncovered path and worth its own issue.

## Rebase (2026-08-24)

Rebased onto `origin/main` (1da88746, 25 commits ahead). One conflict, in `AGENTS.md`: main had added its own new invariant #12 (transaction-log rotation boundary recovery) at the same numbered slot this branch used for two new invariants (the callback-settle-every-path fix and `allow_write_stall`). Resolved by keeping main's #12 verbatim and renumbering this branch's to #13/#14 — both sides' content preserved, nothing dropped, no other file conflicted. Local full suite re-ran clean after rebuilding: 804 passed / 3 skipped / 0 failed, including this PR's own `flush-options.test.ts` (6/6) and `compaction.test.ts` (8/8).

CI at the rebased head (961361f0) flaked on two jobs, neither touching this diff:
- `Test on Deno (macos-latest)` — failed once (an unrelated `commit-teardown.test.ts` SIGABRT), passed clean on rerun.
- `Test on Node.js 22 and windows-latest` — failed identically on both attempts, in `test/user-shared-buffer.test.ts`'s `User shared buffer use-after-free repro (child process)` (schema-sync mode), crashing with `0xC0000409` (`STATUS_STACK_BUFFER_OVERRUN`) in a spawned child. That file/subsystem is untouched by this diff (`git diff origin/main...HEAD --stat` touches only `AGENTS.md`, the flush/compact C++ and TS surface, and the two new test files). This PR's own pre-rebase head passed this exact job clean three days ago, and this repo's `Test` workflow has a high background flake rate independent of any one branch — 16 of the last 60 runs failed, several with near-total matrix failures across unrelated PRs. Possibly related to the recently-closed [#790](https://github.com/HarperFast/rocksdb-js/issues/790) (a shared-buffer finalizer use-after-free at teardown, closed four days before this rebase's base commit) if that fix left residual Windows-specific timing sensitivity, but confirming that is its own investigation and out of scope for a rebase.

### Post-rebase review rounds (961361f0 → 1900f2ef)

The rebase changed only `AGENTS.md` (renumbering), but the harness requires a receipt bound to the exact pushed head, so three more review rounds ran against it (codex/gemini + Harper domain, all adjudicated minor) and found real, fixable issues — none of them in the rebase's own diff:

- The `AGENTS.md` renumbering left 5 code comments pointing at the old invariant numbers (12/13 instead of 13/14) — fixed, and the ones that narrated patch history rather than stating the invariant were trimmed at the same time.
- `allowWriteStall` shipped with no README entry, undercutting its purpose as an operator escape hatch — added to `db.flush()`/`flushSync()`, mirroring the `compact()` section.
- `CompactSync` still resolved the read-only no-op *before* parsing its start/end/bottommost arguments — the exact asymmetry this PR already fixed for `Compact`/`Flush`/`FlushSync`, just missed in the sync/native-only path. Reordered to match, with parity test coverage.
- The docs described `allowWriteStall: true`'s cost as "stalling writers." Two things were missing: a defaulted (waiting) `flush()` parks a whole libuv threadpool worker, not just its own promise — the pool defaults to 4 threads, so a few stalled flushes stall every unrelated `fs`/`dns`/`crypto` call in the process — and opting in is not a rescue for a flush already in flight (no cancellation) and can itself prolong an L0 stop-trigger condition rather than clear it. Both now documented in the JSDoc, README, and AGENTS invariant 14.
- Added the missing async twin of the invalid-value test (`flush({ allowWriteStall: 'yes' })`); the sync case already existed.

One Gemini "blocker" claim (unsynchronized column-family iteration inside `DBDescriptor::flush`) was raised on a delta round where domain adjudication was auto-pruned, so it went to push ungraded; a follow-up round forced the domain leg and it was formally refuted — `flush()` already snapshots and pins column families under `columnsMutex` before extracting native handles (`db_descriptor.cpp:1866-1877`), unchanged by this PR. Several other Gemini major/blocker claims that round (a `finishClose()` wedge, a segfault racing `purgeTransactionLogs`, a use-after-free in the async flush worker) were adjudicated as pre-existing/already-tracked or not matching the actual code; see the round's artifact for the code-level refutation of each.

## For the human reviewer

1. **The default stays `false` (waiting).** Every existing caller is unchanged, which also means the never-settling promise described above remains the out-of-the-box behavior. The alternative is defaulting Harper's `flush()` to `true` and making the unbounded wait opt-in. Flipping later is a one-line change but a semver-visible behavior break. A "no" here costs nothing today and closes off the argument that upstream's default is a trap we're propagating.
2. **Is my reading of the teardown race right?** I concluded the public option does not inherit the crash, because the flush listeners already fire on every background flush — so the option shifts timing rather than introducing callbacks. That is reasoning, not a constructed race, and it is the load-bearing claim for shipping the option at all. If you think a caller can drive `flush({ allowWriteStall: true })` into a concurrent `purgeLogs({ destroy: true })` in a way a background flush cannot, this needs a guard before merge.
3. **Option parsing lives in C++**, with a string-matched error, rather than in `src/database.ts` where the type already exists. That costs error-message quality and duplicates the surface for every future flush option. Cheap to move, but the tests currently assert the native message.
4. **The behavioral half is deliberately unproven.** No test shows `allowWriteStall` changes RocksDB's behavior, so nothing would catch the flag being dropped between TS and `FlushOptions`. Building a deterministic stall fixture is possible, but #755 removed the stall this fixture could produce on demand, and the failure mode of such a test is a wedged libuv thread rather than a red assertion. If you want the coverage anyway, that is a reasonable call to overrule me on. (Reconfirmed by the reconciliation round below, merged with the same finding from Gemini: deleting or inverting `flushOptions.allow_write_stall = allowWriteStall` at `db_descriptor.cpp:1885` leaves every test green.)
5. **The compact reorder rides in this PR** rather than a separate one — it is the same finding as the flush validation reorder, but it does widen review scope and revert granularity.
6. **Narrative comments remain in two test files** (`test/flush-options.test.ts:4-7,46-47,86-88`, `test/write-buffer-manager-stall.test.ts:50-53`), flagged nit-level by the reconciliation round below. Left as-is rather than spending another review round on a comment-only diff; trivial to clean up in a follow-up.

Worth knowing while reading: opting a flush in is **database-wide**. `DBDescriptor::flush` covers every column family on a descriptor that is process-global and shared across `worker_threads`, so the stall reaches every other table and every other handle on that path, not just the caller's. The JSDoc says so now; it did not in the first draft, and it is the thing most likely to surprise a call site.

### Receipt reconciliation round (1900f2ef)

The three post-rebase rounds above ran from a mirror worktree under the MIRROR BRANCH dispatch protocol and, by an oversight in that session, never passed the review CLI's `--branch kris/flush-options` flag — so every receipt they produced was correctly bound to the exact reviewed SHA but keyed to the mirror's own branch name instead of this PR's. The settlement gate reads receipts by (repo, branch, sha) and never found a match, holding the task twice on an already-covered head even though real independent review had run at every pushed commit through `d6ee7a64`.

One more delta round (codex, low-effort/auto-pruned domain+Cursor — a narrow, low-risk delta by the CLI's own leg policy) ran with the correct `--branch`, carrying forward the existing `d6ee7a64` review and working only the intervening docs/comment-only commit. It reconfirmed the already-disclosed test-coverage gap (item 4 above) and flagged one new nit (item 6 above); nothing else changed. No code changes resulted — this round exists to correctly key the receipt to `kris/flush-options` at `1900f2ef`, not to add review content.

### PR-feedback round (5f25e8e0)

Two review threads from Barber AI, both documentation-only, both confirmed against the code before fixing:

- **The `readOnly` docs contradicted this branch's own behavior.** They said write operations throw `ERR_DATABASE_READONLY`, but `flush()`/`compact()` now resolve as no-ops and `flushSync()`/`compactSync()` return undefined ([`Database::Flush`](https://github.com/HarperFast/rocksdb-js/pull/775/changes?w=1#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R637), and the matching branch in `FlushSync`/`Compact`/`CompactSync`). A caller following the docs expected a rejection and got a silent success with no signal that nothing was flushed. The exception is now stated on the `readOnly` bullet and in all four method sections in the README, plus the JSDoc on `RocksDatabase.flush`/`flushSync`/`compact`/`compactSync` and on `StoreOptions.readOnly` / `Store.readOnly`.
- **`flushSync()` deferred to `flush()` for the `allowWriteStall` hazard, which understated it.** The default wait is taken on the *calling* thread; for the sync entry point that is the JS thread, so it freezes the event loop outright rather than parking one of four libuv pool workers — and it is held under `ACQUIRE_OPERATIONS_LOCK`'s `OperationGuard`, whose `operationsInFlight` claim `finishClose()` waits on, so the database cannot be closed out from under it either. That is a strictly larger blast radius than the async case the docs framed as the worse one. Now stated in the `flushSync()` README section, its JSDoc, and the `FlushOptions.allowWriteStall` JSDoc.

`compactSync()` on a read-only database was the one settle case the suite did not assert (`flush()`, `flushSync()`, `flush({allowWriteStall:true})` and `compact()` all were); added, so every newly documented contract is covered. Full suite re-run: **804 passed / 3 skipped / 0 failed**, `pnpm check` clean.

One more delta review round (codex; gemini produced no output, Cursor and domain legs auto-pruned as a narrow low-risk delta) found nothing new — it restated items 4 and 6 above, both already-disclosed decisions. Item 6 in particular: on re-reading, the four flagged comments are non-obvious rationale rather than narration (why a hang means the per-test timeout *is* the assertion; why the malformed-key cases have to reach through `store.db`; why the stalling-WBM case is plumbing-only), so they are left as the author decided.

## Review coverage

Pre-rebase: four rounds. Round 1 (codex + gemini + Harper domain, adjudicated) returned CHANGES on three findings, all fixed: the close path inheriting the waiting default, mode-dependent argument validation, and a stall test that claimed behavioral coverage it did not have. Round 3 (full, adjudicated) settled at severity **minor** and produced the decisions above. Round 4 (delta) is codex **LGTM**.

Post-rebase: three more rounds against the rebased head (961361f0), all adjudicated **minor**, findings and fixes described above. The graded (opposite-family) leg timed out on the last round; gemini + Harper domain still produced independent coverage there. `cursor-composer`/`cursor-grok` were pruned or refused every round (Cursor refuses a diff that touches `AGENTS.md`).

Honest degradations: `cursor-grok` and `cursor-composer` were auto-pruned on every round; the gemini leg produced no output on pre-rebase round 2, and no graded (codex) leg on the final post-rebase round (timeout). Gemini raised several findings across rounds that were refuted rather than fixed — pre-rebase: three claiming uninitialized `argv` slots (`napi_get_cb_info` pads unsupplied slots with `undefined`, and `Compact`/`Get` have used the pattern since before this PR), and one claiming an async-flush use-after-free (`AsyncFlushState` holds a `shared_ptr<DBHandle>` whose `descriptor` is itself a `shared_ptr`, and close gates on `registerAsyncWork`); post-rebase: the column-iteration "blocker" and other majors/blockers named above, all refuted with code citations in the domain adjudication. The review-need grade below reflects the last round that ran with a graded/domain leg selected; the final commit (comment cleanup + docs only) did not need a further round per the review policy's own tiering (trivial changes need no re-run).

Generated by Claude Opus 5; rebase and post-rebase review rounds by Claude Sonnet 5.

<sub>Review-Coverage: authored=claude; ran=codex,gemini; adjudicated=domain; blocked=cursor-grok(no-receipt); declined=cursor-composer; rounds=7 @ 49242df54a01</sub>

<sub>Human-Review-Need: 3 (decisions: readonly-noop-resolves-vs-rejects, close-path-keeps-waiting-default, flag-default-false-follows-rocksdb, option-bag-shape-and-name, docs-carry-the-hazard-not-the-code) @ 49242df54a01</sub>


src/binding/database/database.cpp
```diff
@@ -592,19 +618,30 @@ napi_value Database::FlushSync(napi_env env, napi_callback_info info) {
...
+	if (getFlushOptions(env, argv[2], allowWriteStall) != napi_ok) {
+		::napi_throw_type_error(env, nullptr, "Flush options must be an object with an optional boolean allowWriteStall");
+		return nullptr;
+	}
+
 	if ((*dbHandle)->descriptor->readOnly) {
+		napi_value recv;
+		NAPI_STATUS_THROWS(::napi_get_undefined(env, &recv));
+		napi_value ignored;
+		NAPI_STATUS_THROWS(::napi_call_function(env, recv, resolve, 0, nullptr, &ignored));
```
cb1kenobi, line 641: **Low: read-only `flush()`/`compact()` resolve silently, but the README says read-only writes throw**

This branch settles the promise as a no-op on a read-only database (same for `compact()`, and `flushSync()`/`compactSync()` return undefined). The `readOnly` option docs still say write operations "will throw an error with code `ERR_DATABASE_READONLY`" (README.md:109-111), and none of the four method sections mention the read-only case. A caller following the docs expects a rejection and instead gets a silent success with no signal that nothing was flushed.

Suggested fix: add a line to the `readOnly` bullet — and to `db.flush()`/`db.flushSync()`/`db.compact()`/`db.compactSync()` — stating these are no-ops that resolve rather than throw on a read-only database.

Static review only: confirmed by reading the code path and this PR's own tests, not by exercising the documented contract.

—
_Generated by Barber AI_

README.md
```diff
@@ -371,21 +371,50 @@ db.dropSync();
...
+await db.flush({ allowWriteStall: true });
 ```
 
-### `db.flushSync(): void`
+### `db.flushSync(options?): void`
 
 Flushes all in-memory data to disk synchronously. Note that this can be an expensive operation, so
 it is recommended to use `flush()` if you want to keep the event loop free.
 
+- `options: object` Same as [`flush()`](#dbflushoptions-promisevoid).
```
cb1kenobi: **Low: `flushSync()` defers to `flush()` for options, but the default-wait hazard is larger on this path**

The `allowWriteStall` bullet this points at frames the unbounded default wait in async terms — "the returned promise never settles", "`flush()` runs on the libuv threadpool (default size 4)". On `flushSync()` the same unbounded wait is taken on the JS main thread, so it freezes the event loop outright rather than parking a pool worker, and it is held while `ACQUIRE_OPERATIONS_LOCK`'s `OperationGuard` keeps an `operationsInFlight` claim that `close()` blocks on. That is a strictly larger blast radius than the async case, yet the section above only warns that it "can be an expensive operation" (README.md:409-410). AGENTS invariant 14 does say the wait is "taken on the calling thread"; the user-facing docs do not.

Suggested fix: note here that under the default `allowWriteStall: false` the wait blocks the calling (main) thread, so the async warning applies more strongly, not less.

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `5f25e8e0` — re-review of the incremental diff since `1900f2ef` (one new commit, docs + test only). Both prior Low findings are fixed: the read-only no-op contract is now documented on all four entry points, and `flushSync()`'s larger stall hazard (blocks the JS thread, holds the `operationsInFlight` claim `close()` waits on) is now called out. No new issues. This PR looks good, nice job!

—
_Generated by Barber AI_

AGENTS.md
```diff
@@ -538,6 +538,49 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     physical truncation when the scan detects a torn tail, but its pre-extended zero padding makes an
     entry with a durable header and partially durable payload look complete; detecting that case needs
     a payload checksum. Recovery runs before mappings can be handed to readers.
+14. **A callback-style native method owes its caller exactly one settled callback on every path**:
```
cb1kenobi: **Medium: duplicate invariant number 14 breaks `pnpm check` (confirmed CI-red)**

Merging `main` (which independently landed a new invariant 14, "A recovered active transaction-log file ends on a transaction boundary...") in alongside this PR's own invariant 14 ("A callback-style native method owes its caller exactly one settled callback...") produced two consecutive list items both numbered `14.`, with this branch's own `15.` (`allow_write_stall`) now following the second `14.` instead of becoming `16.`.

`oxfmt --check` renumbers ordered lists and disagrees with the checked-in numbering, so `pnpm check` fails locally and in CI — verified live: the `Check` job on this PR (run 32863379874) fails at `oxfmt --check` on `AGENTS.md` for exactly this reason. This is the same duplicate-invariant-numbering class that made rocksdb-js#768 CI-red.

Suggested fix: renumber this entry to `15.` and the `allow_write_stall` entry below it to `16.` (running `oxfmt` locally without `--check` does this automatically), then grep the file for any cross-references to "invariant 15" that need to shift to 16.

—
_Generated by Barber AI_

cb1kenobi:
Re-reviewed at `a9de6267` — this is a clean merge of `main` into the branch (verified: `git merge-tree --write-tree` of the two parents matches the merge commit's tree exactly, so no manual edits were folded in beyond the mechanical merge). All prior findings still fixed; the `MUST STAY TRUE` invariants from the last review all still hold at this head:

- `allowWriteStall` default (`false`) preserved on every path: `flush(bool allowWriteStall = false)`, the bare `this->flush()` in `finishClose()`, and `db->Flush()` is called from exactly one place in the whole binding (`DBDescriptor::flush()`) — every other flush-adjacent path (backup/checkpoint) runs entirely inside RocksDB's own internals and structurally cannot reach the flag.
- Validation-before-readonly ordering is consistent across all four entry points (`Compact`, `CompactSync`, `Flush`, `FlushSync`).
- The "threadpool blast radius" doc claims check out: `napi_create_async_work` is used for the async path, `flush()` covers every column family snapshotted under `columnsMutex`, and `CommitWorker` owns one lazily-started `std::thread` per instance.

One new issue from the merge itself (posted inline on `AGENTS.md:541`): merging in `main`'s own new invariant 14 alongside this branch's invariant 14 left a duplicate `14.` in the numbered list, which `oxfmt --check` (and therefore `pnpm check`) rejects. **This is confirmed live** — the `Check` job on this PR is currently failing for exactly this reason (run [32863379874](https://github.com/HarperFast/rocksdb-js/actions/runs/32863379874/job/97852591293)), the same duplicate-invariant-numbering class that made rocksdb-js#768 CI-red.

Separately, and lower confidence: `Test on Node.js 26 and ubuntu-latest` is also currently red on this PR, failing `test/verification-table.test.ts > VT identity across DB open/close lifecycle (#1864) > a version cached before close never leaks into a later incarnation (non-resurrection)` with `expected 134217728 not to be 134217728`. That test is explicitly probabilistic by design (40 iterations to make allocator address-reuse "very likely" rather than certain) and touches VT epoch machinery this PR's own diff doesn't change, so it looks inherited from the `main` merge rather than caused by this branch — I could not reproduce it locally (macOS, 1 run, passed). Worth a CI re-run before relying on green, since it's currently failing on the actual head.

`pnpm test:native` and the local `destroy`/`flush`-adjacent suites otherwise pass clean.

—
_Generated by Barber AI_

cb1kenobi (approved):
Love this!

AGENTS.md
cb1kenobi: Verified fixed at `c71e18ac` ("Fix formatting"): the branch's invariants renumbered 14→15, 15→16, no duplicate remains. `oxfmt --check` passes clean locally (182 files) and the `Check` job is green at this exact head. Resolving.

—
_Generated by Barber AI_

cb1kenobi:
Re-reviewed at `c71e18ac` (delta from `a9de6267`: one commit, "Fix formatting", touching only `AGENTS.md`).

Merge-base check: `a9de6267` and `c71e18ac` share the same merge-base with `origin/main` (`c577bc4e`), and `a9de6267` is a direct ancestor of `c71e18ac` — this is a plain fast-forward push, not a rebase or merge, so no tree-level merge verification was needed.

The one open Medium (duplicate `14.` invariant in `AGENTS.md` breaking `oxfmt --check`/`pnpm check`) is fixed: renumbered to 15/16, no duplicate remains. Confirmed both locally (`oxfmt --check` clean, 182 files) and via CI — the `Check` job is green at this exact head (`c71e18ac`). Resolved on the thread.

The separately-flagged probabilistic VT-identity test (`Test on Node.js 26 and ubuntu-latest`) is also green at this head, consistent with the earlier read that it was inherited main flakiness rather than caused by this PR.

CI at this head: all non-Windows jobs pass, including `Check` and the previously-red VT-identity job. Four Windows jobs (`Test on Node.js 22/24/26 and windows-latest`, `Test on Bun windows-latest`) are still pending as of this review — not yet green, reporting as pending rather than assuming pass.

`MUST STAY TRUE` invariants re-verified by reading the rebuilt source at `c71e18ac` (this delta touches only `AGENTS.md`, so these were unaffected, but re-checked rather than assumed):
- `allowWriteStall` default preserved: `flush(bool allowWriteStall = false)` (`db_descriptor.h:551`), bare `this->flush()` in `finishClose()` (`db_descriptor.cpp:424`), and `db->Flush()` is called from exactly one place in the binding — backup/checkpoint paths use RocksDB's own internal `FlushOptions` and structurally cannot reach the flag.
- Validation-before-readonly ordering is consistent across all four entry points (`Compact`, `CompactSync`, `Flush`, `FlushSync` in `database.cpp`) — argument parsing happens before the `readOnly` short-circuit in every case.
- Threadpool blast-radius doc claims check out: `napi_create_async_work` is used for the async path, `flush()` snapshots every column family under `columnsMutex` (`db_descriptor.cpp:2042-2068`), and `CommitWorker` owns one `std::thread` per instance (`commit_worker.h:40,65`).

No new findings this pass.

Worth flagging as a repo-level pattern, not an author mistake: this is the third same-class CI-red in one day — rocksdb-js#768 (duplicate `12.`), this PR's earlier head (13/14→14/15 renumber), and now this (duplicate `14.` again, from a `main` merge). `main` inserting a new AGENTS.md invariant silently forces a renumber on every open branch that also added one, and `oxfmt` enforces the numbering. Worth a structural fix (e.g., a stable ID scheme instead of sequential numbers) rather than relying on every branch catching it at merge time.

—
_Generated by Barber AI_

PR #777, opened 2026-08-13
chore: replace tsx with Node native type stripping

cb1kenobi:
## What

Removes `tsx` in favor of Node's built-in TypeScript type stripping. Every `.ts`/`.mts` file the project runs directly — build/test scripts, worker threads, and spawned test fixtures — now executes under native stripping. `tsx` is dropped as a devDependency.

## Why this is safe now

- The `engines` floor is `^22.18.0 || >=24.0.0` — exactly the range where type stripping is **on by default (unflagged)**. CI runs Node 22/24/26.
- `erasableSyntaxOnly` was already enforced, so the code was already restricted to strippable syntax (no enums/parameter properties).

## Changes

- **Scripts / build**: `coverage`, `coverage:native`, `test:native`, `run.mjs`'s inner spawn, `deps/gtest.gyp`, and `configure-rocksdb.mjs` now invoke `node` directly.
- **`init-rocksdb`**: local imports carry real `.ts` extensions (native strip does not remap `.js`→`.ts` or resolve extensionless). `tsconfig` gains `allowImportingTsExtensions` + `noEmit` so `tsc --noEmit` accepts them; `tsdown` ignores `noEmit` and still emits `dist`.
- **Worker `.mts` (15) + benchmarks**: repointed from `../../src/index.js` to the built `dist`, and the bootstrap helpers now use a plain native `import()` instead of `tsx`'s `register()`. Loading `src` natively would cascade into its bundler-style `.js`-extension import graph that Node can't resolve.
- **Fork-fixture path**: six test files spawned child processes via `node_modules/tsx/dist/cli.mjs`. Fixtures now load `dist`, and `createWorkerBootstrapScript` was extracted to a `src`-free `test/lib/worker-bootstrap.ts` (re-exported from `util.ts`) so fixtures resolve natively without pulling `util`→`src`.
- **Docs**: added an `AGENTS.md` note on the native-strip constraints for future contributors.

## One caveat

`tsx` still appears in `pnpm-lock.yaml` as a transitive **optional peer** of `tsdown`/`vite`/`vitest` — pnpm resolves it regardless of our usage (a from-scratch lockfile keeps it), and it is no longer a direct dependency (`pnpm ls tsx` is empty). Forcing it out entirely would require an override that risks other peers; not worth it. The project no longer *invokes* `tsx` anywhere. The lockfile diff was kept surgical (only the 3-line direct-dependency entry removed).

## Verification

- Full Vitest suite: **756 passed, 1 skipped**
- `test:native` (GoogleTest): **101 passed, 3 skipped** — exercises the migrated `run.mjs` + gyp init-gtest invocation
- `type-check`, `lint`, `fmt:check`: clean
- `pnpm build:bundle` still emits `dist`; `pnpm install --frozen-lockfile` valid

## For the human reviewer

- Highest-value place to look: the **fork-fixture path** (`test/lib/worker-bootstrap.ts` extraction + the 6 `test/*.test.ts` spawn-arg collapses). The subtle constraint is that anything Node loads directly must avoid `src`'s bundler-style resolution — fixtures load `dist` and only `src`-free helpers.
- The `tsconfig` `noEmit: true` addition is load-bearing for `allowImportingTsExtensions`; confirmed `tsdown` still emits `dist` regardless.
- Worker/fixture tests now exercise the built `dist` wrapper rather than `src`; the underlying native addon is identical per process, so lifecycle/teardown behavior is unchanged (all such tests pass).

Generated by Barber AI (Claude Opus 4.8).


AGENTS.md
```diff
@@ -199,6 +199,25 @@ sufficient (env teardown does not honor tsfn acquire counts); see
...
+      step** (Vitest resolves its own `.js` specifiers; only the native-loaded files must use `.ts`).
+      **Stress tests and benchmarks are the deliberate exception**: their workers import the built
+      `dist` (`stress-test/workers/`, `benchmark/setup.ts`) because they exercise the shipped artifact,
+      not source — so `pnpm test:stress` / `pnpm bench` build first (both CI workflows do).
+  - **Mark type-only imports with `type`.** `verbatimModuleSyntax` is enabled: native strip can only
+    erase `import type` / `import { type X }`, not a value-style `import { X }` that happens to be a
+    type — that would survive to runtime and fail against a module (e.g. `dist`) that never exported
+    the type. `tsc` enforces this.
+  - **Fixture helpers must be `src`-free** only where they'd otherwise pull a heavier graph — e.g.
+    `createWorkerBootstrapScript` lives in `test/lib/worker-bootstrap.ts` and is re-exported from
```
kriszyp: `test/lib/util.ts` no longer re-exports `createWorkerBootstrapScript`; the call sites now import `test/lib/worker-bootstrap.ts` directly. Could you remove “and is re-exported from `test/lib/util.ts`” (or restore an intentional re-export) so this native-strip guidance matches the code?

— KrAIs (GPT-5 Codex)

kriszyp (approved):
🤖 Dispatch codex review (verdict COMMENTS).

AGENTS.md
cb1kenobi: Good catch, thank you! You're right — the re-export was removed once every call site moved to importing `test/lib/worker-bootstrap.ts` directly, so I've dropped the stale clause and clarified that the split from `util.ts` (which imports `src`) is the point. Fixed in a62803d4.

— Generated by Barber AI

PR #778, opened 2026-08-13
feat: statistical range key-count estimation (getEstimatedKeyCount range support + CountEstimator)

### LINKED ISSUE #205: Investigate more efficient range count

kriszyp:
## Summary

Adds a statistical range key-count estimate that never iterates, for query planning and pagination reporting (HarperFast/harper#2147). Closes #205, and supersedes the earlier attempt in #311.

**Native `estimateCount(start?, end?)`** combines three RocksDB primitives:

- `GetApproximateMemTableStats` — returns an entry *count* for the memtable portion directly.
- `GetApproximateSizes` (files only, 10% error margin) — approximate on-disk bytes covered by the range.
- `GetPropertiesOfTablesInRange` — per-SST `num_entries` / `num_deletions` / block sizes for only the SSTs overlapping the range, giving a range-local live-entry density (`(entries − deletions) / file bytes`) that converts bytes → count.

Using range-local table properties (instead of #311's whole-CF cached mean with write-count invalidation) keeps the density honest when entry sizes vary across the keyspace and needs no cache or invalidation hooks. Open-ended ranges are estimated as `estimate-num-keys` minus the complementary range — an empty slice is the *smallest* key, so it must never be passed as an upper bound (a correctness bug in #311's open-ended path). Inverted, empty, and zero-length bounds are guarded (they would otherwise underflow RocksDB's uint64 offset subtraction).

**Public API** (shape chosen by Kris: estimates carry a trust signal):

- `db.estimateCount(options?: RangeOptions)` → **`{ count, confidence }`** (`CountEstimate`). `confidence` is a heuristic 0–1 trust indicator, exactly 1 only when the count is exact — derived natively from the estimate's resolution (data-block/memtable-sampling granularity relative to the count), the tombstone fraction of overlapping SSTs, complement-subtraction error for start-only ranges, and *failed statistics calls* (a degraded estimate caps at 0.1; a failed `estimate-num-keys` read returns `{0, 0}`, not a confident empty). `exclusiveStart`/`inclusiveEnd` are honored via the bytewise-successor zero byte.
- `db.getEstimatedKeyCount()` — **unchanged** original no-arg signature (cheap `estimate-num-keys` alias), so existing callers are untouched and one name doesn't cover two cost profiles.
- `db.createCountEstimator(options?)` → `CountEstimator` — rides an iterator: `advance(lastKey, count)` checkpoints progress (e.g. once per page), `estimate()` returns `{count, confidence}` = exact traversed + remainder calibrated by the observed actual/estimated ratio (clamped 8×), memoized per checkpoint; confidence is the exactness-weighted blend, capped at 0.999 until `finish()` declares the traversal complete (then exact with confidence 1). Supports `reverse` and the range bound flags.

## For the human reviewer

- **Cost model (the review's one carried major, accepted + documented):** a bounded estimate enumerates table properties for every SST overlapping the range, synchronously on the JS thread — table-property reads go through the table cache and can do I/O for cold files, and a start-only range does the work of its *complement*. Inherent to the range-local-density design (the cached whole-CF mean alternative is what made #311 wrong); an async variant is additive later. Callers should prefer bounded ranges.
- **`confidence` is now API surface**: callers will encode thresholds against it, so retuning the formula changes their behavior (review ledger point). The semantics doc deliberately promises only "heuristic ordering signal, 1 = exact" — the formula itself is not contract.
- **Whole-DB vs full-range disagreement** (ledger): `estimateCount()` no-bounds uses `estimate-num-keys` while a bounded full range uses bytes×density; they can disagree. Deliberate — the no-bound path stays O(1) and matches `getEstimatedKeyCount()`.
- **Calibration constants** (`CALIBRATION_MIN_TRAVERSED = 16`, 8× clamp) are judgment calls; options can be added later.
- **Caller-owned progress contract:** `advance()` trusts the caller (monotonic cursors, no double-reporting); a wrapping-iterator variant would be additive.
- Estimates deliberately ignore transaction state (committed statistics only; covered by a test).

## Verification

- `test/estimate-count.test.ts` (12 tests): flushed / memtable-only / mixed ranges, open-ended both sides, empty DB (confident 0), inverted range (exact 0), zero-length native bounds, uncommitted-transaction exclusion, monotonic scaling with range width, estimator refinement (confidence must increase), reverse iteration, `finish()` exactness, and a paginated loop driven to completion (pre-finish confidence < 1, exact total after `finish()`). Full suite green at every commit (latest: 767 passed / 2 skipped, 56 files).
- Accuracy/perf (500k entries, varied value sizes 20–220B, flushed): counts within **−0.1% to −5.5%** of exact across full/half/tenth/1% ranges at **9–83µs** vs **1–85ms** exact scans (~1000×). Confidence measured: **0.999** full/half ranges, **0.88** at 1%, **0.21** on a 50-key range (~2× over-report, correctly distrusted), **0.13** on a start-only tail (+39% error from complement subtraction, correctly distrusted). Estimator converged 4.2% → 1.4% error by 50% traversal. (Loose 2× bounds asserted in CI; these are measured local numbers.)
- Downstream validation: the harper planner integration (follow-up PR) was run end-to-end against this branch linked into a harper worktree — real tables, secondary indexes (composite `[value, primaryKey]` keys through `RocksIndexStore`), width-ordered estimates confirmed.
- Native GoogleTest target untouched (N-API-layer change; the vitest suite is end-to-end through the real binding).

## Review coverage

Generated by Claude (Fable 5). Cross-model pre-push review via `prepush-review.mjs`, six rounds:

- **Round 1 (full, `1c4be8d`):** Codex (graded) + Gemini + Harper-domain adjudication — cursor-composer failed (output-format rejection), cursor-grok pruned. Fixed: inverted-range guard, `exclusiveStart`/`inclusiveEnd`, estimator forward off-by-one, `finish()`, memoization, tempered cost claims.
- **Round 2 (delta, `4481bc4`):** Codex (resumed) + Gemini. Fixed: zero-length end bound bypassing the guard on the native surface (napi nullptr for empty buffers).
- **Round 3 (delta, `0ddaf72`):** Codex + Gemini — hardening confirmed clean.
- **Round 4 (full, `b2cb53a`, API change):** Codex (graded) + Gemini + Harper-domain — cursor-composer failed again, grok pruned. Adjudicated major fixed in round 5: failed statistics reported as confident estimates.
- **Round 5 (delta, `0548d42`):** Codex (Gemini leg returned no output this round). Follow-ups fixed in round 6: a successful zero `estimate-num-keys` claimed exactness; partial null table-properties collections didn't degrade.
- **Round 6 (delta, final head `89d437b`):** Codex + Gemini — verdict COMMENTS, both fixes confirmed, no new findings.

Accepted, not changed (with rationale): silent-degrade carries low confidence instead of an event; sync-on-JS-thread cost model (documented); caller-owned progress contract; calibration reads *current* state so concurrent writes behind the cursor can swing a checkpoint (inherent to statistical estimates on a live database, bounded by the 8× clamp); no native GoogleTest for the estimator math (it lives in an N-API translation unit, which cannot link into the gtest target — the vitest suite covers it end-to-end through the real binding).

<sub>Human-Review-Need: 3 @ 75562fc26371</sub>

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,7 +447,7 @@ Estimates the number of keys in the database, or within a key range, returning
 from RocksDB statistics (memtable stats plus approximate SST sizes converted through the entry
 density of the SSTs overlapping the range), so its cost scales with the number of SSTs overlapping
 the range rather than the number of keys. Reading cold table properties can do I/O through the
-table cache, so bounded ranges are preferable. A start-only range is computed as the
+table cache, so narrow ranges are preferable. A start-only range is computed as the
 whole-database estimate minus the complement, so it does the work of the range _below_ `start`.
 Accuracy improves with range size (resolution is bounded by SST data-block granularity, so tiny
 ranges over-report), and recently deleted or overwritten entries may be counted until compaction.
--- a/README.md
+++ b/README.md
@@ -434,11 +434,14 @@ the `expectedVersion` option is used.
 Retrieves the estimated number of keys in the database, or within a key range when one is given.
 Unlike `getKeysCount()`, this never iterates: the estimate is derived from RocksDB statistics
 (memtable stats plus approximate SST sizes converted through the entry density of the SSTs
-overlapping the range), so it stays fast regardless of range size — typically microseconds where an
-exact count takes milliseconds. Accuracy improves with range size (resolution is bounded by SST
-data-block granularity, so tiny ranges over-report), and recently deleted or overwritten entries
-may be counted until compaction. Estimates always reflect committed state; writes pending in a
-transaction are not included.
+overlapping the range), so its cost scales with the number of SSTs overlapping the range rather
+than the number of keys — typically microseconds where an exact count takes milliseconds, though
+reading table properties for cold files can do I/O through the table cache. A start-only range is
+computed as the whole-database estimate minus the complement, so it does the work of the range
+_below_ `start`. Accuracy improves with range size (resolution is bounded by SST data-block
+granularity, so tiny ranges over-report), and recently deleted or overwritten entries may be
+counted until compaction. Estimates always reflect committed state; writes pending in a
+transaction are not included. An inverted range (`start` ≥ `end`) returns 0.
 
 ```typescript
 const estimated = db.getEstimatedKeyCount();
@@ -453,8 +456,10 @@ range. Before any traversal, `estimate()` returns the pure statistical estimate 
 `getEstimatedKeyCount(range)`). As the caller reports progress with `advance(lastKey, count)`
 (e.g. once per page), `estimate()` returns the exact traversed count plus a statistical estimate
 of the remainder, calibrated by the observed ratio of actual-to-estimated entries over the portion
-already traversed — so the estimate converges toward the exact total as iteration proceeds. Set
-`reverse: true` when iterating from `end` toward `start`.
+already traversed — so the estimate converges toward the exact total as iteration proceeds. When
+traversal completes, call `finish()` and `estimate()` returns the exact count. Set
+`reverse: true` when iterating from `end` toward `start`. The caller owns the progress contract:
+cursors must move monotonically through the range and each entry must be reported exactly once.
 
 ```typescript
 const range = { start: 'a', end: 'z' };
--- a/README.md
+++ b/README.md
@@ -429,23 +429,41 @@ if (result === constants.FRESH_VERSION_FLAG) {
 Synchronous version of `get()`. Like `get()`, this can return the `FRESH_VERSION_FLAG` sentinel when
 the `expectedVersion` option is used.
 
-### `db.getEstimatedKeyCount(options?: RangeOptions): number`
+### `db.getEstimatedKeyCount(): number`
 
-Retrieves the estimated number of keys in the database, or within a key range when one is given.
-Unlike `getKeysCount()`, this never iterates: the estimate is derived from RocksDB statistics
-(memtable stats plus approximate SST sizes converted through the entry density of the SSTs
-overlapping the range), so its cost scales with the number of SSTs overlapping the range rather
-than the number of keys — typically microseconds where an exact count takes milliseconds, though
-reading table properties for cold files can do I/O through the table cache. A start-only range is
-computed as the whole-database estimate minus the complement, so it does the work of the range
-_below_ `start`. Accuracy improves with range size (resolution is bounded by SST data-block
-granularity, so tiny ranges over-report), and recently deleted or overwritten entries may be
-counted until compaction. Estimates always reflect committed state; writes pending in a
-transaction are not included. An inverted range (`start` ≥ `end`) returns 0.
+Retrieves the estimated number of keys in the database. This is an alias for
+`db.getDBIntProperty('rocksdb.estimate-num-keys')`; use `estimateCount()` for range support and a
+confidence indicator.
 
 ```typescript
 const estimated = db.getEstimatedKeyCount();
-const rangeEstimate = db.getEstimatedKeyCount({ start: 'a', end: 'z' });
+console.log(estimated);
+```
+
+### `db.estimateCount(options?: RangeOptions): CountEstimate`
+
+Estimates the number of keys in the database, or within a key range, returning
+`{ count, confidence }`. Unlike `getKeysCount()`, this never iterates: the estimate is derived
+from RocksDB statistics (memtable stats plus approximate SST sizes converted through the entry
+density of the SSTs overlapping the range), so its cost scales with the number of SSTs overlapping
+the range rather than the number of keys — typically microseconds where an exact count takes
+milliseconds, though reading table properties for cold files can do I/O through the table cache. A
+start-only range is computed as the whole-database estimate minus the complement, so it does the
+work of the range _below_ `start`. Accuracy improves with range size (resolution is bounded by SST
+data-block granularity, so tiny ranges over-report), and recently deleted or overwritten entries
+may be counted until compaction. Estimates always reflect committed state; writes pending in a
+transaction are not included. An inverted range (`start` ≥ `end`) returns
+`{ count: 0, confidence: 1 }`.
+
+`confidence` is a heuristic 0–1 indicator of how trustworthy `count` is — exactly 1 only when the
+count is exact. It is derived from the estimate's resolution (data-block/memtable-sampling
+granularity relative to the count), the tombstone fraction of the overlapping SSTs, and — for
+start-only ranges — the error compounded by complement subtraction. Treat it as an ordering
+signal (e.g. when to trust an estimate for query planning vs fall back to a heuristic), not a
+statistical bound.
+
+```typescript
+const { count, confidence } = db.estimateCount({ start: 'a', end: 'z' });
 ```
 
 ### `db.createCountEstimator(options?: CountEstimatorOptions): CountEstimator`
@@ -453,13 +471,15 @@ const rangeEstimate = db.getEstimatedKeyCount({ start: 'a', end: 'z' });
 Creates an estimator that progressively refines a range count estimate while the range is being
 iterated — useful for reporting a total alongside a page of results without scanning the full
 range. Before any traversal, `estimate()` returns the pure statistical estimate (same as
-`getEstimatedKeyCount(range)`). As the caller reports progress with `advance(lastKey, count)`
-(e.g. once per page), `estimate()` returns the exact traversed count plus a statistical estimate
-of the remainder, calibrated by the observed ratio of actual-to-estimated entries over the portion
-already traversed — so the estimate converges toward the exact total as iteration proceeds. When
-traversal completes, call `finish()` and `estimate()` returns the exact count. Set
-`reverse: true` when iterating from `end` toward `start`. The caller owns the progress contract:
-cursors must move monotonically through the range and each entry must be reported exactly once.
+`estimateCount(range)`). As the caller reports progress with `advance(lastKey, count)` (e.g. once
+per page), `estimate()` returns the exact traversed count plus a statistical estimate of the
+remainder, calibrated by the observed ratio of actual-to-estimated entries over the portion
+already traversed — so the count converges toward the exact total, and `confidence` (the
+exactness-weighted blend of the traversed portion and the remainder's confidence) converges to 1.
+When traversal completes, call `finish()` and `estimate()` returns the exact count with
+confidence 1. Set `reverse: true` when iterating from `end` toward `start`. The caller owns the
+progress contract: cursors must move monotonically through the range and each entry must be
+reported exactly once.
 
 ```typescript
 const range = { start: 'a', end: 'z' };
@@ -471,7 +491,7 @@ for (const { key } of db.getRange({ ...range, limit: 25 })) {
 	pageSize++;
 }
 estimator.advance(lastKey, pageSize);
-const total = estimator.estimate(); // ~total keys in the range
+const { count, confidence } = estimator.estimate();
 ```
 
 ### `db.getKeys(options?: IteratorOptions): ExtendedIterable`
--- a/README.md
+++ b/README.md
@@ -449,8 +449,10 @@ density of the SSTs overlapping the range), so its cost scales with the number o
 the range rather than the number of keys. Reading cold table properties can do I/O through the
 table cache, so narrow ranges are preferable. A start-only range is computed as the
 whole-database estimate minus the complement, so it does the work of the range _below_ `start`.
-Accuracy improves with range size (resolution is bounded by SST data-block granularity, so tiny
-ranges over-report), and recently deleted or overwritten entries may be counted until compaction.
+Accuracy improves with range size. Resolution is bounded by SST data-block granularity, so a range
+narrower than a block is unreliable in either direction: it may over-report or report 0 for present
+keys, and its low `confidence` is the signal. Recently deleted or overwritten entries may be counted
+until compaction.
 Estimates always reflect committed state; writes pending in a transaction are not included. Set
 `reverse: true` to use `getRange()`'s reverse convention (`start` is the upper bound and `end` is
 the lower bound). An inverted range (`start` ≥ `end`) returns
@@ -475,8 +477,10 @@ range. Before any traversal, `estimate()` returns the pure statistical estimate 
 `estimateCount(range)`). As the caller reports progress with `advance(lastKey, count)` (e.g. once
 per page), `estimate()` returns the exact traversed count plus a statistical estimate of the
 remainder, calibrated by the observed ratio of actual-to-estimated entries over the portion
-already traversed — so the count converges toward the exact total, and `confidence` (the
-exactness-weighted blend of the traversed portion and the remainder's confidence) converges to 1.
+already traversed — so the count converges toward the exact total. `confidence` is the
+exactness-weighted blend of the traversed portion and the remainder's confidence, so it approaches 1
+as the exact portion grows, although a checkpoint may decrease when calibration makes a large
+correction.
 Each checkpoint reads committed state, so a traversal performed against a transaction snapshot may
 be calibrated against data committed after that snapshot.
 When traversal completes, call `finish()` and `estimate()` returns the exact count with
--- a/README.md
+++ b/README.md
@@ -440,19 +440,20 @@ const estimated = db.getEstimatedKeyCount();
 console.log(estimated);
 ```
 
-### `db.estimateCount(options?: RangeOptions): CountEstimate`
+### `db.estimateCount(options?: CountEstimateOptions): CountEstimate`
 
 Estimates the number of keys in the database, or within a key range, returning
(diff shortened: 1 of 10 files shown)


test/estimate-count.test.ts
```diff
@@ -0,0 +1,259 @@
+import { dbRunner } from './lib/util.ts';
+import { describe, expect, it } from 'vitest';
+
+/**
+ * Estimates are statistical (block-granular SST approximation + memtable
+ * skip-list approximation), so assertions use a tolerance factor rather than
+ * exact bounds. Uniform fixed-size entries keep the real accuracy well inside
```
cb1kenobi, line 7: **Low: core design claim (range-local density under varying entry sizes) is untested**

The stated advantage of this approach over #311 is that range-local table properties keep the byte-to-count density honest when entry sizes vary across the keyspace. Every test here uses uniform fixed-size entries (this comment says so explicitly), so the one scenario the design exists to handle has no CI coverage: if the SST density computation regressed to a whole-CF mean (the #311 behavior), these tests would still pass.

Suggested fix: add a case that writes large entries across one key sub-range and small entries across another, flush, then assert `estimateCount` over each sub-range tracks that sub-range's actual count (within the existing 2x factor). A whole-CF mean would skew both toward the average and fail.

—
_Generated by Barber AI_

src/count-estimator.ts
```diff
@@ -0,0 +1,159 @@
...
+	#finished = false;
+	#memoized: CountEstimate | undefined;
+
+	constructor(store: Store, options?: CountEstimatorOptions) {
+		this.#store = store;
+		this.#start = options?.start;
+		this.#end = options?.end;
+		this.#exclusiveStart = options?.exclusiveStart ?? false;
+		this.#inclusiveEnd = options?.inclusiveEnd ?? false;
+		this.#reverse = options?.reverse ?? false;
```
cb1kenobi: **High: `CountEstimator` reverse mode inverts the bound convention `getRange` uses**

`Store.getRange` swaps `start`/`end` when `reverse` is set (src/store.ts:921-927), and the native side maps start -> `iterate_lower_bound` / end -> `iterate_upper_bound`. So in reverse iteration `start` is the **upper** bound and `end` the **lower** one: `{ start: 'f', end: 'b', reverse: true }` yields `f, e, d, c` (test/ranges.test.ts:188). The bound flags follow the swapped edges and default to `true`, not `false`.

`CountEstimator` keeps `start` as the lower bound in reverse, so the same options object it is meant to share with `getRange` produces inverted sub-ranges. `estimateCount` returns `{ count: 0, confidence: 1 }` for an inverted range, so `remaining` is 0 and `estimate()` reports `count === traversed` with `confidence: 0.999` — a confidently wrong total, which is exactly the failure mode `confidence` exists to signal.

The reverse test (test/estimate-count.test.ts:198) uses the opposite convention (`start: KEY(0), end: KEY(N)`) and drives `advance()` by hand, so it never crosses a real reverse traversal. A test that pages through `db.getKeys({ ...range, reverse: true })` and asserts convergence would catch this.

Suggested fix — normalize once in the constructor, mirroring `Store.getRange`, and leave the rest of the class unchanged:

```suggestion
		this.#reverse = options?.reverse ?? false;
		// getRange swaps start/end for reverse, so in reverse `start` is the
		// upper bound and `end` the lower one; bound flags default to true.
		this.#start = this.#reverse ? options?.end : options?.start;
		this.#end = this.#reverse ? options?.start : options?.end;
		this.#exclusiveStart = options?.exclusiveStart ?? this.#reverse;
		this.#inclusiveEnd = options?.inclusiveEnd ?? this.#reverse;
```

README.md:480 ("Set `reverse: true` when iterating from `end` toward `start`") is backwards under the same convention and needs the same correction.

—
_Generated by Barber AI_

src/binding/database/database.cpp
```diff
@@ -922,6 +922,224 @@ napi_value Database::GetCount(napi_env env, napi_callback_info info) {
...
+		// SST bytes overlap the range but no usable density: the SST portion
+		// is missing from the count.
+		result.degraded = true;
+		return result;
+	}
+
+	double density = static_cast<double>(entries - deletions) / static_cast<double>(fileBytes);
+	result.sstCount = static_cast<double>(sstBytes) * density;
+	result.count += result.sstCount;
+	result.entriesPerBlock = density * 4096;
```
cb1kenobi: **Low: `entriesPerBlock` hardcodes RocksDB's default 4096-byte block size**

`density` is live entries per *file* byte, so `density * 4096` only means "live entries per data block" while `BlockBasedTableOptions.block_size` is left at its 4096 default (nothing under `src/binding` sets it today). Anyone tuning `block_size` later would silently skew every `confidence` this feeds, with no test or assertion tying the two together. `TableProperties::num_data_blocks` gives the value directly and drops the assumption.

Suggested fix: add `uint64_t dataBlocks = 0;` beside `fileBytes` (line 984), accumulate `dataBlocks += p.num_data_blocks;` in the loop, then:

```suggestion
	result.entriesPerBlock = dataBlocks > 0 ? static_cast<double>(entries - deletions) / static_cast<double>(dataBlocks) : 0;
```

—
_Generated by Barber AI_

src/count-estimator.ts
```diff
@@ -0,0 +1,159 @@
...
+			return this.#store.estimateCount({
+				start: this.#start,
+				end: this.#end,
+				exclusiveStart: this.#exclusiveStart,
+				inclusiveEnd: this.#inclusiveEnd,
+			});
+		}
+		if (this.#memoized !== undefined) {
+			return this.#memoized;
+		}
```
cb1kenobi, line 117: **Low: the pre-traversal `estimate()` path bypasses memoization**

The class doc says results are "memoized per checkpoint, so repeated reads between `advance()` calls are free", but this early return never touches `#memoized`. Before the first `advance()` — the state a paginating caller sits in while rendering page 1 — every `estimate()` call re-runs `estimateCount`, which the same doc notes can do table-cache I/O for cold SSTs. `advance()` already clears `#memoized`, so caching here is safe.

Suggested fix:

```suggestion
		if (this.#memoized !== undefined) {
			return this.#memoized;
		}
		if (this.#cursor === undefined) {
			this.#memoized = this.#store.estimateCount({
				start: this.#start,
				end: this.#end,
				exclusiveStart: this.#exclusiveStart,
				inclusiveEnd: this.#inclusiveEnd,
			});
			return this.#memoized;
		}
```

—
_Generated by Barber AI_

src/count-estimator.ts
```diff
@@ -0,0 +1,159 @@
...
+	get traversed(): number {
+		return this.#traversed;
+	}
+
+	/**
+	 * Records that iteration has advanced through `count` more entries, ending
+	 * at `lastKey`.
+	 */
+	advance(lastKey: Key | Uint8Array, count = 1): void {
+		this.#cursor = lastKey;
```
cb1kenobi, line 81: **Low: `advance()` with an undefined key silently discards the cursor**

`#cursor` is assigned unconditionally, so reporting an empty page resets it to `undefined` while `#traversed` keeps its value — `estimate()` then falls back to the pure statistical estimate and can return a count *below* `traversed`, breaking the invariant the suite asserts at test/estimate-count.test.ts:251. The README example reaches this: `let lastKey;` stays `undefined` when the range yields nothing, and `advance(lastKey, pageSize)` is called unconditionally (README.md:493-495).

Suggested fix — make a no-op advance a no-op:

```suggestion
		if (lastKey === undefined) {
			return;
		}
		this.#cursor = lastKey;
```

—
_Generated by Barber AI_

src/binding/database/database.cpp
```diff
@@ -922,6 +922,205 @@ napi_value Database::GetCount(napi_env env, napi_callback_info info) {
...
+			// so the density denominator must too.
+			fileBytes += p.data_size + p.index_size + p.filter_size;
+			dataBlocks += p.num_data_blocks;
+		}
+	} else {
+		result.degraded = true;
+	}
+	if (entries <= deletions || fileBytes == 0) {
+		// A nonzero byte estimate without density leaves the SST portion unknown.
+		result.degraded = sstBytes != 0;
```
cb1kenobi: **Medium: `degraded` assignment clobbers a failed table-properties read**

This is an assignment, not an OR, so it discards the `degraded = true` set at line 995 (`GetPropertiesOfTablesInRange` returned a non-ok status) and at line 982 (a null entry in the collection). When the properties read fails *and* `sstBytes == 0`, `entries`/`fileBytes` are both 0, this guard fires, and the function returns `degraded == false` — `estimateConfidence` then reports 0.95 for a count built on a statistics call that failed. That is the regression the round-5 "partial null table-properties collections didn't degrade" fix addressed.

It also contradicts the branch directly below (line 1002): with a *usable* density and `sstBytes == 0` the estimate is degraded (confidence 0.1), but with an *unusable* one it is not. The worse-information path returns the higher confidence.

Suggested fix:
```suggestion
		result.degraded = result.degraded || sstBytes != 0;
```

—
_Generated by Barber AI_

test/estimate-count.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			}
+			expect(estimator.traversed).toBe(N / 2);
+
+			// with half the range traversed exactly, the estimate must be at
+			// least the traversed count, within a tighter overall bound, and
+			// more trusted than the untraversed estimate
+			const refined = estimator.estimate();
+			expect(refined.count).toBeGreaterThanOrEqual(N / 2);
+			expectWithin(refined.count, N, 1.6);
+			expectConfidence(refined.confidence);
```
cb1kenobi: **Low: the refinement test no longer asserts what its comment claims**

`expectConfidence(confidence)` defaults to `min = 0, max = 1`, so this assertion is vacuous — `confidence` is in `[0, 1]` by construction. The comment three lines up still says the refined estimate must be "more trusted than the untraversed estimate", and the `expect(refined.confidence).toBeGreaterThan(initial.confidence)` this replaced was the only check of that claim. README.md:487 still promises `confidence` "converges to 1" as traversal proceeds.

The assertion presumably had to go because `calibrationConfidence = Math.min(calibration, 1 / calibration)` (src/count-estimator.ts:153) can push a checkpoint's confidence *below* the pre-traversal estimate whenever calibration corrects hard — an 8x correction caps the remainder's contribution at 0.125. If that non-monotonicity is intended, the README's convergence wording needs the same qualification; either way the exactness floor is still assertable and cheap:

```suggestion
			expectConfidence(refined.confidence, estimator.traversed / refined.count, 1);
```

—
_Generated by Barber AI_

README.md
```diff
@@ -432,13 +432,71 @@ the `expectedVersion` option is used.
...
+
+Estimates the number of keys in the database, or within a key range, returning
+`{ count, confidence }`. Unlike `getKeysCount()`, this never iterates: the estimate is derived
+from RocksDB statistics (memtable stats plus approximate SST sizes converted through the entry
+density of the SSTs overlapping the range), so its cost scales with the number of SSTs overlapping
+the range rather than the number of keys. Reading cold table properties can do I/O through the
+table cache, so narrow ranges are preferable. A start-only range is computed as the
+whole-database estimate minus the complement, so it does the work of the range _below_ `start`.
+Accuracy improves with range size (resolution is bounded by SST data-block granularity, so tiny
+ranges over-report), and recently deleted or overwritten entries may be counted until compaction.
```
cb1kenobi: **Low: "tiny ranges over-report" is the opposite of what a sub-resolution range does**

`GetApproximateSizes` returns 0 bytes for a range narrower than a data block, so `estimateRangeCount` contributes no SST portion at all — for flushed data that is `count: 0`, and this PR additionally marks that case `degraded` (src/binding/database/database.cpp:1002-1004), so it returns `{ count: 0, confidence: 0.1 }`. The new single-key test encodes exactly this by guarding its assertion behind `if (singleKey.count === 0)` (test/estimate-count.test.ts:43).

So a tiny range does not over-report — it under-reports, to zero. A caller reading this paragraph would read a small nonzero result as an upper bound and a zero as "definitely empty"; both are wrong, and only the low `confidence` says so.

Suggested fix: state both directions, e.g. "resolution is bounded by SST data-block granularity, so a range narrower than a data block is unreliable in either direction — it may over-report, or report 0 for a non-empty range; its low `confidence` is the signal".

—
_Generated by Barber AI_

test/estimate-count.test.ts
kriszyp: Addressed in 9c222a3c: the refinement test now asserts confidence is at least the exact traversed fraction, and the documentation explicitly notes that calibration can make an individual checkpoint decrease. — GPT-5 Codex

src/binding/database/database.cpp
kriszyp: Addressed in 9c222a3c: the no-density branch now preserves any earlier degraded state, so a failed or partial table-properties read cannot be promoted back to high confidence. — GPT-5 Codex

README.md
kriszyp: Addressed in 9c222a3c and e9e080a0: the README, native algorithm documentation, and public API docstring now state that sub-block ranges may over-report or return zero for present keys, with low confidence as the signal. — GPT-5 Codex

test/estimate-count.test.ts
```diff
@@ -0,0 +1,335 @@
...
+			expectWithin(full.count, N, 2);
+			// a large uniform range should be high-confidence
+			expectConfidence(full.confidence, 0.5, 1);
+
+			const singleKey = db.estimateCount({
+				start: KEY(N / 2),
+				end: KEY(N / 2),
+				inclusiveEnd: true,
+			});
+			expect(singleKey.count > 0 || singleKey.confidence <= 0.1).toBe(true);
```
cb1kenobi: **Low: the tiny-range assertion still allows a confidently wrong over-report**

`count > 0` short-circuits the whole expression, so the only branch that checks `confidence` is the report-zero one. The README text added in this same push documents the other direction too ("it may over-report or report 0 for present keys, and its low `confidence` is the signal") — a regression that returned `{count: 90, confidence: 0.99}` for this single-key range would pass. The accuracy-or-distrust invariant can be asserted without coupling to block layout (which is what `106ee40d` was avoiding): either the count is roughly right, or confidence is low.

Suggested fix:
```suggestion
			// A sub-block range may be wrong in either direction, but it must not be confidently wrong.
			expect(singleKey.count <= 2 || singleKey.confidence <= 0.5).toBe(true);
```

—
_Generated by Barber AI_

src/count-estimator.ts
```diff
@@ -0,0 +1,171 @@
...
+	 */
+	finish(): void {
+		this.#finished = true;
+	}
+
+	/**
+	 * Estimates the total number of entries in the full range: the exact
+	 * traversed count plus a calibrated statistical estimate of the remainder.
+	 * `confidence` is the exactness-weighted blend of the traversed portion
+	 * (exact) and the remainder's statistical confidence, so it converges to 1
```
cb1kenobi: **Nit: this JSDoc still makes the unqualified "converges to 1" claim the README just walked back**

README:481-486 and `database.ts` were updated in this push to say confidence "approaches 1 as the exact portion grows, although a checkpoint may decrease when calibration makes a large correction" — `calibrationConfidence = Math.min(calibration, 1/calibration)` is exactly that decrease. This doc comment (lines 97-99) is now the only place stating it monotonically converges.

Suggested fix: reword lines 98-99 to match — "so it approaches 1 as the exact portion grows (though a checkpoint can decrease when calibration makes a large correction), and is exactly 1 after `finish()`."

—
_Generated by Barber AI_

test/estimate-count.test.ts
kriszyp: Addressed in 75562fc2: the assertion now requires the exact one-key result or confidence ≤ 0.5, so both a confident zero and a confident over-report fail. — GPT-5 Codex

src/count-estimator.ts
kriszyp: Addressed in 10f4aaa7: the estimate() JSDoc now matches the README and API docs by stating that confidence approaches 1 as exact progress grows, while a calibration correction can lower an individual checkpoint. — GPT-5 Codex

cb1kenobi:
Reviewed `75562fc2` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

Issue #746, opened 2026-08-01
commit-teardown.test.ts crashes with SIGABRT (mutex lock failed: Invalid argument) on Deno macOS CI


kriszyp:
## Summary

`test/commit-teardown.test.ts` ("Async commit completion vs. worker env teardown") crashes
the repro child process with `SIGABRT` on **Deno + macos-latest** CI, with:

```
libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```

This is a native (C++) crash — `pthread_mutex_lock` returning `EINVAL`, which on libc++'s
`std::mutex` means the mutex was already destroyed (or never initialized) when a thread tried
to lock it. It reproduces in the fixture's normal worker-terminate-mid-commit-stream scenario
(`test/fixtures/fork-commit-teardown.mts`), i.e. exactly the lifecycle race the test exists to
catch (env teardown racing the shared per-database commit thread) — just not the specific
env-cleanup-hook / commit-completion-tsfn race the test's own comment describes, since that
path is guarded by `commitMutex` (see AGENTS.md "Commit execution"). Something else in the
teardown path is locking a mutex whose owning object has already been (or is concurrently
being) destroyed.

## Evidence this is pre-existing and unrelated to any single PR's diff

Same test, same exact crash message, same "iteration=0" first-iteration failure, across three
unrelated PRs in the last 24h — none of which touch commit-teardown/commit-thread code:

| PR / branch | run | `ROCKSDB_JS_COMMIT_THREAD` mode | job |
|---|---|---|---|
| #744 `fix/park-timeout-coordinated-retry` (VT park-timeout only) | [30719858670](https://github.com/HarperFast/rocksdb-js/actions/runs/30719858670/job/91421710769) | `'2'` | Test on Deno (macos-latest) |
| `dependabot/npm_and_yarn/patch-b7a1a6d5c6` (dependency bump) | [30665452962](https://github.com/HarperFast/rocksdb-js/actions/runs/30665452962/job/91271278462) | `'default'` | Test on Deno (macos-latest) |
| `fix/dropped-cf-write-poisons-env` | [30663991359](https://github.com/HarperFast/rocksdb-js/actions/runs/30663991359/job/91266532061) | `'2'` | Test on Deno (macos-latest) |

Hits both commit-thread modes (`default` single-lane and `2` two-lane), so it's not specific to
[...]

PR #780, opened 2026-08-14
fix: close pending transactions with their owning DBHandle; napi-free TransactionHandle::close (#741)

cb1kenobi:
## Root cause: pending transactions leak past their env and get walked at shutdown

A worker env that exits while a transaction is still **pending** leaks its `TransactionHandle`:

- `Transaction::Constructor` → `transactionAdd` stores a **strong** `shared_ptr<TransactionHandle>` in the process-global `DBDescriptor::transactions`.
- The JS wrap finalizer only drops the JS-side ref — it never closes the handle. Only commit/abort call `transactionRemove`.
- So the handle — holding a **live RocksDB transaction + snapshot** and an `env` that dangles once the worker dies — survives in the shared descriptor until the last env's `DBRegistry::Shutdown → finishClose → close()` walks it. That is exactly the #741 production stack (`Shutdown → finishClose → TransactionHandle::close → ~OptimisticTransaction → ~TransactionBaseImpl → _int_free`).

Besides the crash, this is a real resource leak: every worker recycled mid-request leaks an open RocksDB transaction and snapshot until process-wide shutdown.

## The fix

1. **Env-cleanup-hook reap** — `DBRegistry::CloseTransactionsByEnv` → `DBDescriptor::closeTransactionsByEnv`, wired into the module's per-env cleanup hook: every transaction owned by the dying env's handles is closed on that env's own thread while the env is still valid. The invariant (AGENTS.md item 12): an env's pending transactions are reaped by its cleanup hook.

   **Deliberately NOT in `DBHandle::close()`.** A first iteration reaped there and it broke the legitimate close-then-commit overlap: `db.transaction()` awaits its callback before committing, so a commit is routinely one microtask behind `db.close()` and must still run — reaping rejected those commits with "Database not open", and under Deno's scheduling stranded the caller's commit promise (`txn-close-commit-uaf` timed out on all three Deno CI platforms; back to main-parity after the rework). At env teardown no such continuation can exist, so the hook is the only safe reap point. Between a user `db.close()` and env death, an open transaction's handle intentionally lingers (bounded, reaped at env exit).
2. **`TransactionHandle::close()` is now napi-free** — `env`/`jsDatabaseRef`/`envThreadId` are gone. The JS database is passed to `UseLog` per-call by the TS layer (its only consumer), so close is safe from any thread and any teardown phase, removing the cross-thread `napi_delete_reference` hazard class entirely.

## Evidence

Repro: `test/lingering-txn-shutdown.test.ts` + `test/fixtures/fork-lingering-txn-shutdown.mts` — N workers exit with a pending transaction while an anchor pins the descriptor; the last worker's shutdown walks the leaked handles.

Linux/glibc (node:24 arm64, Debian bookworm, native — no sanitizers), 10 leakers per run:

| build | leaker runs | committed-txn control | no-txn control |
|---|---|---|---|
| `main` (3ea9a0fb) | **0/10 pass** — aborts: `corrupted size vs. prev_size`, `corrupted double-linked list`, `free(): invalid pointer`, `malloc_consolidate(): invalid chunk size`, 1× SIGSEGV | 3/3 | 2/2 |
| `main` + `MALLOC_PERTURB_` | **0/4 pass** | — | — |
| **this branch** (owner-reap iteration) | **10/10 pass** (incl. `MALLOC_PERTURB_`) | 3/3 | 2/2 |
| **this branch** (final env-hook reap) | **10/10 pass** | 3/3 | — |

Those abort strings are the #741 family verbatim (incl. the exact `corrupted size vs. prev_size` from the still-skipped `vt-lock-tracker-churn` residual).

### Which change carries the fix?

Isolation arms (same container and method; `main` plus exactly one change; 8 leaker runs each):

| arm | result |
|---|---|
| `releaseByOwner` only | **8/8 pass** |
| napi-free `close()` only | **8/8 pass** |

Each change is independently sufficient on Linux (arms measured on the first iteration; the leaker choreography exercises the same teardown reap as the final env-hook shape), which pins the corrupting write precisely: `Shutdown` walks the leaked handles of dead envs, `close()`'s thread-identity guard passes when the closing thread has recycled a dead worker's pthread (`std::thread::id` collision), and `napi_delete_reference(dead env, dead ref)` corrupts the heap. Removing the napi call breaks the chain at the write; reaping the env's transactions at env death breaks it at the source. Both are kept: the reap also fixes the resource leak (a live RocksDB transaction + snapshot per abandoned worker transaction), and napi-free `close()` removes the hazard class structurally.

The repro test runs on Linux (red on main, green here). On macOS it is `skipIf(darwin)`: a Guard-Malloc-only fault in Node's second-pass napi finalizer drain persists even with the fix — it never reproduces natively or on glibc, and looks like a Node-core teardown artifact worth a separate minimal repro (no rocksdb involved).

Full Vitest suite (final tree, release build): 757 passed / 2 skipped / 0 failures locally. Deno `txn-close-commit-uaf` verified back to main-parity locally (12.5s pass, was a 60s timeout on all three Deno platforms with the first iteration); the new repro test is gated to Linux+Node per the repo's existing teardown-repro convention (`notify-teardown-uaf`).

## Relationship to #745

The dominant #741 crash reproduces on `main` without any of #745's guarded paths engaging (verified: zero `napi_delete_reference` calls, zero `TransactionHandle::close` calls at fault time on the pre-fix repro), and #745's head does not prevent this repro (5/6 crash under the macOS harness). This PR fixes the lifecycle hole those guards work around; #745's `stateMutex` remains relevant for the same-env close-vs-commit-thread TOCTOU (AGENTS item 12a gaps untouched here).

## Review coverage

Cross-model pre-push review (`prepush-review.mjs`, `--author claude`, risk=high/scope=full) at `675bfd5e`:

| leg | model family | result |
|---|---|---|
| graded | **Codex** | ok — 1 blocker, 1 major, 1 nit |
| findings | **Gemini** (`agy`) | ok — 1 major |
| domain adjudication | Claude | **FAILED** (`--permission-mode dontAsk` rejected by the installed CLI) |

Two outside families produced coverage; the Harper domain leg failed, so **the findings below were never adjudicated** (no false-positive filter) — each was verified by hand against the code instead. Receipt published `Human-Review-Need: 4` partly for that degradation. Cursor Grok/Composer were pruned by auto-policy.

What it changed, all in `e9345cce`:

- **Codex corroborated the review blocker independently** and scoped it the same way — narrow timeout hardening here, full admission-and-drain contract in #784. `TransactionHandle::close()` now drains *before* mutating anything and, on drain timeout, returns without touching `state`, log/VT state, or `txn`.
- **Both outside models caught a leak the human review and I both missed**: `close()` can only self-remove from `DBDescriptor::transactions` while it can still reach the descriptor through its `DBHandle`, and a user-closed handle has already reset that pointer — precisely the case this reap exists to catch, so the entry leaked for the process lifetime. `closeTransactionsByEnv` now removes it directly.

Follow-up worth noting for anyone re-running this: the `prepush-review.mjs` domain leg is currently broken against the installed `claude` CLI (`dontAsk` is no longer a valid `--permission-mode`).

Refs #741


src/load-binding.ts
```diff
@@ -378,7 +381,7 @@ export type NativeDatabase = {
 	setIteratorState(buffer: Buffer | Uint8Array): void;
 	tryLock(key: BufferWithDataView, callback?: () => void): boolean;
 	unlock(key: BufferWithDataView): void;
-	useLog(name: string): TransactionLog;
+	useLog(name: string, jsDatabase: NativeDatabase): TransactionLog;
```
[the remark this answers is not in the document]
cb1kenobi: This is an internal private API that is wrapped by `store.useLog()` which does pass in the db.

test/transaction-log.test.ts
```diff
@@ -49,6 +49,18 @@ describe('Transaction Log', () => {
 				expect(db.listLogs()).toEqual(['bar', 'foo']);
 			}));
 
+		it('should throw instead of crash when the transaction is closed', () =>
```
cb1kenobi, line 52: FYI: This causes Node to straight up crash in 2.7.1.

src/binding/database/db_descriptor.cpp
```diff
@@ -1146,6 +1146,32 @@ void DBDescriptor::transactionRemove(std::shared_ptr<TransactionHandle> txnHandl
...
+		for (auto& [id, txnHandle] : this->transactions) {
+			if (txnHandle && txnHandle->dbHandle && txnHandle->dbHandle->env == env) {
+				toClose.push_back(txnHandle);
+			}
+		}
+	}
+
+	for (auto& txnHandle : toClose) {
+		DEBUG_LOG("%p DBDescriptor::closeTransactionsByEnv closing transaction %u (env=%p)\n", this, txnHandle->id, env);
+		txnHandle->close();
```
kriszyp, line 1171: Moving the reap out of user-called `DBHandle::close()` fixes the ordinary microtask overlap, but environment cleanup can still run while native transaction work is active—for example, when a worker is terminated during commit. `TransactionHandle::close()` currently writes `state = Aborted` before waiting, racing the commit worker, then stops waiting after five seconds and deletes `txn` even if RocksDB is still using it. That can continue through freed memory and can mark transaction-log state aborted while the data commit succeeds.

The fix needs the full ordering invariant for this cleanup path: claim close and prevent new work acquisition, wait for every operation that already acquired native state to finish, and only then mutate `state`, release log/VT state, or destroy `txn`. Merely lengthening or removing the timeout does not fix the pre-wait state race or a zero-work-observed → new-work-registers gap. Please add a worker-termination regression with a delay inside native commit execution exceeding five seconds. #784 tracks extending the same contract across every async and synchronous entry point; #780 must make the path it introduces safe before merge.

— KrAIs (Codex)

src/binding/transaction/transaction.cpp
```diff
@@ -80,12 +80,9 @@ napi_value Transaction::Constructor(napi_env env, napi_callback_info info) {
 	bool coordinatedRetry = false;
 	NAPI_STATUS_THROWS(rocksdb_js::getProperty(env, argv[1], "coordinatedRetry", coordinatedRetry));
 
-	napi_ref jsDatabaseRef;
-	NAPI_STATUS_THROWS(::napi_create_reference(env, argv[0], 0, &jsDatabaseRef));
-
 	// create shared_ptr on heap so it persists after function returns
 	std::shared_ptr<TransactionHandle>* txnHandle = new std::shared_ptr<TransactionHandle>(
-		std::make_shared<TransactionHandle>(*dbHandle, env, jsDatabaseRef, disableSnapshot)
+		std::make_shared<TransactionHandle>(*dbHandle, disableSnapshot)
```
kriszyp, line 85: Making `TransactionHandle::close()` N-API-free solves the teardown constraint, but a separate ownership leak remains: the NativeTransaction finalizer only resets the wrapper shared pointer while `DBDescriptor::transactions` retains a strong pointer. In a long-lived environment, an unreachable uncommitted transaction can therefore retain its RocksDB transaction, snapshot, DB handle, VT intents, and transaction-log pending state until database or environment shutdown.

This is now tracked in #785, with #784 as the close-synchronization prerequisite. It does not need to expand #780 if this PR stays narrowly scoped to environment-exit reaping, links #785, and avoids claiming that transaction lifetime generally follows its JS owner.

— KrAIs (Codex)

src/binding/transaction/transaction.cpp
```diff
@@ -1166,8 +1177,13 @@ napi_value Transaction::UseLog(napi_env env, napi_callback_info info) {
 	napi_value transactionLogCtor;
 	NAPI_STATUS_THROWS_ERROR(::napi_get_named_property(env, exports, "TransactionLog", &transactionLogCtor), "Failed to get 'TransactionLog' constructor");
 
-	napi_value jsDatabase;
-	NAPI_STATUS_THROWS_ERROR(::napi_get_reference_value(env, (*txnHandle)->jsDatabaseRef, &jsDatabase), "Failed to get 'jsDatabase' reference");
+	napi_value jsDatabase = argv[1];
```
kriszyp: The new liveness guard fixes the null dereference, but this argument is still checked only as `napi_object`, not as the transaction’s owning database. A public `Store.useLog()` call can supply a transaction context from database B while `this.db` belongs to database A; the returned TransactionLog then resolves the transaction ID in A. Because IDs are descriptor-local and commonly collide, `addEntry()` can silently attach the entry to an unrelated transaction. Please unwrap the NativeDatabase before binding the store and require its DBHandle to equal `txnHandle->dbHandle`; reject mismatches without incrementing the pending log count.

— KrAIs (Codex)

src/load-binding.ts
```diff
@@ -63,7 +63,10 @@ export type NativeTransaction = {
 	putSync(key: Key, value: Buffer | Uint8Array, txnId?: number): void;
 	removeSync(key: Key): void;
 	setTimestamp(timestamp?: number): void;
-	useLog(name: string | number): TransactionLog;
+	// jsDatabase is accepted (and ignored) so Store.useLog can pass it
+	// unconditionally for both StoreContext shapes; only the transaction
+	// variant consumes it.
+	useLog(name: string | number, jsDatabase?: NativeDatabase): TransactionLog;
```
kriszyp: The requiredness is inverted relative to runtime: Transaction `useLog` now requires the database argument, while Database `useLog` ignores the extra argument and still works with only a name. As written, an invalid raw transaction call type-checks while a valid `store.db.useLog(name)` call does not. Please make the transaction parameter required and the database parameter optional, or express the internal two-argument call through a private overload.

— KrAIs (Codex)

test/lingering-txn-shutdown.test.ts
```diff
@@ -0,0 +1,110 @@
...
+		const { code, signal } = await new Promise<{
+			code: number | null;
+			signal: NodeJS.Signals | null;
+		}>((resolve, reject) => {
+			const env: NodeJS.ProcessEnv = { ...process.env };
+			if (process.platform === 'darwin') {
+				env.DYLD_INSERT_LIBRARIES = '/usr/lib/libgmalloc.dylib';
+				env.MallocScribble = '1';
+			}
+			const child = spawn(process.execPath, [fixturePath, generateDBPath(), '0'], { env });
```
kriszyp: Passing a leaker count of zero starts only the anchor and final workers, so this is not actually a transaction-worker-churn control. The fixture already supports a `committer` role; please run the same ten sequential workers with that role so the topology and NativeTransaction lifecycle match while commit removes each transaction from the registry. That would isolate the pending-registry leak as the differentiating condition.

— KrAIs (Codex)

test/workers/vt-lock-tracker-churn-worker.mts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Several of these run concurrently against one DB path with the Verification
+ * Table materialized, all writing a single hot key via coordinatedRetry
+ * transactions so write intents constantly install/join/release on the SAME
+ * LockTracker across threads. A single commit attempt per iteration (no retry
+ * loop) mirrors the proven cross-thread repro shape: retrying in a loop lets
+ * write-intent holders pile up on one key faster than they drain, which
+ * doesn't reproduce the crash any more reliably but does make the test slow.
+ *
+ * Deliberately uses only Commit()/Abort() (not CommitSync/putSync/getSync) --
```
kriszyp: These two new VT churn fixture files are not referenced by any test or script, and this explanation attributes protection to `tryRegisterAsyncWork()` even though that API is absent from this branch. Please either wire the fixture into an explicit regression/manual command and describe the actual env-cleanup synchronization, or remove it so the PR does not imply coverage it lacks.

— KrAIs (Codex)

AGENTS.md
```diff
@@ -372,6 +372,40 @@ sufficient (env teardown does not honor tsfn acquire counts); see
...
 
+12. **An env's pending transactions are reaped by its cleanup hook — never by `DBHandle::close()`**:
+    `transactionAdd` stores a strong `shared_ptr<TransactionHandle>` in the process-global
+    `DBDescriptor`, and only commit/abort call `transactionRemove` (the JS wrap finalizer just
+    drops the JS-side ref). A worker env that exited with a transaction still pending therefore
+    leaked that handle — holding a live RocksDB transaction + snapshot — into the shared
+    descriptor, with its `env` left dangling once the worker died. The last env's
+    `DBRegistry::Shutdown → finishClose → close()` then walked those corpses and corrupted the
+    glibc heap (production signatures: `corrupted size vs. prev_size`, `corrupted double-linked
+list`, `free(): invalid pointer`; HarperFast/rocksdb-js#741 — reproduced 10/10 on Linux/glibc
```
kriszyp: The wrapped `list` line here and `ref)` at line 403 are the only unindented continuations in this invariant. CommonMark can treat them as lazy continuations, but the source is visibly inconsistent and easier to misread. Please restore the surrounding four-space indentation.

— KrAIs (Codex)

kriszyp (commented):
## Review outcome: changes needed before merge

#780 is the right narrow replacement for the dominant reproduced #741 failure: it reaps pending transactions at environment exit and removes the transaction-owned N-API reference. At head `630a3efb`, two correctness problems still block that fix:

1. **Environment cleanup can destroy a transaction while native work is active.** The cleanup path calls `close()`, which mutates `state` before waiting and, after a five-second timeout, releases log/VT state and deletes `txn` even if RocksDB still uses it. The inline comment now states the required ordering: prevent new acquisition, drain already-acquired native work definitively, then mutate or destroy state. #784 tracks extending that invariant across every entry point.
2. **`useLog()` can bind through a foreign database object.** Because the new argument is checked only as an object, a transaction from database B can be paired with database A; descriptor-local transaction IDs commonly collide, so a log entry can silently attach to an unrelated transaction. The binding must validate exact DBHandle ownership before changing log state.

The native type declarations, committed-worker control, and unreferenced VT churn fixtures also need the inline corrections so the API contract and regression evidence match the implementation.

The GC-retained pending-transaction problem is real but can remain outside this PR now that it is tracked in #785. The omitted #745 lifecycle scope is durable under #741:

- #783 — parked coordinated-retry TSFNs outliving their environment
- #784 — complete `TransactionHandle` close/work synchronization
- #785 — GC-abandoned pending transactions
- #786 — remaining `DBHandle` N-API references reachable from shutdown
- related: #746 — Deno/macOS commit-teardown mutex crash

Accordingly, #780 should continue to use `Refs #741`: merging it must not close #741 or be described as fully superseding #745. #745 can stay closed as the obsolete implementation vehicle.

Reviewed head: `630a3efbc67d7705638204dfdafffc02cfca9d78`. `git diff --check` passes.

— KrAIs (Codex)

kriszyp:
Another thing I will note: I don't see any info in the PR description about which models were used to review this before pushing (should be at least two outside models). Just a note for future PRs.

cb1kenobi:
Good call on the model coverage — I'd skipped it, and running it was worth the detour. Ran `prepush-review.mjs` (`--author claude`, risk=high/scope=full) at `675bfd5e`: **Codex** (graded) + **Gemini** (`agy`) both produced coverage; the Harper **domain adjudication leg failed** (`--permission-mode dontAsk` is no longer valid for the installed `claude` CLI), so those findings are unadjudicated and I verified each by hand. Details now in the PR description.

Two outcomes, both in `e9345cce`:

**1. Your close-during-active-work blocker — fixed, and Codex reached it independently.** It also scoped it exactly as you did: narrow timeout hardening here, the full admission-and-drain contract in #784. `waitForAsyncWorkCompletion()` now reports whether it actually drained; `close()` drains *before* touching anything and, on timeout, returns without mutating `state`, releasing log/VT state, or deleting `txn` — the in-flight commit owns its own cleanup. A leaked transaction is recoverable; a use-after-free and a log/data disagreement are not.

The regression test took two attempts and the first one was worthless — worth recording. `ROCKSDB_JS_COMMIT_DELAY_MS` fires *after* execute completes, so it can't reach the drain at all. My first new seam sat at the end of `executeCommitWork`, which held the work registered but left `txn` unused during the stall: it passed identically with and without the fix (5/5 both arms on glibc). Moving it to immediately before `rocksdb::Transaction::Commit()` gives a real red/green — **3/3 SIGSEGV before the hardening**, clean after, and since it's a plain dereference rather than silent heap corruption it reproduces natively on macOS with no allocator instrumentation, so that test isn't platform-gated.

**2. A leak both outside models flagged that neither of us caught.** `close()` can only self-remove from `DBDescriptor::transactions` while it can still reach the descriptor through its `DBHandle` — and a handle the user closed earlier has already reset that pointer. That's exactly the case this reap exists to catch, so the entry (a strong ref to the handle, and through it the `DBHandle`) leaked for the life of the process. `closeTransactionsByEnv` now removes it directly.

The other six inline comments are addressed in `675bfd5e`, with one deliberate deviation: on the AGENTS.md indentation, adding the four spaces puts them *inside* an inline code span that straddles the line break (```corrupted double-linked` / `list```), changing what renders — and `oxfmt` strips it right back out, which is how it got that way. I rewrapped so the spans sit on one line instead: uniform indentation, correct rendering, formatter-stable.

Also filed separately: `prepush-review.mjs`'s domain leg is currently broken against the installed `claude` CLI.

—
_Generated by Barber AI_

src/binding/transaction/transaction.cpp
kriszyp: Making `TransactionHandle::close()` N-API-free solves the teardown constraint, but the unreachable-transaction root remains: the NativeTransaction finalizer only resets the wrapper’s shared pointer, while `DBDescriptor::transactions` retains a strong pointer. If an uncommitted `Transaction` is garbage-collected while its database remains open, its snapshot, locks/VT intents, log pending count, and DB handle remain alive until database close. After resolving the in-flight-close synchronization above, please close/abort from the finalizer or change the registry ownership model, and add a forced-GC regression against a long-lived database.

— KrAIs (Codex)

kriszyp, line 92: Making `TransactionHandle::close()` N-API-free solves the teardown constraint, but the unreachable-transaction root remains: the NativeTransaction finalizer only resets the wrapper’s shared pointer, while `DBDescriptor::transactions` retains a strong pointer. If an uncommitted `Transaction` is garbage-collected while its database remains open, its snapshot, locks/VT intents, log pending count, and DB handle remain alive until database close. After resolving the in-flight-close synchronization above, please close/abort from the finalizer or change the registry ownership model, and add a forced-GC regression against a long-lived database.

— KrAIs (Codex)

kriszyp (commented):
> Also filed separately: prepush-review.mjs's domain leg is currently broken against the installed claude CLI. (--permission-mode dontAsk is no longer valid for the installed claude CLI), 

What version of the CLI are you using? I think the latest should support dontAsk. But I posted https://github.com/HarperFast/skills-internal/pull/126 to tolerate this. And did you say you filed an issue (that I should link)?

Anyway, codex thinks there is still a pointer to fix.

src/binding/transaction/transaction.cpp
cb1kenobi: Good catch, and this is now resolved by merging `main` into the branch — specifically [#768 "fix(transaction): weak-ref the transaction registry so a dropped handle releases itself"](https://github.com/HarperFast/rocksdb-js/pull/768), which landed the finalizer-close path this comment asks for:

- **Close/abort from the finalizer.** The `NativeTransaction` finalizer now calls `onWrapperCollected()` *before* dropping its `shared_ptr` (the registry's own strong ref means the reset alone can't destroy the handle). That routes through `closeOrphanIfUnused()` → `close()`, which releases the snapshot, VT intents/locks, log pending count, and the RocksDB transaction. The one deferral is `state == Committing` (a commit still owns the handle); the commit-completion paths close it instead once it settles, and live dependents (a cold-cache async get holding a `shared_ptr`, or a transaction-backed iterator keeping `activeIteratorCount` nonzero) retry the close when they release.

- **Forced-GC regression against a long-lived database.** `test/transaction-orphan-gc.test.ts` covers exactly this — e.g. *"should release a transaction dropped without commit or abort"* and *"should release a transaction dropped after a failed commit"*, both asserting release while the database stays open, plus the iterator/async-work deferral cases. Green locally (full suite: 818 passed / 2 skipped / 0 failures).

Two notes on how this composes with what #780 introduced:

1. `TransactionHandle::close()` is now **napi-free**, so the finalizer-driven close from #768 is safe from any thread and any teardown phase — the `napi_delete_reference(dead env)` hazard that motivated #741 is gone structurally.
2. #780's env-cleanup-hook reap (`CloseTransactionsByEnv`) and #768's finalizer reap are complementary and idempotent: the hook handles a worker **env exiting with a transaction still pending** (wrapper never GC'd), the finalizer handles a **wrapper GC'd while its env is alive**. Both funnel through the same `closed`-gated `close()` + idempotent `transactionRemove`, so a handle reachable by both is closed exactly once. This is written up as AGENTS.md invariant 15, cross-referencing #768's invariant 13.

Thanks for keeping the ownership model honest here.

— Generated by Claude (Opus 4.8)

cb1kenobi:
## Cross-model pre-push review — `5e8ad5b0` (delta since `675bfd5`)

Ran the standard `prepush-review.mjs` route (`--author claude`) over the 123-commit main-merge delta. **Codex** (graded) + **Gemini** contributed; the **Harper domain adjudication leg failed to launch** (`claude` not on `PATH` in this environment) and **cursor-grok** refused the diff (it declines diffs that modify `AGENTS.md`). Because the adjudication leg didn't run, I adjudicated the outside findings by hand against the code. `Human-Review-Need: 4` is a deterministic-policy floor from the high-risk native surface **plus** the degraded coverage — not an unresolved real finding.

### Codex (graded)
Verdict **COMMENT(S)**. Confirmed both prior-round findings resolved, no new correctness regression in the merged orphan-cleanup interaction:
- Teardown blocker resolved — `close()` checks drain status and preserves `txn` on timeout.
- Registry-leak major resolved — env cleanup removes the handle directly even when `db.close()` already reset the descriptor pointer.

### Gemini findings — adjudication

| # | Sev | Verdict | Reasoning |
|---|---|---|---|
| 5 | blocker | **false positive** | `lingering-txn-worker.mts` runs as a `worker_threads` Worker — `process.exit(0)` there stops only the worker; the parent resolves `worker.once('exit')` (the choreography's core). The test asserts `signal === null`, so a teardown crash fails it. The differential evidence (main SIGABRTs 0/10, branch 10/10) requires individual worker exits. |
| 2 | major | **false positive / by-design** | `close()` sets `closed=true` at entry, so `isClosed()` is `true` on the timeout path and `~TransactionHandle()`→`close()` no-ops — no inconsistent-state re-entry, no UAF. This is the intended, bounded leak-on-timeout (full admission-and-drain deferred to #784). Codex corroborated no UAF. |
| 1 | minor | not actionable | `UseLog` is called once per log bind, not a hot path; the `instanceof` guard is required to make `napi_unwrap` safe before the pointer-identity check. |
| 3 | minor | incorrect | `napi_instanceof` does not error on a `null`/primitive left operand (it returns `false`), so the explicit "Invalid argument, expected Database instance" branch is reached — no raw-status throw. |
| 4 | minor | endorsement | Confirms the cross-database ownership check is correct (not a defect). |
| 5b | minor | by-design | The macOS/Deno/Bun skip is deliberate and documented — #741 is a Linux/glibc production repro; macOS carries only a gmalloc-only Node teardown artifact. |
| 6 | nit | **corroborated (Codex + Gemini)** | Comment density in the teardown paths. Judgment: kept — these are load-bearing *why* comments (deliberate leak, napi-free rationale, foreign-db check) that AGENTS.md invariant 15 explicitly relies on ("do not simplify this into `DBHandle::close()`"). This subsystem documents teardown hazards by convention. |

**Net: no real blockers or majors.** One corroborated nit (comment density), kept by design.

Environment note for re-runs: the domain adjudication leg needs `claude` on `PATH`; it was absent here, so the run was Codex + Gemini only.

— Generated by Claude (Opus 4.8)

kriszyp (approved):
Great!

PR #789, opened 2026-08-20
Reject in-flight async gets when an orphaned transaction is GC'd

cb1kenobi:
Pins the `NativeTransaction` wrapper with a `napi_ref` for the life of a cache-miss async get, so dropping the JS `Transaction` cannot finalize (and `close()`) the handle under a still-running worker. The read completes; the snapshot is released once the last ref drops and the wrapper is collected. In-flight commit is unchanged.

This is Kris's requested approach: use N-API's own strong reference instead of cancelling the get and closing from the last complete callback. `Transaction::Get` refs `jsThis`; `db.getBinary({ transaction })` now passes the native wrapper into `Database::Get` (numeric `txnId` remains for other callers and still has the complete-callback close backstop).

## For the human reviewer

No open judgment calls — Kris chose `napi_ref` over reject-on-GC. Residual: `waitForAsyncWorkCompletion`'s 5s timeout vs a still-running `Get` is a pre-existing close/abort hazard, not introduced here. The new tests force GC after drop and before await, but they do not hold the worker on a barrier, so they do not fail a missing pin as hard as a stalled-sibling test would.

## Verification

`node --expose-gc ./node_modules/vitest/vitest.mjs test/transaction-orphan-gc.test.ts` — 8 passed. Also `test/transactions.test.ts` + `test/transaction-cross-column-family.test.ts` — 180 passed with the pin commit. `pnpm type-check` clean via pre-commit.

## Review coverage

Authored by Cursor Grok 4.6. Cross-model review of `3f2836c8` (`--full`) then delta of `b4516e75` via `prepush-review.mjs` vs `origin/kris/txn-registry-weakptr`: gemini via agy (default model) ✓, gpt-5.6-sol (codex graded) ✓, Harper domain ✗ (claude not on PATH). cursor-grok disabled as authoring family. `11100850` is a follow-up for the foreign-database check and was not re-reviewed.

Complexity: complicated

<sub>Human-Review-Need: 4 @ 11100850237d</sub>


src/binding/transaction/transaction_handle.cpp
```diff
@@ -520,17 +537,20 @@ napi_value TransactionHandle::get(
...
 		[](napi_env env, napi_status status, void* data) { // complete
-			auto state = reinterpret_cast<AsyncGetState<TransactionHandle*>*>(data);
+			auto state = reinterpret_cast<AsyncGetState<std::shared_ptr<TransactionHandle>>*>(data);
 			state->deleteAsyncWork();
 
 			if (status != napi_cancelled) {
 				resolveGetResult(env, "Transaction get failed", state);
 			}
 
+			if (state->handle && state->handle->wrapperCollected.load()) {
```
kriszyp: Please close only when this completion leaves no other transaction work executing. With two cache-miss gets outstanding, the first completion can reach here while `activeAsyncWorkCount == 1`; `close()` then waits at most five seconds (`src/binding/napi/async.h:166-180`) before deleting `txn`. If the second RocksDB `Get` is still stalled after that timeout, it resumes against the freed transaction—the shared pointer protects `TransactionHandle`, not its raw `txn`. A centralized “close if wrapper collected and active count is zero” transition, called after each unregister, would let the last execution close safely. Please cover multiple pending gets as well as the existing single-get case.

— KrAIs (Codex)

src/binding/transaction/transaction_handle.cpp
```diff
@@ -466,29 +495,42 @@ napi_value TransactionHandle::get(
...
+				state->handle->close();
+			}
+
 			delete state;
```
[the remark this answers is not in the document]
cb1kenobi: Agreed on last-get close. Landed in 2ee75bdd by re-entering `onWrapperCollected()` from the get complete callback — it already returns while `activeAsyncWorkCount > 0`, then `close()`s. Did not take the `delete state` then count-0 snippet: `unregisterAsyncWork` already runs in execute via `signalExecuteCompleted()`, so `delete state` does not decrement the count and gating after delete would still `close()` on the first complete. Added a two-get rejection test.

test/transaction-orphan-gc.test.ts
```diff
@@ -148,6 +149,85 @@ describe('orphaned transactions', () => {
...
+		})
+	);
+
+	itWithGC('should reject two in-flight async gets when the transaction is dropped', () =>
+		dbRunner(async ({ db, dbPath }) => {
+			await db.put('foo', 'bar');
+			await db.put('baz', 'qux');
+			await db.flush();
+
+			setTxnGetExecuteDelayMsForTesting(250);
```
kriszyp: This case does not appear to fail on the pre-fix implementation. Both gets receive the same 250 ms delay, while `close()` waits up to five seconds: either both workers finish before the first completion runs, or the second finishes safely while the first completion waits. In both cases the old code still rejects both promises and removes the registry entry. Could we use a deterministic stagger or barrier so the first completion runs while the second worker remains blocked, assert that the transaction stays registered without blocking the event loop, then release the second worker and assert cleanup? That would specifically guard the new last-active-get invariant.

— KrAIs (Codex)

kriszyp (commented):
Could we strong reference (`napi_ref`) `this` (`jsThis`), which should entrain the transaction, to pin the transaction in memory, using the natural/native mechanism, rather than creating a separate mechanism for tracking?

test/transaction-orphan-gc.test.ts
cb1kenobi: Took the `napi_ref` approach from the review instead, so this last-get-close race is no longer the get-path invariant: the wrapper cannot be finalized while the get holds a strong ref, and the read completes rather than rejecting. The two-get test now asserts both reads resolve and the snapshot is released after GC. A barrier that holds one worker past `close()`'s 5s wait would still be the way to lock the *old* complete-callback close, but that path is gone for pinned gets.

`11100850`

cb1kenobi:
@kriszyp Done — async get now holds a strong `napi_ref` on `NativeTransaction` (`jsThis` from `Transaction::Get`, and the wrapper passed through `db.getBinary({ transaction })`). That keeps the finalizer from running until the worker finishes, so we dropped reject-on-GC and the execute-delay seam. Numeric `txnId` gets still cannot pin (no wrapper in that call); complete still re-enters `onWrapperCollected` for that backstop.

`11100850`

src/binding/transaction/transaction.cpp
```diff
@@ -870,7 +870,7 @@ napi_value Transaction::Get(napi_env env, napi_callback_info info) {
 		if (vtSlot) vtObserved = vtSlot->load(std::memory_order_acquire);
```
kriszyp, line 870: Could we defer orphan closure while a transaction-backed iterator is still alive? `getRange()` returns an iterator that does not retain the JS `NativeTransaction`; native construction unwraps a raw transaction handle at `src/binding/iterator/db_iterator.cpp:123-167`, and `DBIteratorHandle` retains only the database handle while creating its RocksDB iterator from `txnHandle->txn` at `src/binding/iterator/db_iterator_handle.cpp:31-51`. A caller can therefore retain the iterator, drop the transaction, and trigger this finalizer, which deletes the RocksDB transaction underneath the still-usable iterator. This conflicts with the explicitly supported outstanding-read-iterator lifetime described at `src/transaction.ts:97-103` and risks invalid reads or native memory misuse. Please make transaction iterators participate in transaction lifetime tracking (or retain the wrapper) and add a GC regression that drops the transaction but continues the iterator before allowing orphan cleanup.

— KrAIs (Codex)

<sub>(`src/binding/transaction/transaction.cpp:116` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

src/store.ts
```diff
@@ -837,9 +837,16 @@ export class Store {
 			// return a number indicates it is using the default buffer
 			if (result === NOT_IN_MEMORY_CACHE_FLAG) {
 				// is not in memory cache, use async get since this will involve disk access
+				const txn = (options as DBITransactional | undefined)?.transaction;
```
kriszyp, line 840: Could we validate the transaction wrapper before the synchronous cache probe as well? `context.getSync()` receives only the descriptor-local numeric ID at `src/store.ts:830-835`; the new ownership check at `src/binding/database/database.cpp:752-754` runs only after that probe returns `NOT_IN_MEMORY_CACHE_FLAG`. If databases A and B both have transaction ID 1, `dbA.get(key, { transaction: txnB })` can therefore read through A's transaction on a cache hit, while the identical cold-cache call reaches the new check and throws. That makes correctness depend on cache residency and can read under the wrong snapshot. Please pass/unwrap the wrapper on the fast path or otherwise validate ownership before `getSync()`, with colliding-ID warm- and cold-cache regressions.

— KrAIs (Codex)

kriszyp (commented):
Looks much better! A couple of things to consider for iterators.

cb1kenobi:
Closing this as **superseded by the rebased base** (`kris/txn-registry-weakptr`, [#768 — Release a transaction dropped without commit or abort, instead of leaking its read snapshot for the life of the process](https://github.com/HarperFast/rocksdb-js/pull/768)).

When #768 was rebased onto `main`, its branch also grew a full orphan-lifetime design that covers everything this PR set out to do — just with a different mechanism:

- **Orphan close is reference-counted on the base.** `onWrapperCollected()` + `closeOrphanIfUnused()` defer the close until the last dependent releases, tracking async work (`activeAsyncWorkCount`) and transaction-backed iterators (`activeIteratorCount` via `registerIterator`/`unregisterIterator`). This PR's approach — pinning the JS wrapper with a `napi_ref` during async gets — is an alternative to the same end, so it's fully superseded rather than additive.
- **Test coverage already exists on the base.** In-flight async-get survival, iterator survival, dropped-after-rejected-commit, and the never-read orphan are all covered there (`expectDependentFixtureSurvives('async-get' | 'iterator')` and friends). This PR's versions are the same scenarios rewritten against the `napi_ref` design.
- **The foreign-database guard is largely redundant.** The base routes transactional gets through `descriptor->transactionGet(txnId)`, scoped to that database's own registry, so a foreign transaction is already rejected structurally.

**One finding worth keeping on the radar:** transaction ids are per-descriptor (`++nextTransactionId`), so a cross-descriptor id *collision* could let `transactionGet(txnId)` silently return the wrong database's transaction instead of erroring. This PR's descriptor-equality check closed that, but it's a narrow edge, independent of the orphan-lifetime work, and better handled as its own small correctness change than by carrying a near-empty PR. Happy to file it separately if we want it tracked.

Nice work here — the design ended up converging on the same guarantees, and the base is the cleaner home for it.

<sub>— Generated by Claude (Opus 4.8)</sub>

PR #791, opened 2026-08-20
fix(usb): stop finalizer from dereferencing tsfn-owned callbackRef during shutdown (#790)

### LINKED ISSUE #790: User shared buffer finalizer dereferences a freed callbackRef during shutdown (SIGSEGV)

cb1kenobi:
Fixes #790.

## Problem

Sending `ctrl-c` to quit Harper crashed during shutdown (SIGSEGV) on macOS. The faulting frame:

```
napi_get_reference_value                       <- EXC_BAD_ACCESS
rocksdb_js::userSharedBufferFinalize
... BufferFinalizer::FinalizeBufferCallback
... Environment::RunCleanup / FreeEnvironment  <- worker/process teardown
```

`getUserSharedBuffer(key, buf, { callback })` registers a listener whose `napi_ref` ownership is **transferred to the listener's threadsafe function** in `addListener` (the tsfn finalizer `napi_delete_reference`s it). The buffer's `UserSharedBufferFinalizeData` also stashed that same raw ref as a borrowed pointer. At shutdown the listener is torn down (`removeListenersByOwner` on `db.close()`, or `removeListenersByEnv` on env teardown) → the tsfn finalizer deletes the ref → then `userSharedBufferFinalize` calls `napi_get_reference_value` on the **freed ref**: a use-after-free.

On a plain build the freed slot usually reads back benign, so it only bit intermittently in production. It needs no worker or GC — a plain main-thread `db.close()` + natural process exit reproduces it, because the ref lifetime belongs to the tsfn, not the finalizer.

## Fix

Remove the listener **by identity** instead of re-resolving a ref the finalizer doesn't own:

- `EventEmitter::addListener` now returns the `std::shared_ptr<ListenerCallback>` it already allocates (was `napi_ref`).
- `UserSharedBufferFinalizeData` holds a `std::weak_ptr<ListenerCallback>` instead of the raw `napi_ref`.
- New env-free overload `EventEmitter::removeListener(key, target)` releases the tsfn (which queues the ref deletion on the JS thread, as before) and erases the map entry — **without touching the napi_ref**.

The finalizer locks the weak_ptr and, only if the listener is still registered, removes it by identity. If it was already torn down (`db.close()` runs `removeListenersByOwner` before `descriptor.reset()`), the lock fails and there's nothing to do. The map holds the sole strong ref to the `ListenerCallback`, and close/env-teardown/finalizer all run on the same env thread, so there's no cross-thread lifetime race.

The `buffer.cancel()` / `db.removeListener(key, fn)` path (removal by function value) and the "cleanup callbacks on GC" behavior are unchanged.

## Testing

- New macOS-only, Guard-Malloc-gated child-process regression test (`test/user-shared-buffer-shutdown.test.ts` + fixture): opens a DB, gets user-shared buffers **with callbacks**, `db.close()`, keeps them referenced, and exits naturally so `RunCleanup` finalizes them. **SIGSEGVs before this change, passes after** (verified 20+ runs under Guard Malloc across three repro variants). Gated to macOS because a plain build reads benign freed memory and can't fault deterministically.
- All existing user-shared-buffer, notify, events, global-events, notify-teardown, and shutdown suites pass. `pnpm check` (type-check, lint, format) is clean.

## Reviews

Deep review (two independent passes) and cross-model review (Codex + Gemini): **no crash-safety, concurrency, lifetime, or security blockers.** Follow-ups addressed here: repro DB dirs are now cleaned up (honoring `KEEP_FILES`), and an overbroad finalizer comment was tightened. Two Gemini "majors" were false positives (the `store.ts` layer unwraps `options.callback` before native; `.mts` type-stripping is unflagged at the repo's engines floor and existing fixtures spawn identically — the test runs green).

## Pre-existing follow-ups (out of scope, noted for tracking)

- `getUserSharedBuffer` allocates `finalizeData` before `napi_create_external_arraybuffer`; on that (rare) failure it leaks. RAII cleanup would tidy it.
- If `defaultBuffer` isn't an ArrayBuffer, the listener is registered before the validation throw, so it lingers until `close()`.

### DIFF
--- a/src/binding/database/database.cpp
+++ b/src/binding/database/database.cpp
@@ -1415,21 +1415,21 @@ napi_value Database::GetUserSharedBuffer(napi_env env, napi_callback_info info) 
 	std::string keyStr(key + keyStart, keyEnd - keyStart);
 
 	// if we have a callback, add it as a listener
-	napi_ref callbackRef = nullptr;
+	std::shared_ptr<ListenerCallback> listener;
 	napi_valuetype type;
 	NAPI_STATUS_THROWS(::napi_typeof(env, argv[2], &type));
 	if (type != napi_undefined) {
 		if (type == napi_function) {
 			DEBUG_LOG("Database::GetUserSharedBuffer key start=%u end=%u:", keyStart, keyEnd);
 			DEBUG_LOG_KEY_LN(keyStr);
-			callbackRef = (*dbHandle)->descriptor->addListener(env, keyStr, argv[2], *dbHandle);
+			listener = (*dbHandle)->descriptor->addListener(env, keyStr, argv[2], *dbHandle);
 		} else {
 			::napi_throw_error(env, nullptr, "Callback must be a function");
 			return nullptr;
 		}
 	}
 
-	return (*dbHandle)->descriptor->getUserSharedBuffer(env, keyStr, *dbHandle, argv[1], callbackRef);
+	return (*dbHandle)->descriptor->getUserSharedBuffer(env, keyStr, *dbHandle, argv[1], std::move(listener));
 }
 
 /**
--- a/src/binding/database/db_descriptor.cpp
+++ b/src/binding/database/db_descriptor.cpp
@@ -1569,15 +1569,10 @@ static void userSharedBufferFinalize(napi_env env, void* unusedData, void* hint)
 
 	if (auto dbHandle = finalizeData->dbHandle.lock()) {
 		DEBUG_LOG("userSharedBufferFinalize GC'd dbHandle=%p\n", dbHandle.get());
-		// Remove the listener registered for this buffer's key by identity, but
-		// only while it is still live. The listener's napi_ref is owned by its
-		// threadsafe function, which deletes it once the listener is torn down
-		// (on close / env teardown), so re-resolving the ref here would be a
-		// use-after-free during shutdown (HarperFast/rocksdb-js#790). A live
-		// weak_ptr means the listener is still in the emitter's map; if it has
-		// already been removed (e.g. by close's removeListenersByOwner, which
-		// runs before descriptor.reset()), the lock fails and there is nothing
-		// to do.
+		// Remove this buffer's listener by identity, never by its napi_ref: the
+		// ref is owned by the listener's tsfn (see UserSharedBufferFinalizeData),
+		// so re-resolving it here was the shutdown UAF in #790. removeListener is
+		// a no-op if the listener is already gone (close / env teardown).
 		if (auto listener = finalizeData->listener.lock()) {
 			if (dbHandle->descriptor) {
 				DEBUG_LOG("%p userSharedBufferFinalize removing listener for key:", dbHandle.get());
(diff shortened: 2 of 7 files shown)


kriszyp (approved):
Great!

PR #792, opened 2026-08-20
fix(txnlog): scan recovery framing from the file, not a whole-image buffer

cb1kenobi:
Stacked on #723. Open-time recovery now walks v1 framing with positional header reads instead of allocating a whole-file buffer, and `findLastCompleteTransactionEnd()` is gone. An I/O failure throws `DBException` rather than classifying as a torn tail.

## For the human reviewer

1. **A read error during active-file recovery fails database open** rather than logging and skipping recovery as before. Alternative: catch at `TransactionLogStore::load()` and degrade. Reversible with one catch. "No" means we can append past a torn tail we never classified.

2. **Positional header walk with a 64 KiB nearby window** rather than mapping the file, reading fixed chunks, or reusing `MemoryMap`. Peak memory is O(1); a change of mind is contained to `ScanReader`. Mapping was rejected because recovery must run before mappings are handed to readers (Windows truncate).

3. **Validation still slurps a whole image** (`validateTransactionLogImage`) while recovery streams. Alternative: convert validation too. `backups.verify()` keeps the O(file) profile; reversible now that the callback exists.

4. **`scanTransactionLogForRecovery(TransactionLogFile&)` is public** and takes `fileMutex`; in-lock callers use `scanRecoveryLocked()`. Alternative: tests lock and call `scanRecoveryLocked()` themselves. The mutex is not recursive.

5. **`readFullyFromFile` retries `EINTR` unboundedly.** Alternative: a bound. A signal storm could spin under `fileMutex`.

Unresolved leftovers that do not need a ruling: corruption-resync still does a positional read per plausible chain hop (cold path only); the Windows `SetFilePointerEx` change has no >2GiB test on Windows (`open()` would map 2GiB); there is no Vitest that an unreadable active log rejects `open()`.

## Verification

Native GoogleTest: 125 tests, 122 passed (3 MADV_COLD skipped on macOS), including `TypicalPayloadsAmortizesReads`, header-only large-payload walks, throw-not-truncate on I/O failure, file/buffer parity, and a POSIX sparse header at `0x80000000`. Vitest `test/transaction-log.test.ts` + `test/transaction-log-crash-recovery.test.ts`: 82 passed, including SIGKILL crash recovery. Deep review of the recovery-scan diff at `1ada20b2`: 0 findings.

## Review coverage

Authored by Cursor Grok 4.6. Cross-model review: gpt-5.6-sol (codex) ✓, gemini via agy (default model) ✓, Harper domain adjudication (claude-opus-5) ✓ on `1ada20b2` (the delta at `797e416f` pruned domain). cursor-grok disabled (authoring family). cursor-composer disabled (an earlier round refused a diff that changes `AGENTS.md`). Receipt @ 797e416f.

Complexity: complicated

<sub>Human-Review-Need: 3 @ 797e416f2d3a</sub>

### DIFF
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -1,7 +1,9 @@
+#include <cerrno>
 #include <cmath>
 #include <fstream>
 #include <limits>
 #include <sstream>
+#include <string>
 #include <system_error>
 #include <vector>
 #include "napi/global_events.h"
@@ -151,6 +153,41 @@ void TransactionLogFile::open(const double latestTimestamp) {
 	}
 }
 
+bool TransactionLogFile::readFullyFromFile(uint32_t offset, void* dest, uint32_t n) {
+	auto* out = static_cast<char*>(dest);
+	uint32_t remaining = n;
+	uint32_t at = offset;
+	while (remaining > 0) {
+		int64_t got = this->readFromFile(out, remaining, static_cast<int64_t>(at));
+		if (got < 0) {
+#ifdef PLATFORM_POSIX
+			if (errno == EINTR) {
+				continue;
+			}
+#endif
+			return false;
+		}
+		if (got == 0) {
+			return false;
+		}
+		out += got;
+		at += static_cast<uint32_t>(got);
+		remaining -= static_cast<uint32_t>(got);
+	}
+	return true;
+}
+
+RecoveryScan TransactionLogFile::scanRecoveryLocked() {
+	uint32_t fileSize = this->size.load(std::memory_order_relaxed);
+	return scanTransactionLogForRecovery(
+		fileSize,
+		[](void* context, uint32_t offset, void* dest, uint32_t n) {
+			return static_cast<TransactionLogFile*>(context)->readFullyFromFile(offset, dest, n);
+		},
+		this
+	);
+}
+
 uint32_t TransactionLogFile::scanForLastCompleteTransactionEnd() {
 	std::lock_guard<std::mutex> fileLock(this->fileMutex);
 
@@ -163,17 +200,14 @@ uint32_t TransactionLogFile::scanForLastCompleteTransactionEnd() {
 		return 0; // header-only or empty: no transactions at all
 	}
 
-	std::vector<char> buffer(fileSize);
-	int64_t bytesRead = this->readFromFile(buffer.data(), fileSize, 0);
-	if (bytesRead < 0 || static_cast<uint32_t>(bytesRead) != fileSize) {
-		DEBUG_LOG("%p TransactionLogFile::scanForLastCompleteTransactionEnd Failed to read file: %s (read=%lld, size=%u)\n",
-			this, this->path.string().c_str(), static_cast<long long>(bytesRead), fileSize);
-		return 0;
+	RecoveryScan scan;
+	try {
+		scan = this->scanRecoveryLocked();
+	} catch (const DBException& error) {
+		throw DBException(std::string(error.what()) + ": " + this->path.string());
 	}
-
-	uint32_t end = findLastCompleteTransactionEnd(buffer.data(), fileSize);
-	this->lastCompleteTransactionEnd.store(end, std::memory_order_relaxed);
-	return end;
+	this->lastCompleteTransactionEnd.store(scan.lastCompleteTransactionEnd, std::memory_order_relaxed);
+	return scan.lastCompleteTransactionEnd;
 }
 
 void TransactionLogFile::recoverTail(uint32_t protectedPosition) {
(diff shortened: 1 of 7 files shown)


src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -188,17 +225,12 @@ void TransactionLogFile::recoverTail(uint32_t protectedPosition) {
...
+	try {
+		scan = this->scanRecoveryLocked();
+	} catch (const DBException& error) {
+		throw DBException(std::string(error.what()) + ": " + this->path.string());
 	}
-
-	RecoveryScan scan = scanTransactionLogForRecovery(buffer.data(), fileSize);
 	// Publish the complete-transaction boundary from this same scan so the store can
 	// seed its committed watermark without re-reading the file.
 	this->lastCompleteTransactionEnd.store(scan.lastCompleteTransactionEnd, std::memory_order_relaxed);
```
kriszyp, line 236: When an interrupted transaction begins in a newly rotated file, this file has no local flagged boundary, so the whole uncommitted prefix is retained. `load()` seeds the watermark from an older file, but the next successful commit advances it to `nextLogPosition`; committed queries then traverse and expose the retained aborted entries before the new transaction. Please repair the complete rotation-spanning tail, or fail/quarantine the store when it cannot be repaired safely. A regression should reopen this scenario, commit one new transaction, and verify that only the previously completed transaction plus the new one are visible.

<sub>(`src/binding/transaction_log/transaction_log_file.cpp:324` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

kriszyp, line 236: At this point recovery has identified a trailing batch as uncommitted, but an `eraseTail()` failure only logs and allows the store to open. A later commit advances the monotonic watermark past these known-aborted entries and exposes them as committed. Please propagate repair failure (including truncation durability failure) so the store cannot accept writes until the tail is safely removed, and add an injected-failure test.

<sub>(`src/binding/transaction_log/transaction_log_file.cpp:349` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

src/binding/transaction_log/transaction_log_recovery.cpp
```diff
@@ -72,37 +150,33 @@ RecoveryScan scanTransactionLogForRecovery(const char* data, uint32_t fileSize)
...
 			static_cast<uint64_t>(pos) + TRANSACTION_LOG_ENTRY_HEADER_SIZE + length > fileSize) {
-			// This frame is broken. If valid framing resumes after it, this is
-			// mid-file corruption: truncating would drop committed entries that
-			// are still framed, so leave it for the caller to surface. Otherwise
-			// it is a torn tail we can safely drop back to `pos`.
-			if (validFramingResumes(data, pos + 1, fileSize)) {
+			// Intact frames after the break are mid-file corruption; truncating would
+			// discard them. A torn tail has nothing valid behind it.
+			if (validFramingResumes(source, pos + 1)) {
 				return scan(RecoveryScan::Kind::MidFileCorruption, pos);
```
kriszyp, line 175: `validFramingResumes()` proves that intact frames exist after this break, but returning here leaves `lastCompleteTransactionEnd` at the last boundary before the corruption. If `txn.state` is behind, committed transactions after the break remain outside the recovered watermark, so the reader never reaches the `CorruptFrameError`/resync path intended to preserve them. Please return the resync offset and continue scanning transaction flags after it while retaining the first corruption location; cover a break followed by flagged commits with an older `txn.state`.

kriszyp (commented):
Cool, I think I had looked at doing this, but glad you are actually carrying it out.

cb1kenobi:
Thanks — dug into all three Codex threads. Each targets behavior that's **identical in the base branch** (#723, `kris/txnlog-committed-position-recovery`): this PR only swaps the recovery scan's whole-file buffer for windowed positional reads and makes an I/O failure throw `DBException` instead of being misread as a torn tail. It changes neither the discard/quarantine policy nor the `MidFileCorruption` watermark semantics.

- **Rotation-spanning uncommitted tail** (`transaction_log_file.cpp` ~L324) and **`eraseTail()` failure only logs** (~L349): unchanged here — you anchored both to lines outside this diff. These are repair-policy calls owned by #723's `load()` seeding (invariant 12).
- **`MidFileCorruption` leaves the watermark behind the break** (`transaction_log_recovery.cpp` ~L174): the `lastCompleteEnd`/`MidFileCorruption` classification is byte-for-byte the base's; the line is only in the diff because the function region shifted. Returning a resync offset and scanning transaction flags past it is a cross-layer change spanning the `RecoveryScan` contract, `TransactionLogStore::load()` watermark seeding, and the TS reader's resync (`findResyncPosition` / `CorruptFrameError`, untouched here) — beyond this focused I/O refactor.

Suggest tracking all three against #723, where the repair-policy and watermark/resync design lives. #3 is the only one whose code sits in a file this PR edits — happy to take it on as a follow-up in this stack if you'd prefer it here.

Unrelated CI note: `Test on Deno (macos-latest)` is red on the flaky `commit-teardown` teardown-race test (Deno is green on ubuntu/windows) — worth a re-run before judging CI.

Generated by Barber AI 🤖

PR #801, opened 2026-08-27
Add 'writeStall' event on RocksDB write-stall condition changes

cb1kenobi:
## Summary

Adds a per-database `'writeStall'` event plus a `db.isWriteStalled()` pull, so an application can detect that RocksDB is throttling or blocking writes instead of discovering it as an unexplained throughput collapse. This is the observability surface for the `dbWriteBufferSize`-oversubscription pathology found while testing #788 (an undersized global memtable budget shared by many column families → constant premature flushes → write stall).

## The event: `'writeStall'` (push, per column family)

Fires when a column family **enters** a stall, with `(columnFamily, previousCondition, currentCondition)` — `currentCondition` is `'delayed'` (rate-limited) or `'stopped'` (blocked).

RocksDB oscillates a CF's stall condition many times per second during a sustained stall, so the event is **rising-edge only and rate-limited**: it fires on entry and then at most once per `ROCKSDB_JS_WRITE_STALL_DEBOUNCE_MS` (default 1000 ms) while the CF keeps stalling. It does **not** push recovery — reliably distinguishing a brief dip from a real recovery isn't possible without a timer thread (deliberately avoided), so recovery is a pull.

The debounce is a Node-free FSM (`core/write_stall_debounce.h`) that advances regardless of listeners (so a detach mid-stall can't strand state) and is unit-tested deterministically in `test/native/write_stall_debounce_test.cc`.

## The pull: `db.isWriteStalled(): boolean`

Authoritative live check over the DB-wide `rocksdb.is-write-stopped` / `rocksdb.actual-delayed-write-rate`. `true` when writes are delayed or stopped. It's the recovery/current-state counterpart to the event, and it's the push/pull mirror of stats that already exist (`getStats()` exposes `rocksdb.stall.micros` and the `rocksdb.db.write.stall` histogram). Database-wide (the write controller is shared across CFs), so it answers "is anything stalling writes" while the event names *which* CF.

```js
db.on('writeStall', (cf, prev, cur) => log.warn(`stall entered on ${cf}: ${prev} -> ${cur}`));
if (db.isWriteStalled()) log.warn('writes currently throttled or blocked');
```

## How it works

`OnStallConditionsChanged` on the existing `TransactionLogEventListener` → `DBDescriptor::emitWriteStall` → the FSM decides (under its own lock, released before `notify`; per-CF ordering is already RocksDB-serialized) → emit via the per-db event emitter. No napi on the background thread; the emit is wrapped so a `bad_alloc` can't unwind into RocksDB. The debounce window is resolved once at `open()` on the JS thread (no `::getenv` on the background thread). Per-CF state is retired on column-family drop.

## Tests

- `test/native/write_stall_debounce_test.cc` — the FSM: rising emits, oscillation rate-limit, blip-doesn't-swallow-next-stall, zero-window, forget, per-CF independence.
- `test/write-stall.test.ts` — end-to-end wiring: a real stall fires a valid rising-edge event; `isWriteStalled()` is wired.
- `test/write-stall-relisten.test.ts` — the detach → recover-unobserved → re-attach → re-emit path (child process, `DEBOUNCE_MS=0`, deterministic).

## Review

Two cross-model rounds (codex + gemini + Harper-domain) plus a human review from @kriszyp. Resolved: the debounce latch/flap (redesign above), the round-1 lock-across-notify over-correction, `::getenv` on the background thread, CF-drop map growth, and the flaky wall-clock test. Documented in README (event + `isWriteStalled`) and AGENTS.md (env var).

Known follow-up (not introduced here): `OnStallConditionsChanged` / `OnFlushBegin` / `OnFlushCompleted` all take a transient `lockDescriptor()` pin without the purge-retry the registry's ownership contract calls for — a pre-existing hazard shared by all three listener callbacks, tracked separately.

## Open decision

Recovery is intentionally pull-only (`isWriteStalled()`), not a pushed `'normal'` event. Punctual pushed recovery would need a per-descriptor timer; can add it if wanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1057,29 +1057,34 @@ db.on('error', (err) => {
 
 ### Event: `'writeStall'`
 
-Emitted when a column family's RocksDB write-stall condition changes — the push signal that writes
-are being throttled or blocked (for example when `dbWriteBufferSize` is too small for the number of
-column families and RocksDB thrashes on premature flushes). Listeners receive three string
-arguments:
+Emitted when a column family **enters** a RocksDB write-stall — the push signal that writes are being
+throttled or blocked (for example when `dbWriteBufferSize` is too small for the number of column
+families and RocksDB thrashes on premature flushes). Listeners receive three string arguments:
 
-- `columnFamily: string` — the column family whose condition changed (`'default'` for the primary).
+- `columnFamily: string` — the column family that entered the stall (`'default'` for the primary).
 - `previousCondition: string` — `'normal' | 'delayed' | 'stopped'`.
-- `currentCondition: string` — `'normal' | 'delayed' | 'stopped'`, where `'delayed'` means writes are
-  being rate-limited and `'stopped'` means they are blocked until a flush frees a memtable/L0 slot.
+- `currentCondition: string` — `'delayed'` (writes rate-limited) or `'stopped'` (writes blocked until
+  a flush frees a memtable/L0 slot).
 
-The event is **edge-triggered and debounced** per column family: the entry into a stall fires
-promptly, and the recovery back to `'normal'` is debounced (`ROCKSDB_JS_WRITE_STALL_DEBOUNCE_MS`,
-default 1000 ms) so a rapidly oscillating condition can't flood listeners. Because it is
-edge-triggered, a listener attached mid-stall may not see a rising edge until the next episode; for
-the authoritative current state, read `db.getDBIntProperty('rocksdb.actual-delayed-write-rate')`
-(nonzero when throttled) or `'rocksdb.is-write-stopped'`.
+The event is **rising-edge only and rate-limited** per column family. RocksDB oscillates a CF's stall
+condition many times per second during a sustained stall, so the event fires when a CF _enters_ a
+stall and then at most once per `ROCKSDB_JS_WRITE_STALL_DEBOUNCE_MS` (default 1000 ms) while it stays
+stalling — never a flood. **Recovery is not pushed**: reliably distinguishing a brief dip from a real
+recovery isn't possible without a timer, so for current state use the pull below rather than waiting
+for a `'normal'` event.
+
+For "are writes stalled right now?", call [`db.isWriteStalled()`](#dbiswritestalled-boolean) — it
+reads the live write-controller state and is authoritative. Note it is **database-wide** (the write
+controller is shared across all column families), so it answers "is anything stalling writes" while
+the event names _which_ CF; a listener attached mid-stall should consult it rather than expect a
+missed rising edge.
 
 ```typescript
 db.on('writeStall', (cf, prev, cur) => {
-	if (cur === 'delayed' || cur === 'stopped') {
-		console.warn(`RocksDB write stall on ${cf}: ${prev} -> ${cur}`);
-	}
+	console.warn(`RocksDB write stall entered on ${cf}: ${prev} -> ${cur}`);
 });
+// ...and to check the current state on demand:
+if (db.isWriteStalled()) console.warn('writes are currently throttled or blocked');
 ```
 
 ## Event API
--- a/README.md
+++ b/README.md
@@ -1055,6 +1055,33 @@ db.on('error', (err) => {
 });
 ```
 
+### Event: `'writeStall'`
+
+Emitted when a column family's RocksDB write-stall condition changes — the push signal that writes
+are being throttled or blocked (for example when `dbWriteBufferSize` is too small for the number of
+column families and RocksDB thrashes on premature flushes). Listeners receive three string
+arguments:
+
+- `columnFamily: string` — the column family whose condition changed (`'default'` for the primary).
+- `previousCondition: string` — `'normal' | 'delayed' | 'stopped'`.
+- `currentCondition: string` — `'normal' | 'delayed' | 'stopped'`, where `'delayed'` means writes are
+  being rate-limited and `'stopped'` means they are blocked until a flush frees a memtable/L0 slot.
+
+The event is **edge-triggered and debounced** per column family: the entry into a stall fires
+promptly, and the recovery back to `'normal'` is debounced (`ROCKSDB_JS_WRITE_STALL_DEBOUNCE_MS`,
+default 1000 ms) so a rapidly oscillating condition can't flood listeners. Because it is
+edge-triggered, a listener attached mid-stall may not see a rising edge until the next episode; for
+the authoritative current state, read `db.getDBIntProperty('rocksdb.actual-delayed-write-rate')`
+(nonzero when throttled) or `'rocksdb.is-write-stopped'`.
+
+```typescript
+db.on('writeStall', (cf, prev, cur) => {
+	if (cur === 'delayed' || cur === 'stopped') {
+		console.warn(`RocksDB write stall on ${cf}: ${prev} -> ${cur}`);
+	}
+});
+```
+
 ## Event API
 
 `rocksdb-js` provides a EventEmitter-like API that lets you asynchronously notify events to one or
--- a/README.md
+++ b/README.md
@@ -631,6 +631,25 @@ const blobFiles = db.getDBIntProperty('rocksdb.num-blob-files');
 const numKeys = db.getDBIntProperty('rocksdb.estimate-num-keys');
 ```
 
+### `db.isWriteStalled(): boolean`
+
+Whether RocksDB is currently applying write backpressure to this database —
+delaying (rate-limiting) or fully stopping writes. Reads the live
+`rocksdb.is-write-stopped` and `rocksdb.actual-delayed-write-rate` properties.
+
+This is the authoritative, live pull counterpart to the
+[`'writeStall'` event](#event-writestall): the event pushes per-column-family
+entries into a stall (rate-limited), while this reports the current state on
+demand and can never be stale. It is **database-wide** — the write controller is
+shared across every column family, so it answers "are writes stalled anywhere"
+rather than for one column family.
+
+```typescript
+if (db.isWriteStalled()) {
+	console.warn('writes are currently throttled or blocked');
+}
+```
+
 ### `db.getRange(options?: IteratorOptions): ExtendedIterable`
 
 Retrieves a range of keys and their values. Supports both synchronous and asynchronous iteration.
--- a/src/binding/database/db_descriptor.cpp
+++ b/src/binding/database/db_descriptor.cpp
@@ -339,7 +339,14 @@ public:
 		if (!desc) {
 			return;
 		}
-		desc->emitWriteStall(info.cf_name, info.condition.prev, info.condition.cur);
+		// Runs on a RocksDB background thread; an exception unwinding into RocksDB
+		// is undefined behavior. A dropped stall notification is acceptable, so
+		// contain anything the emit path throws (e.g. a bad_alloc from the payload
+		// allocation) rather than let it escape.
+		try {
+			desc->emitWriteStall(info.cf_name, info.condition.prev, info.condition.cur);
+		} catch (...) {
+		}
 	}
 
 private:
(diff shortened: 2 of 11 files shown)


cb1kenobi:
## Cross-model review (standard route, `feat/write-stall-event` @ db372569)

**Coverage:** codex (graded, outside-family) + cursor-composer (outside-family) both succeeded. Gemini failed (environmental — `gemini-3.1-pro-high` not available in this location), and the Harper-domain adjudicator didn't run (`claude` is a shell alias, not on `PATH` in the review sandbox), so findings were adjudicated inline. Outside-family coverage is intact via codex + cursor-composer.

### Blocker — fixed in 255d0b17

Both reviewers independently caught a real bug I introduced: `emitWriteStall` gated the **whole** method on `events.hasListeners()`, so the per-CF debounce FSM never advanced while unobserved. Sequence: attach a `writeStall` listener → CF stalls (`stalledReported = true`) → detach the listener (or a `worker_threads` env teardown drops the count to 0) → CF recovers unobserved (early-returns before clearing state) → re-attach → next stall is swallowed as "no change" → **the new listener never gets a rising edge**, permanently.

Fix: run the FSM update unconditionally and gate only the allocation + `notify` on `hasListeners()` (mirroring how `setLastError` gates the emit, not the store). Added `test/write-stall-relisten.test.ts` — a deterministic child-process regression (`ROCKSDB_JS_WRITE_STALL_DEBOUNCE_MS=0`: detach → recover unobserved → re-attach → stall must re-emit); it fails on the pre-fix code.

### Other findings — dispositions

- **Concurrent callbacks could reorder events (major).** Fixed alongside the blocker: the debounce mutex is now held across the emit, so per CF the state transition and the enqueue are atomic.
- **Exceptions escaping into RocksDB (major).** `OnStallConditionsChanged` now wraps the emit in `try/catch(...)` — a `bad_alloc` from the payload path can't unwind into RocksDB's background thread (UB). (The sibling `OnBackgroundError`/`OnFlush*` callbacks have the same pre-existing exposure; out of scope to change here.)
- **Docs (minor).** Fixed: `'writeStall'` documented in README, `ROCKSDB_JS_WRITE_STALL_DEBOUNCE_MS` in AGENTS.md.
- **Comment verbosity (nit).** Trimmed the duplicated rationale.
- **Debounce map not pruned on CF drop (minor).** Accepted for now — bounded by unique CF names over process lifetime; noted as a follow-up.
- **`delayed`↔`stopped` escalation collapsed into "stalled" (significant, by design).** Intentional for the alert semantics; a consumer needing the exact current condition reads `rocksdb.actual-delayed-write-rate` / `rocksdb.is-write-stopped`. Tied to the open event-shape/name question — worth revisiting if we want per-severity signal.

All tests green (`write-stall`, `write-stall-relisten`, `events`, `background-error`); `pnpm check` clean.

---
*Generated by Barber AI (Claude Opus 4.8)*

src/binding/database/db_descriptor.cpp
```diff
@@ -2064,6 +2089,101 @@ void DBDescriptor::setLastError(std::string json) {
...
+	const bool isStalled = current != rocksdb::WriteStallCondition::kNormal;
+	const uint64_t debounceMs = writeStallDebounceMs();
+	const auto now = std::chrono::steady_clock::now();
+	std::lock_guard<std::mutex> lock(this->writeStallDebounceMutex);
+	auto& entry = this->writeStallDebounce[columnFamily];
+	if (isStalled && !entry.stalledReported) {
+		entry.stalledReported = true;
+		entry.lastEmit = now;
+	} else if (!isStalled && entry.stalledReported) {
+		if (debounceMs > 0 && now - entry.lastEmit < std::chrono::milliseconds(debounceMs)) {
```
kriszyp: This drops recovery rather than debouncing it. For example: a rising edge at t=0 sets `stalledReported`; recovery at t=200 ms returns here; if the CF then remains normal, no callback later emits recovery or clears the state. Its next stall is consequently suppressed, and consumers may believe the original stall persisted indefinitely. Please schedule/coalesce a pending recovery and cancel it if the stall resumes, or use another bounded-delivery design that eventually publishes settled state. Add a regression using the default nonzero window that requires both eventual `normal` and a rising event in the following episode—the current relisten test forces the window to zero and accepts any second event, so it bypasses this failure.

src/binding/database/db_descriptor.cpp
```diff
@@ -324,6 +328,27 @@ class TransactionLogEventListener : public rocksdb::EventListener {
...
 	}
 
+	// Surface RocksDB write-stall transitions to JS. Fires when a column family
+	// crosses between kNormal/kDelayed/kStopped — the earliest push signal that
+	// writes are being throttled (kDelayed) or blocked (kStopped), e.g. the
+	// dbWriteBufferSize-oversubscription thrash. Runs on a RocksDB background
+	// thread; emitWriteStall touches no napi (async emit) and debounces, so this
+	// is cheap and non-blocking, same discipline as OnBackgroundError.
+	void OnStallConditionsChanged(const rocksdb::WriteStallInfo& info) override {
+		auto desc = this->state->lockDescriptor();
```
kriszyp, line 338: `lockDescriptor()` adds a transient `shared_ptr` reference. If the final `DBHandle` closes while this callback is running, `PurgeIfUnreferenced()` sees registry + callback references, skips teardown, and this callback releases its reference without the retry required by the registry's ownership contract. The descriptor and open RocksDB can then remain registered indefinitely. Please avoid adding a non-retrying pin here—for example, track actual handle ownership separately and defer teardown until callbacks drain, or arrange a safe deferred purge after releasing the callback pin—and cover a background stall callback racing the final cross-env close.

src/binding/database/db_descriptor.h
```diff
@@ -330,6 +330,40 @@ struct DBDescriptor final : public std::enable_shared_from_this<DBDescriptor> {
...
+	 * (entering delayed/stopped) fires promptly; the falling edge (back to normal)
+	 * is debounced so oscillation dips don't flap. Guarded by
+	 * `writeStallDebounceMutex`; bounded by the column-family count.
+	 */
+	struct WriteStallDebounceEntry {
+		bool stalledReported = false;
+		std::chrono::steady_clock::time_point lastEmit{};
+	};
+	std::mutex writeStallDebounceMutex;
+	std::unordered_map<std::string, WriteStallDebounceEntry> writeStallDebounce;
```
kriszyp: This map is not bounded by the current column-family count: `unregisterColumnFamily()` never removes an entry. Repeatedly creating and dropping unique names grows it for the descriptor's lifetime, and recreating a name can inherit stale `stalledReported` state and suppress its first alert. Please retire the debounce entry as part of successful column-family unregister/drop under the same mutex, accounting for any late callback, and test drop/recreate with the same name.

kriszyp (commented):
I love this, this is an awesome idea/approach!
BTW, are there any stats available in our current stats reporting on write stalls?

> gemini-3.1-pro-high not available in this location

I think you may need to log in to the global region instead of the US region to get access to gemini-3.1-pro

Anyway, I think codex has a couple more fixes to consider.

src/binding/database/db_descriptor.cpp
kriszyp: This drops recovery rather than debouncing it. For example: a rising edge at t=0 sets `stalledReported`; recovery at t=200 ms returns here; if the CF then remains normal, no callback later emits recovery or clears the state. Its next stall is consequently suppressed, and consumers may believe the original stall persisted indefinitely. Please schedule/coalesce a pending recovery and cancel it if the stall resumes, or use another bounded-delivery design that eventually publishes settled state. Add a regression using the default nonzero window that requires both eventual `normal` and a rising event in the following episode—the current relisten test forces the window to zero and accepts any second event, so it bypasses this failure.

src/binding/database/db_descriptor.cpp
kriszyp: `lockDescriptor()` adds a transient `shared_ptr` reference. If the final `DBHandle` closes while this callback is running, `PurgeIfUnreferenced()` sees registry + callback references, skips teardown, and this callback releases its reference without the retry required by the registry's ownership contract. The descriptor and open RocksDB can then remain registered indefinitely. Please avoid adding a non-retrying pin here—for example, track actual handle ownership separately and defer teardown until callbacks drain, or arrange a safe deferred purge after releasing the callback pin—and cover a background stall callback racing the final cross-env close.

src/binding/database/db_descriptor.h
kriszyp: This map is not bounded by the current column-family count: `unregisterColumnFamily()` never removes an entry. Repeatedly creating and dropping unique names grows it for the descriptor's lifetime, and recreating a name can inherit stale `stalledReported` state and suppress its first alert. Please retire the debounce entry as part of successful column-family unregister/drop under the same mutex, accounting for any late callback, and test drop/recreate with the same name.

cb1kenobi:
Thanks Kris — and the region tip worked, Gemini ran cleanly this round.

**On your question — yes, write-stall stats already exist as pull metrics.** `db.getStats()` surfaces cumulative stall time as `rocksdb.stall.micros` and the `rocksdb.db.write.stall` histogram, and `rocksdb.actual-delayed-write-rate` / `rocksdb.is-write-stopped` give the live state. That actually shaped the redesign below: the event is now the **push** counterpart to those **pull** stats, and I added a small `db.isWriteStalled()` wrapper as the ergonomic live check.

Codex's three findings were all real — addressed in `1107b49d`:

1. **Trailing-recovery drop / latched state (major)** — fixed by a redesign. The debounce is now a Node-free FSM (`core/write_stall_debounce.h`, GoogleTested in `test/native/write_stall_debounce_test.cc`): **rising-edge only**, rate-limited to one emit per CF per window, re-armed on the falling edge. No latch, and a short blip can no longer swallow the next genuine stall's alert. **Recovery is a pull, not a dropped push** — a no-timer design genuinely can't tell a brief dip from a real recovery, and pushing recovery punctually would need a per-descriptor timer thread, which we chose to avoid. `db.isWriteStalled()` is the recovery/current-state signal.

2. **`lockDescriptor()` transient pin without purge-retry (major)** — real, but **pre-existing and not introduced here**: it's the identical pin the existing `OnFlushBegin` / `OnFlushCompleted` callbacks already use (same `DBEventListenerState::lockDescriptor`). A correct fix is architectural — a purge-retry from the background-thread callbacks, shared across all three — so I'd rather do it as a focused follow-up than bolt a narrow one-off onto just the new callback. Left the thread open and will file a follow-up issue referencing all three call sites; flagging for your call on whether it blocks this PR.

3. **Debounce map not pruned on CF drop (minor)** — fixed: retired in `unregisterColumnFamily` (which also stops a recreated same-name CF from inheriting stale state).

A parallel cross-model pass (codex + gemini + Harper-domain) corroborated #1 and #3 and caught three more, all fixed here: the round-1 "hold the lock across `notify`" was an over-correction (reverted — per-CF ordering is already RocksDB-serialized, and the header warns the callback must not block), the `::getenv` on the background thread (now resolved once at open on the JS thread), and the wall-clock test's both-directions flakiness (the FSM is proven natively now; the vitest tests cover wiring only).

**One design decision worth your eyes:** recovery is intentionally pull-only (`isWriteStalled()`) rather than a pushed `'normal'` event. If you'd rather have punctual pushed recovery, that's the timer option — happy to add it if you think it's worth the lifecycle machinery.

---
*Generated by Barber AI (Claude Opus 4.8)*

src/binding/database/db_descriptor.cpp
cb1kenobi: Tracked as #807 — this is the identical transient `lockDescriptor()` pin the existing `OnFlushBegin`/`OnFlushCompleted` callbacks already use, so I've filed it as a pre-existing, all-three-call-sites fix rather than a one-off here. Left this thread open for your call on whether it blocks #801.

Generated by Barber AI

src/binding/database/db_descriptor.cpp
```diff
@@ -324,12 +328,37 @@ class TransactionLogEventListener : public rocksdb::EventListener {
...
```
kriszyp: `lockDescriptor()` adds a transient `shared_ptr` reference. If the final `DBHandle` closes while this callback is running, `PurgeIfUnreferenced()` sees registry + callback references, skips teardown, and this callback releases its reference without the retry required by the registry's ownership contract. The descriptor and open RocksDB can then remain registered indefinitely. Please avoid adding a non-retrying pin here—for example, track actual handle ownership separately and defer teardown until callbacks drain, or arrange a safe deferred purge after releasing the callback pin—and cover a background stall callback racing the final cross-env close.

src/binding/database/db_descriptor.cpp
```diff
@@ -2064,6 +2100,74 @@ void DBDescriptor::setLastError(std::string json) {
...
+	return ms;
+}
+
+void DBDescriptor::emitWriteStall(
+	const std::string& columnFamily,
+	rocksdb::WriteStallCondition previous,
+	rocksdb::WriteStallCondition current
+) {
+	// The FSM (WriteStallDebounce) decides emit under its own lock, released before
+	// notify: per-CF ordering is already serialized by RocksDB, and this callback
```
kriszyp: RocksDB can invoke event-listener callbacks from the threads producing those events; it does not provide a same-column-family serialization guarantee that makes unlocking here safe. For example, callback A can decide to emit `normal -> delayed`, pause before `notify()`, while a recovery and later rising callback enqueue first, causing JS to observe the newer alert before A. Please preserve decision-to-enqueue ordering without holding a broad lock across slow work—such as a per-CF dispatch sequence/queue—and add a concurrent callback test.

README.md
```diff
@@ -1055,6 +1055,38 @@ db.on('error', (err) => {
...
+  a flush frees a memtable/L0 slot).
+
+The event is **rising-edge only and rate-limited** per column family. RocksDB oscillates a CF's stall
+condition many times per second during a sustained stall, so the event fires when a CF _enters_ a
+stall and then at most once per `ROCKSDB_JS_WRITE_STALL_DEBOUNCE_MS` (default 1000 ms) while it stays
+stalling — never a flood. **Recovery is not pushed**: reliably distinguishing a brief dip from a real
+recovery isn't possible without a timer, so for current state use the pull below rather than waiting
+for a `'normal'` event.
+
+For "are writes stalled right now?", call [`db.isWriteStalled()`](#dbiswritestalled-boolean) — it
```
kriszyp, line 1095: The `#dbiswritestalled-boolean` heading does not exist in this README, so this new link is dead. Please add the method's API section/heading or link to an existing section.

kriszyp (approved):
Great stuff!

src/binding/database/db_descriptor.cpp
cb1kenobi: Fixed in 994b609f. `WriteStallDebounce::onTransition` now invokes the emit callback **while holding its lock**, so a CF's decision→enqueue is one atomic step — a later transition can't overtake an in-flight one and reorder what JS sees. Since the enqueue is a non-blocking threadsafe-function call, the critical section stays short (not "slow work"). Added a GoogleTest `EmitRunsUnderTheLockAndNeverOverlaps` that hammers one CF from 8 threads and asserts the emit callback is never entered concurrently.

Generated by Barber AI

README.md
cb1kenobi: Fixed in 994b609f — added the `db.isWriteStalled(): boolean` API-reference section, so `#dbiswritestalled-boolean` now resolves.

Generated by Barber AI

Issue #812, opened 2026-08-31
readOnly:true open races a live writer's compaction and reports a healthy database as corrupt; no OpenAsSecondary alternative exists


kriszyp:
## Summary

`RocksDatabase.open(path, { readOnly: true })` maps to [`rocksdb::DB::OpenForReadOnly`](https://github.com/HarperFast/rocksdb-js/blob/main/src/binding/database/db_descriptor.cpp#L1308). That call is **not safe against a database another process is actively writing**: it replays the MANIFEST to build the SST file list and then opens each file, holding no reference on any of them. A compaction in the writing process can unlink an input SST between those two steps, and the open then fails with a message that blames the MANIFEST:

```
Corruption: IO error: No such file or directory:
  While open a file for random read: <path>/000046.sst: No such file or directory
  The file <path>/MANIFEST-000005 may be corrupted.
```

**Nothing is corrupt.** That wording is RocksDB's generic text for *a file named in the MANIFEST was not there when I opened it*. The database is healthy; the reader simply raced a compaction.

Two problems, one root:

1. **The error misdiagnoses a healthy database as corrupt.** This is the part that can cause harm: an operator who reaches for a read-only open during an incident, sees "MANIFEST may be corrupted", and starts a repair or a restore is acting on a false report against intact data.
2. **There is no safe alternative in this library.** RocksDB's API for a follower reader that tolerates the primary deleting files is `DB::OpenAsSecondary` (plus `TryCatchUpWithPrimary()` to advance it). `grep -rn "OpenAsSecondary\|secondary" src/` returns nothing — the option is not exposed at any layer.

## Why this is not just a test concern

`harper` ships an operator-facing read-only mode. [`resources/databases.ts:364-368`](https://github.com/HarperFast/harper/blob/main/resources/databases.ts#L364-L368) sets `options.readOnly = true` whenever `isReadOnlyMode()` is true, which is driven by `HARPER_READONLY=1` or a `--readonly` / `--read-only` argv flag. Nothing stops that being pointed at a data directory a live Harper is writing — "bring up a read-only instance against the live data dir and look at it" is a natural incident move, and it is exactly the case that races.

## Reproduction

Deterministic-ish, no Harper needed:

1. Open a database read-write and write enough to produce several SSTs.
2. Start continuous write + `flush()` traffic so compactions run.
3. From a second handle or process, loop `RocksDatabase.open(path, { name, readOnly: true })` / `close()`.

[...]

PR #814, opened 2026-09-01
Add secondary (live-follower) open mode via secondaryPath + catchUpWithPrimary; report read-only open races accurately instead of as corruption

kriszyp:
Opening a database with `readOnly: true` against a directory another process is actively writing could fail with RocksDB's "The file MANIFEST-… may be corrupted" wording (or a blanket "Database does not exist") when the writer's compaction, blob GC, or flush deleted a file mid-open — a false report against intact data that invites a repair or restore during an incident. This PR fixes the report and adds the mode that makes a live follower actually correct: RocksDB secondary instances.

**Part B — honest classification.** A Node-free classifier ([`isMissingSstOpenRace` in `core/open_status.cpp`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-ea47d3bede9ec86d595f54b8ac74c977092c6b34c37c2a10f089705e295877ceR35), GoogleTest-covered) recognizes a status naming a numbered `.sst`, `.blob`, or `.log` at a filename-token boundary, then the open path confirms that file is absent inside the captured database directory without relying on localized OS error text. The read-only branch throws a distinct [`ERR_CONCURRENT_COMPACTION` error](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1429) that explains the race, preserves the original RocksDB status text, and stops claiming corruption or nonexistence. All three file classes were observed empirically (a parallel writer reclaims compaction inputs, GC'd blob files, and flushed WAL segments). Genuine corruption (checksum mismatch, bad MANIFEST record), a genuinely missing column family, and a genuinely missing database keep their existing failure shapes.

**Part A — secondary mode.** A new `secondaryPath` open option switches the open to [`DB::OpenAsSecondary`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1420): `path` stays the primary's data directory, and `secondaryPath` names the secondary instance's own workspace (created if missing, required to be outside the primary directory). New [`db.catchUpWithPrimary()`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-724f334a2b47a5e7cbcab252ce04fc232071d0bc058c96ed6fd4ac96e5dff308R431) / `db.catchUpWithPrimarySync()` methods replay the primary's MANIFEST + WAL to advance the follower (`ERR_NOT_SECONDARY` on other handles), and `db.secondaryPath` identifies a follower handle. A secondary implies `readOnly` (an explicit `readOnly: false` throws; a `null` secondaryPath is absent, matching the native layer) and forces `maxOpenFiles: -1` — the eagerly opened, fd-held table and blob files are what make the primary's deletions safe, so an explicit bounded value is rejected rather than silently ignored.

Three enforcement pieces are load-bearing. **Registry identity**: `DBKey` gains the workspace, with [`descriptorKey()`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-9c6adfee55db6c03245ed90527ff634ed44cc8d3d1dd743e17f08dfd2f3b9f7eR57) as the single reconstruction point so every close/purge path (CloseDB and the backup/backup-stream/checkpoint purge-retry destructors) finds a secondary's entry instead of leaking it; `DestroyDB` now claims and closes every descriptor for the path. The database path is [resolved to filesystem identity once per open](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R157), and that one string is both registry keys, what RocksDB opens, and what destroy deletes — so two spellings of one directory cannot hold two descriptors and open one workspace twice, which is the in-process half of exclusivity that the kernel lock cannot cover where it degrades. Callers still see the spelling they passed: identity keys the registries, it is not what any API hands back. **Workspace exclusivity**: RocksDB itself accepts a second `OpenAsSecondary` on the same workspace (proved by a native test) even though the instances corrupt each other's state, so the binding takes a [kernel advisory lock on `<secondaryPath>/.secondary.lock`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1409) (the backup-lock utility, with its documented FUSE/NFS degrade caveats) and the registry rejects in-process reuse across primaries. **Read-only log access**: a read-only or secondary open no longer mutates the primary's transaction logs — discovery [skips the retention purge, `recoverTail` truncation, and segment retirement](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-42d2e6ef64957846e9ce8c84051a6fd7e8f6307406deaf7bc1bb3840938b8979R1252) (reader truncation of a live writer's active segment is how acknowledged writes vanish — invariant 5, the harper#2016 class), log files [open with read-only descriptors](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-d98372e40553e26b1d451658a7b2ffd09695ece19306e042f079920d3a69172bR280) and never create, repair, or remove append-boundary markers, and a [writable open is refused](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-aa72a872f0cd6331a10dc7d4e35522a3579fefc637e6b9b6b0baf6f15019cb91R72) while read-only-loaded stores are live in-process. That guard only meets both handles when they land on one registry entry, so the log-store registry keys on the same resolved identity: two spellings of one directory used to open two entries over one log tree and let a writer truncate a segment a reader had mapped, whose next access is an uncatchable SIGBUS.

## For the human reviewer

Judgment calls from the review's decision ledger, with what was chosen and why. Items 1 and 2 were consulted before implementation (needs-input, 2026-09-01) and ruled: proceed with secondary mode over BlobDB backed by the empirical regression net, and spell the API `secondaryPath`.

1. **Shipping over upstream-unsupported BlobDB** (ruled): upstream documents secondary+BlobDB as unsupported ([facebook/rocksdb#13296](https://github.com/facebook/rocksdb/issues/13296)) while this library enables blob files unconditionally. Chosen: ship, with `test/native/secondary_blob_test.cc` as the assertion-only regression net (the pinned build fd-holds blob files exactly like SSTs — verified via `/proc/self/fd`) and the caveat documented in README/TSDoc/AGENTS. Alternative: disable blob files for secondaries — an on-disk format decision. A RocksDB upgrade that regresses this fails the native suite.
2. **API shape** (ruled): `secondaryPath?: string`, presence switches the mode; rejected `secondary: {path}` (no second sub-option exists) and a `mode` enum (collides with optimistic/pessimistic).
3. **A secondary's log view is what this process has resident, not a frozen snapshot**: `catchUpWithPrimary()` advances the RocksDB view only. A store this process never discovered stays not-found (`useLog()` throws rather than conjuring or lazily loading one), and a cross-process primary's new stores and appends need a reopen — but a store an in-process writer holds open is the same object, so its appends are visible. That is not unsafe state: the log write completes before the RocksDB commit for every writer, so log-leads-database is the normal direction and every consumer already tolerates it. Alternative: a uniform snapshot contract, which would mean binding each reader to a private store copy; documented in README, TSDoc and AGENTS instead, so changing it later is a docs-and-behavior break.
4. **Writable open refused while read-only-loaded stores are live in-process**: the alternative (upgrade in place with recovery) truncates under live readers holding maps — the invariant-14 shape. Makes in-process open *order* load-bearing: open the writer before the follower, or close the follower first. The same rule is enforced per store at resolve time, so a store that appears after a writer is already open cannot be adopted by it either — that rejection reaches JS as an error at both `useLog` surfaces (a regression test covers it, because an escaping C++ exception there aborts the process rather than failing a call).
5. **A retired transaction-log segment ends on a transaction boundary.** A Windows zero-fill that cannot land retires the segment instead of leaving it appendable; because a retired segment cannot erase anything, its persisted boundary is its only eraser, so it retires at the last complete transaction rather than at the framing end — otherwise the next segment's first flagged batch closes the orphaned prefix and two source transactions read as one (invariant 14 spans rotations). Erase and retire now share one predicate. Retirement is also decided at load rather than on the first write, which means an unwritable marker fails the open; a reviewer may prefer degrading to read-only there instead.
6. **`destroy()` closes every descriptor for the path**, secondaries included (then errors if references remain). Alternative: reject destroy while a follower is open — a behavior change for existing two-handle callers (a read-write + read-only pair already destroys today).
7. **Hard rejections over silent normalization**: explicit `readOnly: false` or bounded `maxOpenFiles` alongside `secondaryPath` throw. Strictness surfaces config mistakes; the cost is that a shared options object with `maxOpenFiles` set cannot be reused for secondaries verbatim.
8. **Deferred, documented**: concurrent `catchUpWithPrimary()` calls each hold a libuv worker for the whole replay and queue on the per-database mutex while holding it, so overlapping calls can exhaust the default four-thread pool (the shape invariant 16 describes). Coalescing waiters onto one in-flight replay removes it without a new thread and is the obvious follow-up; for now the README and TSDoc say to await one call before starting the next. The registry hazards exposed by this PR are fixed: waits re-find their entries after every wake, keep an explicit closing descriptor visible across spurious wakes, and recheck path-wide workspace ownership; destroy serializes against every registry key for the physical path, so a reader or fresh-workspace secondary cannot open mid-delete. Still left for separate issues (pre-existing): a POSIX `truncateFile()` failure leaves a torn segment appendable where this PR built the symmetric Windows retirement, the registry mutex is held across slow opens, non-race read-only IOErrors still collapse to "Database does not exist", and `catchUpWithPrimary()`'s replay is uncancellable, so a close that waits on it blocks until it finishes.

9. **Known limitation, documented not fixed**: a cross-process follower reads a transaction-log file through a `MAP_SHARED` overlay covering an unrecovered torn tail, so if the primary restarts and its recovery truncates below a mapped page, the follower's scan of that region takes SIGBUS — an uncatchable kill, and precisely when a follower matters most. Windows is protected by mandatory sections; POSIX is not, and invariant 14's "POSIX needs none" reasoning covered only a same-process writer. The fix is to serve read-only stores through positional reads rather than a shared mapping, which is a change to the read path that deserves its own review rather than a late addition here. Recorded in AGENTS invariant 18.

10. **`destroy()` does not delete a secondary's workspace directory.** It closes every follower on the path and releases their locks, but the workspace is a caller-chosen directory outside the database, so removing it is not destroy's to do. A reviewer who wants destroy to clean it up should say so; the alternative is silently deleting a directory the caller named and may reuse.
11. **The cross-key open/destroy serialization is inspection-backed.** The independent reviewer traced the condition-variable and mutex ordering without finding a lost wake or deadlock, but there is not yet a deterministic worker-thread regression that repeatedly opens read-only or fresh-workspace secondary handles while another thread destroys the same physical path. Adding such a stress seam/test is the strongest follow-up; rejecting this choice would mean delaying the fix while the previously demonstrated open-during-delete hole remains. Separately, `test/lock-tracker.test.ts`'s park test failed twice under Bun on Windows during this work and is green now; this branch touches nothing in the park, verification-table or commit path.

12. **Destroy holds the process-global database-registry mutex through RocksDB deletion and workspace removal.** This is deliberate conservative serialization: it prevents a different registry key for the same physical path from opening after descriptor claims are erased but before deletion finishes. The cost is that an unusually slow or large destroy stalls unrelated registry operations process-wide; the independent reviewer found no re-entry or deadlock. Narrowing that critical section later requires a path-wide destruction sentinel that survives entry erasure.

Where to look hardest: the read-only transaction-log load paths (`transaction_log_store.cpp`, `transaction_log_file_posix.cpp`/`_windows.cpp`) — they touch invariant-5 machinery, and the Windows side is untested locally (CI covers it).

## Verification

- [`test/secondary.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-b13532c1cfc155a66620b4a189dff523504d7975c59500e1e94b1c4839ab4cf7R284) runs a follower against a primary in a separate **process** — writing, flushing and compacting while the follower catches up and reads — and asserts the view advances only on catch-up, never backwards, and never fails on a reclaimed file. The [race test](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-b13532c1cfc155a66620b4a189dff523504d7975c59500e1e94b1c4839ab4cf7R613) proves the defect rather than asserting it: a worker-thread writer drives continuous write/flush/compaction traffic while read-only opens race it — measured runs: ~10% of opens fail classified as `ERR_CONCURRENT_COMPACTION` (never corruption, never "does not exist") plus stale-snapshot read failures — and the same churn through secondary opens, catch-ups, and reads runs clean. Catch-up visibility is asserted in both directions, including >2KB blob-resident values; the test skips (never false-fails) on an environment that cannot reproduce the race.
- [`test/readonly.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-9287184cf7d321ee9733c10f18c5d4dc12a6b1601b9ca0ad5ee6d91ba86889f6R226) covers the Part B classification end-to-end plus the new no-mutation guarantees: a read-only open leaves a torn transaction-log tail untouched (byte-for-byte) where a writable open truncates it, a writable open is refused while read-only-loaded stores are live, a writer cannot adopt a read-only-loaded store through either `useLog` surface, and the writable-open refusal holds when the two handles spell the path differently (that last one fails without the identity re-key).
- [`test/native/secondary_blob_test.cc`](https://github.com/HarperFast/rocksdb-js/pull/814/changes?w=1#diff-fe1a371332bc4f15b78bd07330643be3c918c60f45411c2569d7d1dd16be17ebR37) (assertion-only) is the secondary+BlobDB regression net: blob reads through a secondary, catch-up after primary blob GC, cold reads of GC'd blobs with fd-hold evidence on Linux, and the proof that RocksDB does not reject a shared workspace — the reason the kernel lock exists. `test/native/open_status_test.cc` covers the classifier over observed status shapes and corruption/missing-CF/missing-DB counterexamples; `test/native/transaction_log_erase_tail_test.cc` covers the Windows zero-fill fallback, the retire-on-failure path, and that retirement lands on a transaction boundary.
- Full gates on this head: `pnpm build` passed; `pnpm test` (859 passed, 3 skipped), `pnpm test:native` (179 passed), and `pnpm check` clean. The affected `test/secondary.test.ts`, `test/readonly.test.ts`, and `test/destroy.test.ts` suites also passed after the final race-hardening changes. Cross-process workspace exclusion is verified with a spawned child process.

Refs #812

Complexity: complicated

— Claude Fable 5.1

<sub>Review-Coverage: authored=codex; ran=claude; declined=gemini,cursor-grok,cursor-composer,domain; rounds=6 @ 276d38251993</sub>

<sub>Human-Review-Need: 4 @ 276d38251993</sub>


src/binding/database/database.cpp
```diff
@@ -426,6 +426,192 @@ napi_value Database::CompactSync(napi_env env, napi_callback_info info) {
...
+		},
+		[](napi_env env, napi_status status, void* data) { // complete
+			auto state = reinterpret_cast<AsyncCatchUpState*>(data);
+
+			state->deleteAsyncWork();
+
+			if (status != napi_cancelled) {
+				if (state->status.ok()) {
+					napi_value undefined;
+					NAPI_STATUS_THROWS_VOID(::napi_get_undefined(env, &undefined));
```
cb1kenobi, line 625: **High: Cancelled catch-ups leak the close pin**

After work is queued, `handedOff` prevents the RAII claim from decrementing `operationsInFlight`, but the replacement decrement runs only in the execute callback. If close or environment teardown cancels queued work before execution, this completion path deletes the state with the counter still positive, so `finishClose()` waits forever and wedges close/destroy. Simplest fix: make the state own an exactly-once RAII release, or explicitly release the counter in the `napi_cancelled` completion path before purging the descriptor.

—
_Generated by Barber AI_

src/binding/transaction_log/transaction_log_store_registry.cpp
```diff
@@ -141,27 +167,38 @@ void TransactionLogStoreRegistry::DiscoverStores(const std::string& dbPath) {
...
+			}
 			auto store = TransactionLogStore::load(
 				dirEntry.path(),
-				config->transactionLogMaxSize,
-				config->transactionLogRetentionMs,
-				config->transactionLogMaxAgeThreshold
+				config.transactionLogMaxSize,
+				config.transactionLogRetentionMs,
+				config.transactionLogMaxAgeThreshold,
+				config.readOnly
```
cb1kenobi: **Critical: Sticky writer mode lets a secondary mutate primary logs**

`DiscoverStores` passes the path-global `config.readOnly`, but `Register` permanently flips that field to false after any writer and `Unregister` does not restore it. If a secondary keeps the entry alive after that writer closes, another process can become primary and create a log store; opening another secondary then loads it as writable, allowing retention deletion and `recoverTail()` truncation against the live primary. Simplest fix: pass the opening descriptor's `options.readOnly` into `DiscoverStores`/`load` and track writable-store adoption independently.

—
_Generated by Barber AI_

src/binding/transaction_log/transaction_log_store_registry.cpp
```diff
@@ -216,8 +256,17 @@ std::shared_ptr<TransactionLogStore> TransactionLogStoreRegistry::ResolveStore(
 			instance.get(), name.c_str(), dbPath.c_str());
 	}
 
+	// A read-only caller's log view is frozen at open: an unresolved store is
+	// not found — never created (mkdir in what may be a foreign live primary's
+	// tree) and never lazily loaded/published (see the header comment).
+	if (callerReadOnly) {
```
cb1kenobi, line 277: **High: Read-only handles resolve the writer's live log view**

The existing-store lookup returns the path-global store before `callerReadOnly` is checked. With the supported writer-before-secondary ordering, the secondary therefore sees entries appended after it opened while its RocksDB view remains stale until catch-up, so a consumer can process log entries for data absent from its database snapshot. Simplest safe fix: bind each read-only descriptor to a read-only store snapshot captured at open, or reject log reads when only a shared writable store is available.

—
_Generated by Barber AI_

src/binding/database/db_registry.cpp
```diff
@@ -272,34 +299,82 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
...
-			}
-			return true; // database exists and is not closing
+	// Wait for any closing database on this path to be fully removed. The map
+	// node must not be held across the wait: DestroyDB erases every entry for
+	// the path, so a reference into it would dangle and its condition variable
+	// would be destroyed with this thread still parked on it. Re-find the entry
+	// after every wake and park on whatever condition the CURRENT entry has —
+	// an entry erased and re-created while we waited carries a new condition,
+	// and staying on the old one would miss its notify.
+	while (true) {
```
cb1kenobi, line 386: **High: Workspace ownership becomes stale after waiting**

The cross-primary `secondaryPath` scan runs only before this loop releases `databasesMutex`. While this opener waits for its old descriptor to close, another primary can open the same workspace; after waking, this code creates its entry without rechecking ownership. On filesystems where the advisory lock degrades to a no-op, both secondary instances then use and can corrupt the same workspace.

Simplest fix: rerun the cross-primary workspace scan after every wake and immediately before creating or opening the entry.

—
Reviewed 01583e43

src/binding/core/open_status.cpp
```diff
@@ -0,0 +1,65 @@
...
+	}
+	if (status.IsPathNotFound()) {
+		return true;
+	}
+	// POSIX strerror text, Windows FormatMessage text ("The system cannot find
+	// the file specified" / "... the path specified"), and RocksDB's own
+	// subcode name as printed by Status::ToString.
+	return message.find("No such file or directory") != std::string::npos ||
+		message.find("The system cannot find") != std::string::npos ||
+		message.find("PathNotFound") != std::string::npos;
```
cb1kenobi: **Medium: Localized OS errors bypass race classification**

The fallback recognizes only English POSIX and Windows not-found messages. When RocksDB wraps the original status as `Corruption`, its structured subcode is lost; on a non-English host, the localized `strerror`/`FormatMessage` text matches none of these strings, so the same live-writer race is again reported as corruption or nonexistence.

Simplest fix: preserve and classify the original structured `PathNotFound` status before it is wrapped, or extract the named file and verify its absence without depending on localized message text.

—
Reviewed 01583e43

test/secondary.test.ts
```diff
@@ -0,0 +1,628 @@
...
+		const secondaryB = new RocksDatabase(dbPathB, { secondaryPath });
+		try {
+			primaryA.open();
+			primaryB.open();
+			secondaryA.open();
+			// The workspace in the message is the resolved spelling (the native
+			// layer resolves it once so two spellings of one directory cannot
+			// read as two workspaces), which differs from the argument wherever
+			// the temp directory is a symlink (macOS).
+			expect(() => secondaryB.open()).toThrow(
```
cb1kenobi: **Medium: macOS assertion expects the unresolved primary path**

The implementation now resolves the database path as well as the workspace, but this regex still embeds raw `dbPathA`. On macOS, temporary paths commonly change from `/var/...` to `/private/var/...`, so the error contains the resolved `existingKey.path` and this assertion fails across the macOS runtime jobs.

Simplest fix: resolve `dbPathA` for the expected value or match its basename as the workspace assertion already does.

—
Reviewed 01583e43

src/binding/database/db_registry.cpp
```diff
@@ -151,12 +150,23 @@ void DBRegistry::DestroyDB(const std::string& path) {
...
-	std::shared_ptr<DBDescriptor> descriptor;
-	std::shared_ptr<std::condition_variable> condition;
-
-	// Claim the descriptor under the lock but leave the entry in the map until
-	// the close completes (same discipline as CloseDB): the entry is how the
-	// env-cleanup hooks (RemoveListenersByEnv / ReleaseCommitCompletionsByEnv)
+	// Entries are keyed by resolved identity, so a caller's spelling has to be
+	// resolved the same way or destroy would walk past this path's descriptors
+	// and delete their files out from under them.
+	const std::string identityPath = rocksdb_js::resolveIdentityPath(path).string();
```
cb1kenobi: **Critical: Destroy re-resolves a mutable display path**

`DBDescriptor::path` now retains the caller's spelling, while the registry and open target use the captured `identityPath`. If a database is opened through a symlink and that link is repointed before `destroy()`, this line resolves the new target: the original descriptor is not claimed, and `rocksdb::DestroyDB`/`remove_all` delete the newly linked database while the original remains open. Simplest fix: pass the descriptor's captured `identityPath` into `DestroyDB` for instance destruction; resolve raw paths only for destroy calls without an open descriptor.

—
Reviewed cfeda82c

src/binding/database/db_descriptor.cpp
```diff
@@ -1409,8 +1541,8 @@ std::shared_ptr<DBDescriptor> DBDescriptor::open(const std::string& path, const
 	logConfig.transactionLogMaxAgeThreshold = options.transactionLogMaxAgeThreshold;
 	logConfig.transactionLogMaxSize = options.transactionLogMaxSize;
 	logConfig.transactionLogRetentionMs = std::chrono::milliseconds(options.transactionLogRetentionMs);
-	TransactionLogStoreRegistry::Register(path, logConfig);
-	TransactionLogStoreRegistry::DiscoverStores(path);
+	TransactionLogStoreRegistry::Register(descriptor->identityPath, logConfig);
```
cb1kenobi, line 1545: **High: Transaction logs still follow a mutable path**

The registry is keyed by `identityPath`, but `logConfig.transactionLogsPath` retains the caller's relative or symlinked spelling. After `process.chdir()` or a symlink repoint, resolving a new log store writes it under a different directory while RocksDB continues using the captured database identity, splitting committed data from its transaction log. Simplest fix: store the resolved physical transaction-log path in the config and keep a separate display spelling for returned paths.

—
Reviewed 4f873af5

src/binding/database/db_registry.cpp
```diff
@@ -272,34 +299,87 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
...
+			break; // database exists and is not closing
 		}
-		return true; // database doesn't exist, can proceed
-	});
+		DEBUG_LOG("%p DBRegistry::OpenDB Database \"%s\" is closing, waiting for removal\n", instance.get(), path.c_str());
+		// Drop the registry's reference so the closer can finish, then own the
+		// condition across the wait: the closer notifies it after erasing.
+		current.descriptor.reset();
+		std::shared_ptr<std::condition_variable> condition = current.condition;
+		condition->wait(lock);
```
cb1kenobi, line 419: **High: A spurious wake can reopen during close**

The waiter clears `current.descriptor` before this unpredicated condition-variable wait. A permitted spurious wake then sees the null descriptor and proceeds to open while the previous descriptor is still closing, allowing overlapping RocksDB or secondary-workspace access and possible corruption where locking is unavailable. Simplest fix: retain an explicit closing/generation marker and loop until the closer has erased or completed that entry, rather than treating a null descriptor as proof that closing finished.

—
Reviewed 4f873af5

src/binding/transaction_log/transaction_log_store_registry.cpp
```diff
@@ -141,27 +166,41 @@ void TransactionLogStoreRegistry::DiscoverStores(const std::string& dbPath) {
...
+	if (config.transactionLogsPath.empty() || !std::filesystem::exists(config.transactionLogsPath)) {
 		DEBUG_LOG("%p TransactionLogStoreRegistry::DiscoverStores No transaction logs path or directory does not exist for \"%s\"\n",
 			instance.get(), dbPath.c_str());
 		return;
 	}
 
 	std::lock_guard<std::mutex> storeLock(entry->storesMutex);
 
-	for (const auto& dirEntry : std::filesystem::directory_iterator(transactionLogsPath)) {
+	for (const auto& dirEntry : std::filesystem::directory_iterator(config.transactionLogsPath)) {
```
cb1kenobi, line 181: **Medium: Read-only log discovery races primary retention**

A secondary reaches this throwing enumeration after RocksDB opens successfully. If the live primary retires a transaction-log segment while `DiscoverStores` or `TransactionLogStore::load` is processing it, the iterator or subsequent read-only file open throws and aborts the secondary open, making the supported follower mode intermittently unavailable during normal rotation and retention. Simplest fix: for read-only discovery, treat not-found entries that vanish during enumeration/open as a retry or skip them with a bounded rescan, while preserving other I/O failures.

—
Reviewed 7f286960

test/secondary.test.ts
```diff
@@ -0,0 +1,697 @@
...
+			secondary.open();
+
+			const settled = secondary.catchUpWithPrimary().then(
+				() => 'settled',
+				() => 'settled'
+			);
+			// destroy() tears down through finishClose(), which waits on the
+			// in-flight claim rather than the async-work tracker, and throws while
+			// the operation still holds a descriptor reference.
+			expect(() => secondary.destroy()).toThrow();
```
cb1kenobi: **Medium: this teardown test never reaches teardown**

`Database::Destroy` runs `THROW_IF_READONLY((*dbHandle)->descriptor, "Destroy failed: ")` before anything else, and a secondary always has `descriptor->readOnly == true` (`DBRegistry::OpenDB` rejects the open otherwise). So `secondary.destroy()` throws the read-only guard immediately and never reaches `DBRegistry::DestroyDB` or `finishClose()` — `expect(...).toThrow()` passes for the wrong reason.

The next line then does `await expect(settled).resolves.toBe('settled')`, so the replay has already finished by the time `secondary.close()` runs on line 421. Neither destroy nor close ever overlaps an in-flight `catchUpWithPrimary()`, which means the contract this test was written to protect is still uncovered: a double release of the `operationsInFlight` claim underflows the counter `finishClose()` waits on unbounded, and a missed release wedges close and destroy forever. Both regressions would keep this test green.

Suggested fix: tear down through the writable handle while the catch-up is still pending — `primary.destroy()`, the way "should close secondaries on destroy" does — and/or call `secondary.close()` before awaiting `settled` rather than after.

—
Reviewed 20e0eecd

src/binding/transaction_log/transaction_log_file_windows.cpp
```diff
@@ -616,11 +664,93 @@ bool TransactionLogFile::truncateFile(uint32_t newSize) {
...
+	// the stale bytes behind it in one buffered write for any tail smaller than
+	// that — the common case, a single partial entry — and the cache could
+	// still persist the boundary page first. A sector is the smallest unit a
+	// drive writes atomically, so a crash either lands these zeros or leaves the
+	// torn tail; there is no third state.
+	constexpr uint32_t chunkSize = 64 * 1024;
+	constexpr uint32_t sectorSize = 512;
+	uint32_t tailLength = entriesEnd - newSize;
+	std::vector<char> zeros(std::min(chunkSize, tailLength), 0);
+	uint32_t finalChunkEnd = newSize + std::min(sectorSize, tailLength);
```
cb1kenobi: **Medium: the boundary write is sector-sized but not sector-aligned, so the barrier still leaves a reorder window**

`finalChunkEnd = newSize + min(sectorSize, tailLength)` makes the last write 512 bytes long, but `newSize` is an entry/transaction framing offset, not a sector boundary — so that write covers the remainder of one physical sector plus the head of the next (and on a 4Kn drive even an aligned 512-byte write is a read-modify-write of a 4 KiB sector). The comment's premise, "a sector is the smallest unit a drive writes atomically, so a crash either lands these zeros or leaves the torn tail," therefore does not hold for the bytes above `newSize` that fall in the *next* sector: nothing is flushed between them and the boundary byte, which is exactly the ordering the barrier exists to establish.

Failure: on Windows with `SetEndOfFile` blocked by a live mapping, a torn tail (or an `eraseTail` discarding whole entries of an unclosed transaction) larger than 512 bytes, an unclean crash during the final write, and the drive persisting the sector containing `newSize` but not the following one. The next open's recovery then sees a clean zero end-of-entries marker at `newSize` with live entry bytes still behind it, `size` becomes `newSize`, appends resume there, and the first shorter batch leaves those stale bytes reading as an entry — the invariant-5 hazard this path was added to prevent. The window is smaller than the previous 64 KiB version, but it is not closed.

Suggested fix: end the final chunk at the boundary's own sector so every byte in a higher sector is behind the barrier:

```cpp
uint32_t finalChunkEnd = std::min(entriesEnd, (newSize / sectorSize + 1) * sectorSize);
```

Bytes that share `newSize`'s sector genuinely cannot be ordered against it; say that in the comment rather than claiming atomicity.

—
Reviewed e5402db1

cb1kenobi:
Cool! This is dope.

src/binding/database/db_descriptor.cpp
```diff
@@ -1331,16 +1348,146 @@ std::shared_ptr<DBDescriptor> DBDescriptor::open(const std::string& path, const
...
+			}
+		};
+		asciiLower(secondaryStr);
+		asciiLower(primaryStr);
+#endif
+		if (secondaryStr == primaryStr ||
+			(secondaryStr.size() > primaryStr.size() &&
+				secondaryStr.compare(0, primaryStr.size(), primaryStr) == 0 &&
+				secondaryStr[primaryStr.size()] == '/')
+		) {
```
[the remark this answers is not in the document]
kriszyp: Fixed on the branch. The nesting check now handles a primary identity that already ends at a root separator, including POSIX and Windows roots.

src/binding/database/db_registry.cpp
```diff
@@ -272,13 +293,34 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
...
+	// a different primary on the same workspace is rejected.
+	if (!options.secondaryPath.empty()) {
+		for (const auto& [existingKey, existingEntry] : instance->databases) {
+			if (existingKey.secondaryPath == options.secondaryPath && existingKey.path != path) {
+				throw rocksdb_js::DBException(
+					"secondaryPath \"" + options.secondaryPath + "\" is already in use by database \"" +
+					existingKey.path + "\"; each secondary instance requires its own workspace directory"
+				);
+			}
+		}
```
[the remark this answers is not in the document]
kriszyp: Fixed on the branch. Secondary workspaces are resolved once at the API boundary and the captured physical identity is used consistently for conflict checks and registry keys.

src/binding/database/database.cpp
kriszyp: Fixed in 766ff4fa and 01583e43. AsyncCatchUpState owns an exactly-once in-flight release, including napi_cancelled completion, and the teardown regression now exercises the pending operation.

src/binding/transaction_log/transaction_log_store_registry.cpp
kriszyp: Fixed in 206c2d19. Discovery and store creation use the opening handle's mode rather than sticky path-global writer state, so a secondary cannot adopt a writable log store after the original writer closes.

src/binding/transaction_log/transaction_log_store_registry.cpp
kriszyp: Kept intentionally and documented. Transaction-log stores are process-global by physical database identity, and the log is the ordered commit stream shared by primary and followers; a secondary controls when its RocksDB view advances with catchUpWithPrimary(). Creating a separate frozen log snapshot would diverge from that contract and would not make the two independently refreshed views atomic.

src/binding/database/db_registry.cpp
kriszyp: Fixed in 6e401e5a. OpenDB reruns the path-wide secondary-workspace conflict scan after every close wait and immediately before opening/creating the registry entry.

src/binding/core/open_status.cpp
kriszyp: Fixed in 6e401e5a. Race classification no longer depends on localized error text: it extracts a stable numbered RocksDB filename, confines the probe to the captured database identity, and classifies only when that file is actually absent.

test/secondary.test.ts
kriszyp: Fixed in 7f286960. The portability assertions use the resolved physical identity, covering macOS /var to /private/var canonicalization.

src/binding/database/db_registry.cpp
kriszyp: Fixed in 6e401e5a and f3d74fa5. Instance destroy uses the DBHandle's captured physical identity even after close; it never re-resolves the caller-facing symlink spelling. A regression test repoints the symlink and verifies that only the originally opened database is destroyed.

src/binding/database/db_descriptor.cpp
kriszyp: Fixed in 6e401e5a. The transaction-log directory is captured as a physical identity at open, with the original spelling retained separately for display/API results. Tests cover symlink repointing.

src/binding/database/db_registry.cpp
kriszyp: Fixed in 6e401e5a. The descriptor remains explicitly present and closing across the condition-variable wait; spurious wakes loop until close completes and the entry is erased.

src/binding/transaction_log/transaction_log_store_registry.cpp
kriszyp: No change needed after tracing the retention path. Normal retention removes individual retired segments, and read-only load already tolerates a segment disappearing between enumeration and open. Whole-store/root purge is an explicitly caller-coordinated operation, not normal primary retention. The recovery guidance was clarified in 276d3825 while preserving fail-closed behavior for genuine I/O/corruption errors.

test/secondary.test.ts
kriszyp: Fixed in e5402db1. The regression now tears down through the writable primary while catch-up is pending, so it covers both exactly-once release and the unbounded close-wait failure mode.

src/binding/transaction_log/transaction_log_file_windows.cpp
kriszyp: Fixed in 164a0eb0 and 9a9426e7. The final zero-fill chunk ends at the boundary byte's own sector, with overflow-safe arithmetic, and the comment states the same-sector limitation instead of claiming stronger atomicity.

Issue #816, opened 2026-09-01
Committed txnlog reads deliver the unflagged prefix of a transaction torn at a mid-file break


kriszyp:
## Summary

A committed (`readUncommitted: false`) transaction-log read can deliver the well-framed entries that precede a mid-file framing break even though they belong to a transaction that never committed. The break tore the last entry of that transaction, so its `TRANSACTION_LOG_ENTRY_LAST_FLAG` entry was never written and `Transaction::Commit()` never ran (the log write precedes the RocksDB commit); the earlier entries of the same batch are intact frames and are indistinguishable from committed ones by framing alone.

## Shape

```
[header] e1(flag) e2(flag) e3 e4(torn: length overruns) | resync | batch C (flag) ...
```

`e3` and `e4` are one transaction. `e3` is a complete, unflagged frame; `e4` is torn. Batch C was appended later (legacy writer that kept appending after an ENOSPC short write — the shape #748 retired for new writes). A committed read yields `e1, e2, e3`, then throws `CorruptFrameError { resyncPosition: C }`, then yields `C…`. `e3` is a phantom: nothing in the store ever applied it.

## Where it comes from

- Rotated (non-current) files: pre-existing since #750 — the reader resyncs past a break and the committed bound for a rotated file is its `size`, so the prefix was already delivered.
- The current file: reachable since the scan started advancing `lastCompleteTransactionEnd` past a break (the fix for #815). Before that the watermark stopped before the first break and amputated every committed entry after it — a larger loss, which is why the watermark now trusts the resync heuristic — but the torn transaction's prefix was also excluded as a side effect.

## What a proper fix looks like

A hole-aware committed read: the reader needs to know, per file, the offset where the torn group starts (the first entry after the last flagged entry before the break) and where framing resumes, and throw `CorruptFrameError` at the *group start* rather than at the break in committed mode. That needs:

1. `RecoveryScan` to report the torn group start for each break (not only the first break).
2. Per-file storage of that range on `TransactionLogFile`, populated for rotated files too — today only the current file is scanned at open, and older files only until the watermark seed is found, so this adds a lazy scan on first read of a rotated file with a break.
3. A native accessor the JS reader queries when it maps a file (`getLogMemoryMap`), plus one property compare per entry in the read loop.

Uncommitted reads should keep delivering the prefix: in-flight entries are part of that contract.

## Downstream

[...]

PR #817, opened 2026-09-01
fix(txnlog): resume the recovery scan and timestamp index past a mid-file framing break

### LINKED ISSUE #815: Recovery scan and timestamp index stop at a mid-file framing break, hiding every committed entry after it
## Summary

A transaction log with a mid-file framing break (a torn frame that intact, committed entries follow — the shape #748 retirement now prevents for new appends but which existing logs can still carry) is only *partially* readable after the store reopens, even though `query()` has resynced past such breaks since #750. Two native walks of the framing stop at the first break instead of resuming where the reader does:

1. **Committed-read watermark amputation.** `scanTransactionLogForRecovery` classifies the file `MidFileCorruption` at the first break and returns immediately, so `lastCompleteTransactionEnd` — which `load()` seeds `lastCommittedPosition` from (#723) — covers only the entries *before* the break. Every committed read is clamped to that watermark. With frame 39 of 60 torn, a committed drain returns 39 entries and no error: the 20 intact entries after the break are silently unreachable.
2. **Timestamp-index seek blindness.** `findPositionByTimestamp` strides through the broken frame's declared length, which carries `lastIndexedPosition` past `size` and freezes the index. Every seek at or after the break reports "past this file" (`0xFFFFFFFF`), so a consumer that resumes from a timestamp near the break gets nothing even in uncommitted mode.

Together these are what left a replication follower holding 39 of 60 rows after a tear was injected into the leader's log: the leader's sender read to the watermark, reported end-of-log, and parked.

## Reproduce

`test/transaction-log.test.ts` › `crash recovery (truncate-on-open)` › `mid-file break on reopen` (added by the fix PR): write 60 entries, close, overwrite frame 39's length with `0x7fffffff`, reopen, drain. On `main` before the fix the committed drain yields 39 entries with no `CorruptFrameError`; seeking `timestamps[45]` yields nothing. Native: `TransactionLogRecovery.WatermarkAdvancesPastAMidFileBreak`, `TransactionLogTimestampIndex.SeeksPastAMidFileBreak`.

## Expected

Both walks resume at the same offset `findResyncPosition()` in `transaction-log-reader.ts` uses (a run of `RESYNC_MIN_FRAMES` well-formed frames, or a chain landing exactly on the written extent), so the watermark and the index cover every intact entry in the file. The break itself is still surfaced as `CorruptFrameError` with `resyncPosition`; the file is never truncated when intact frames follow the break.

A break in the index walk cannot be an in-flight append: `size` is bumped only after the bytes land, so a nonzero header below `size` is a complete entry and a length overrunning `size` is a break.

## Affected

`@harperfast/rocksdb-js` 2.8.0 only — the committed watermark seeded from the recovery scan arrived in 2.8.0 (#723), and 2.7.1 treats any corrupt frame as terminal regardless. Pinned by harper `main` (v5.3 line); not in a shipped Harper release.

kriszyp:
A transaction log with a mid-file framing break — a torn frame that intact, committed entries follow — is now fully readable after reopen. Two native walks of the framing stopped at the first break even though `query()` has resynced past such breaks since #750: the open-time recovery scan seeded the committed-read watermark (`lastCompleteTransactionEnd`) from the entries *before* the break, so a committed drain of a 60-entry log torn at frame 39 returned 39 entries and no error; and the timestamp index strode through the broken frame's declared length past the written extent and froze, so every seek at or after the break reported "past this file". Both walks now resume at the same offset the JS reader's `findResyncPosition()` uses ([`findFramingResumeOffset()`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-edc5e16c84874b5fcccc16d99d63a57bc55ef0c45f1d72827edba02fdcd926d2R117), shared by [the scan](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-edc5e16c84874b5fcccc16d99d63a57bc55ef0c45f1d72827edba02fdcd926d2R225) and [the index](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R776)), so the watermark covers every complete transaction in the file and the index covers every intact entry. The break is still surfaced as `CorruptFrameError` with `resyncPosition`; a file with a break is never truncated, including when a torn tail follows the break.

The shared resume rule also learned about pre-extended segments: on Windows a segment is zero-padded to its maximum size, and the native walks run before `size` has been corrected to the end of the data, so a chain of fewer than eight frames in front of the padding could never "reach EOF" and classified as a torn tail — recovery would have truncated committed entries. A chain landing exactly on [the end of the nonzero bytes](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-edc5e16c84874b5fcccc16d99d63a57bc55ef0c45f1d72827edba02fdcd926d2R123) now counts the same as one landing on EOF.

Review feedback on this PR found the other half of that accounting in the index walk. It searches the **mapped** region ([`min(size, mapSize)`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R763)), which stops short of the written extent whenever one batch exceeded `transactionLogMaxSize` or the limit was lowered — and an empty search there was treated as "nothing follows the break", parking `lastIndexedPosition` at the extent, past bytes the walk never read. That loss is permanent: a later, larger map resumes from `lastIndexedPosition` and never looks below it. An empty result is now scoped to the bytes that were searchable, so the walk [stays at the break and reports an unindexed tail](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R783) — the same treatment the header bound-check above it already gives a tail the map does not cover — and parks at the extent only when the whole extent was searchable and the break is therefore a torn tail. A short search also [gives up the "chain lands on the written extent" signal](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R777): the region ends on an arbitrary cut, so a chain landing there proves nothing, and accepting one would let a garbage chain inside the corrupt gap resume the walk — whose bogus timestamp then caps this running-maxima index and hides every real entry behind it.

Fixes #815. This is the engine half of the harper-pro `txnlogTearReplication` red (follower stuck at 39/60 rows); harper core and harper-pro follow with their own PRs on top of the 2.8.1 release.

## For the human reviewer

Framing-Verdict: better-alternative-exists (8fc19599da00) — resolved by ruling: system-level resync (core surfaces the mid-log-break halt, harper-pro forces the bounded base-copy) *plus* fix the engine watermark amputation and seek blindness here, rather than either alone.

1. **The committed watermark now trusts the resync heuristic.** [`lastCompleteTransactionEnd` advances past a break](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-edc5e16c84874b5fcccc16d99d63a57bc55ef0c45f1d72827edba02fdcd926d2R233) to the last flagged entry the scan can reach, which means the committed-read bound accepts a run of frames that `findFramingResumeOffset()` judged to be real log data (≥ 8 well-formed frames, or a chain landing exactly on the written extent). Alternative: keep the watermark pinned before the first break and leave only uncommitted reads able to see past it. Chosen because the JS reader already delivers those frames as entries under the same heuristic, so pinning the watermark buys no additional safety, only the 39/60 amputation. A "no" here reverts to the amputation and pushes the whole recovery onto the base-copy resync in harper-pro. Cheap to reverse (one branch in the scan).
2. **Accepted and filed, not fixed here: the unflagged prefix of the torn transaction is now reachable on committed reads of the current file** (#816). When the break tore the last entry of a batch, the batch's earlier entries are intact frames below the advanced watermark, so a committed read delivers them before throwing at the break — entries the source never committed. Pinning the watermark only for that shape would silently re-amputate it (the sender would park with no error), which is worse under the ruling's invariant than a loud break plus a phantom the forced copy repairs; a proper fix is a hole-aware committed read (per-file torn-group range surfaced to the reader, including lazily for rotated files, where the same leak has existed since #750). That is a new engine surface, out of scope for the red fix. Both the PR review and every pre-push round re-raised it as the top finding; the ruling on this task was to keep #817 scoped and land the hole-aware reader in #816.
3. **The index walk treats a frame whose length overruns `size` as a break, never as an in-flight append.** This rests on `writeEntriesV1` bumping `size` only after `writev` returns and on the POSIX overlay exposing bytes only up to an entry boundary, so a nonzero header below `size` is always a complete entry. If that invariant is ever relaxed (a writer publishing `size` before the bytes land), the walk would skip a live entry's index slot until `resetTimestampIndex()`. Same assumption the base code made implicitly by striding `HEADER + length`; this PR makes it explicit in the comment. Where to look hardest: [the resume branch of `findPositionByTimestamp`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R759) and the [`min(size, mapSize)` bound](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R763) on the byte search.
4. **A short map's empty search stops the index at the break; only a full-extent search parks it at the extent.** [The torn-tail branch](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R793) still jumps `lastIndexedPosition` to `size` so later appends get indexed and the torn bytes never do — but only when the whole written extent was searchable, which is the only case that proves nothing follows. Staying at the break costs nothing per seek: the extent that failed is [memoized](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-01368d0a7e6260e12be3080a46522d2a748c253d88a35b493f66f8f082b32e5dR251), since only a larger map can change the answer, and [the guard](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R764) answers a repeat seek without re-scanning the corrupt gap under the store's `dataSetsMutex`. Every assignment that moves `lastIndexedPosition` off the break — including `resetTimestampIndex()` — clears the memo, so the two cannot disagree; that is the invariant to check hardest. A native-test-only counter (`resyncSearchCountForTests`, `ROCKSDB_JS_NATIVE_TESTS` only) makes the elision assertable rather than assumed. The landing rule is switched off for a short search through [`endIsWrittenExtent`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-edc5e16c84874b5fcccc16d99d63a57bc55ef0c45f1d72827edba02fdcd926d2R124), which defaults to true so the recovery scan — which reads the real file — is unchanged.
5. **"End of the nonzero bytes" as a second written-extent signal.** [`nonzeroEnd()`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-edc5e16c84874b5fcccc16d99d63a57bc55ef0c45f1d72827edba02fdcd926d2R42) is computed once per scan, reading the file backwards from its physical size, only when a break has already been found. The residual: a run whose *last payload ends in zero bytes* still misses the signal (its chain lands past the nonzero end) and, if shorter than eight frames and in front of padding, still classifies as torn — Windows-only, and narrower than the base behaviour, which truncated every such run; the rule is recorded in [AGENTS.md invariant 11](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R459). A false positive needs a forged chain landing on that exact offset (one in 2³² per candidate). Alternative: derive the extent from the caller's corrected `size` — but the callers are the ones computing `size`, so the scan has nothing to be handed. Fixing it exactly needs a persisted logical append extent, which is a format change and a follow-up, not part of this red fix; accepting `pos >= nonzeroEnd()` instead would let any forged length landing anywhere in megabytes of padding qualify, which is a much worse trade.
6. **Classification stays pinned to the first break** ([`firstBreak`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-edc5e16c84874b5fcccc16d99d63a57bc55ef0c45f1d72827edba02fdcd926d2R230)) while the walk continues past it (`MidFileCorruption`, `validEnd` = first break) — unchanged from base, which also stopped classifying at the first break and never truncated after it.

Declined: the resync byte search runs under `fileMutex` with no upper bound other than the file (Codex, round 1). It runs once per file per open, only when a break exists, and spans only the corrupt gap; the adjudicator agreed it is a latency cliff on an already-corrupt file, not a hang. The memo above removes the repeat-seek half of the same concern raised against the index walk (which holds the store's `dataSetsMutex`). The overlay-lag concern from the same round ("a live append mistaken for corruption") does not apply: the shared range ends at an entry boundary and the walk never reads a header at or past `size`, so a lagging overlay reads as zeros — the existing end-of-data branch.

Declined: "the partial-map retry cannot recover through the public reader, because a current file is always mapped at `maxFileSize`" (pre-push round 6-9). The retry is what makes recovery possible at all. While the segment is current, `maxFileSize` is the same bound the *reader* maps under (`TransactionLogStore::getMemoryMap`), so no entry past it was readable in the first place; once the segment rotates or is reopened as non-current it is mapped at its own `size`, the whole extent becomes searchable, and the walk — still parked at the break — resumes and indexes the run. Under the previous behaviour `lastIndexedPosition` had already jumped past those entries, so rotation could never recover them.

## Verification

- Fails-on-base: with `src/` reverted to `origin/main` (tests kept), the new behaviour tests in `TransactionLogRecovery.*` / `TransactionLogTimestampIndex.*` fail and the 3 new JS [`mid-file break on reopen`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-22a3c4e9b693561b0d3d6534ae81880ea0dec05f14fc12b7bbc837a5748e59e5R2077) tests fail with `expected 39 to be 59`, `expected 20 to be 58`, `expected 39 to be 59` — the same 39/60 shape as the harper-pro red. The `TransactionLogResumeOffset.*` tests do not compile on base (new symbol). The padding rule proven the same way: with `reachesWrittenExtent` reduced to `pos == fileSize`, [`ShortRunReachingThePaddingIsNotTruncated`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-3113b0b5416781a95ab3ef9fad57195b021e3f38992632b828139c5d6ef24b9aR216) fails (`TruncateTail` instead of `MidFileCorruption`, watermark 69 instead of 177) and `ShortRunBeforeThePaddingIsIndexedAndEndsTheFile` fails (seek reports past-file).
- Fails-on-branch for the new index rule: [`ShortMapDoesNotSkipTheUnsearchedPostBreakTail`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-3113b0b5416781a95ab3ef9fad57195b021e3f38992632b828139c5d6ef24b9aR750) fails with the previous `lastIndexedPosition = writtenExtent` line restored — the short-map seek and both later full-map seeks report `4294967295` (past-file) instead of the break and the run's offsets — and, with the memo guard disabled, the counter assertions fail `2`/`3` against the expected `1`/`2`. [`ShortMapCutIsNotTreatedAsTheWrittenExtent`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-3113b0b5416781a95ab3ef9fad57195b021e3f38992632b828139c5d6ef24b9aR799) and [`ACutIsNotTheWrittenExtent`](https://github.com/HarperFast/rocksdb-js/pull/817/changes?w=1#diff-3113b0b5416781a95ab3ef9fad57195b021e3f38992632b828139c5d6ef24b9aR677) both fail with `endIsWrittenExtent` forced true (the index one resumes at the run instead of holding at the break).
- Green: `pnpm test:native` 180/180; `pnpm test` 837 passed, 3 skipped; `pnpm check` clean. Note for anyone re-running the native suite: `pnpm test:native` rebuilds only when the binary is missing or the RocksDB pin changed, so a C++ edit needs `NATIVE_TEST_REBUILD=1` or it silently tests the previous binary.
- Windows CI (first push): the four `mid-file break on reopen` tests failed on every Windows leg with `errno -4094` (`ERROR_USER_MAPPED_FILE`) opening `1.txnlog`. The tests queried the log before closing the database to collect seek timestamps; that query maps the file and the map outlives `close()` while its entries are reachable, so the truncating rewrite that tears the frame is refused on Windows. They now read the timestamps from the frame headers on disk after close, so nothing is mapped when the file is rewritten (`test/transaction-log.test.ts` `headerTimestamps`). The new native test needs the same care from the other direction: Windows indexes the whole file against a full-extent map inside `openFile()`, so it drops that mapping and the index before asserting. Both are proven by the Windows legs on this push.
- End-to-end route: harper-pro `integrationTests/cluster/txnlogTearReplication.test.mjs` against a core pinned to this build, in the harper-pro PR that lands on top (its own PR body carries the 60/60 result).

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=codex; blocked=gemini(auth); declined=cursor-grok,cursor-composer,domain; rounds=11 @ 7012af701590</sub>

<sub>Human-Review-Need: 4 @ 7012af701590</sub>

### DIFF
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -713,7 +713,15 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 	// Set when indexing stops early at a committed-but-not-yet-visible tail (a concurrent append we
 	// couldn't read this pass); used below to start the scan at lastIndexedPosition rather than EOF.
 	bool stoppedAtUnindexedTail = false;
-	while (this->lastIndexedPosition < this->size) {
+	while (true) {
+		uint32_t writtenExtent = this->size.load(std::memory_order_relaxed);
+		if (this->lastIndexedPosition >= writtenExtent) {
+			break;
+		}
+		if (static_cast<uint64_t>(this->lastIndexedPosition) + TRANSACTION_LOG_ENTRY_HEADER_SIZE > memoryMap->mapSize) {
+			stoppedAtUnindexedTail = true;
+			break;
+		}
 		double entryTimestamp = readDoubleBE(mappedFile + this->lastIndexedPosition);
 		// The header's own timestamp slot is not an entry, so a legitimate value of exactly
 		// zero there (e.g. an unset/epoch file timestamp) must not be mistaken for the
@@ -745,6 +753,20 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 			}
 			break;
 		}
+		uint32_t entryLength = readUint32BE(mappedFile + this->lastIndexedPosition + 8);
+		if (entryLength == 0 ||
+			static_cast<uint64_t>(this->lastIndexedPosition) + TRANSACTION_LOG_ENTRY_HEADER_SIZE + entryLength > writtenExtent) {
+			// A framing break. It cannot be an in-flight append: size is bumped only after the
+			// bytes are written, so a non-zero header inside the written extent is a complete
+			// entry. Striding through the declared length would carry the walk past every
+			// later entry and freeze the index there, so resume where framing does (the same
+			// rule readers use for CorruptFrameError.resyncPosition) or, for a torn tail, at
+			// the written extent so later appends still get indexed.
+			uint32_t searchable = std::min(writtenExtent, static_cast<uint32_t>(memoryMap->mapSize));
+			uint32_t resume = findFramingResumeOffset(mappedFile, searchable, this->lastIndexedPosition + 1);
+			this->lastIndexedPosition = resume != 0 ? resume : writtenExtent;
+			continue;
+		}
 		// check that the timestamp is greater than any previously indexed timestamp,
 		// otherwise we don't record it, because we want to start at the first position with a timestamp that
 		// is greater than the requested timestamp:
@@ -752,8 +774,7 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 			// insert with a hint to go at the end (constant time?)
 			positionByTimestampIndex.insert(positionByTimestampIndex.end(), {entryTimestamp, this->lastIndexedPosition});
 		}
-		// read size of the entry and move on
-		this->lastIndexedPosition += TRANSACTION_LOG_ENTRY_HEADER_SIZE + readUint32BE(mappedFile + this->lastIndexedPosition + 8);
+		this->lastIndexedPosition += TRANSACTION_LOG_ENTRY_HEADER_SIZE + entryLength;
 	}
 	// now do the actual search: just a search for the lower bound
 	auto it = this->positionByTimestampIndex.lower_bound(timestamp);
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -768,6 +768,9 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 				stoppedAtUnindexedTail = true;
 				break;
 			}
+#ifdef ROCKSDB_JS_NATIVE_TESTS
+			++this->resyncSearchCountForTests;
+#endif
 			uint32_t resume = findFramingResumeOffset(mappedFile, searchable, this->lastIndexedPosition + 1);
 			if (resume != 0) {
 				this->lastIndexedPosition = resume;
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -470,6 +470,7 @@ void TransactionLogFile::resetTimestampIndex() {
 	std::lock_guard<std::mutex> indexLock(this->indexMutex);
 	this->positionByTimestampIndex.clear();
 	this->lastIndexedPosition = TRANSACTION_LOG_FILE_TIMESTAMP_POSITION;
+	this->resyncSearchedExtent = 0;
 }
 
 uint32_t TransactionLogFile::countEntries() const {
@@ -760,9 +761,17 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 			// land, so a nonzero header below the written extent is a complete entry. Resume
 			// where framing does, or at the written extent so later appends are still indexed.
 			uint32_t searchable = std::min(writtenExtent, static_cast<uint32_t>(memoryMap->mapSize));
+			if (searchable <= this->resyncSearchedExtent) {
+				// These bytes already failed to yield a resume for this same break; only a larger
+				// map can change the answer, and the search spans the whole corrupt gap under the
+				// store's dataSetsMutex, so repeating it per seek would stall the store.
+				stoppedAtUnindexedTail = true;
+				break;
+			}
 			uint32_t resume = findFramingResumeOffset(mappedFile, searchable, this->lastIndexedPosition + 1);
 			if (resume != 0) {
 				this->lastIndexedPosition = resume;
+				this->resyncSearchedExtent = 0;
 				continue;
 			}
 			if (searchable < writtenExtent) {
@@ -772,12 +781,14 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 				// never look below it, so entries there stay unindexed and seeks past them report
 				// past-file. Stop at the break instead (the same treatment the header bound-check
 				// above gives a tail the map doesn't cover) and re-search once the map grows.
+				this->resyncSearchedExtent = searchable;
 				stoppedAtUnindexedTail = true;
 				break;
 			}
 			// The whole written extent was searchable and nothing resumes: a torn tail. Park at
 			// the written extent so later appends are still indexed; the torn bytes never are.
 			this->lastIndexedPosition = writtenExtent;
+			this->resyncSearchedExtent = 0;
 			continue;
 		}
 		// check that the timestamp is greater than any previously indexed timestamp,
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -761,7 +761,23 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 			// where framing does, or at the written extent so later appends are still indexed.
 			uint32_t searchable = std::min(writtenExtent, static_cast<uint32_t>(memoryMap->mapSize));
 			uint32_t resume = findFramingResumeOffset(mappedFile, searchable, this->lastIndexedPosition + 1);
-			this->lastIndexedPosition = resume != 0 ? resume : writtenExtent;
+			if (resume != 0) {
+				this->lastIndexedPosition = resume;
+				continue;
+			}
+			if (searchable < writtenExtent) {
+				// The map stops short of the written extent, so "nothing resumes" only rules out
+				// the bytes we could read. Parking at the written extent would strand the rest
+				// permanently — a later, larger map resumes from lastIndexedPosition and would
+				// never look below it, so entries there stay unindexed and seeks past them report
+				// past-file. Stop at the break instead (the same treatment the header bound-check
+				// above gives a tail the map doesn't cover) and re-search once the map grows.
+				stoppedAtUnindexedTail = true;
+				break;
+			}
+			// The whole written extent was searchable and nothing resumes: a torn tail. Park at
+			// the written extent so later appends are still indexed; the torn bytes never are.
+			this->lastIndexedPosition = writtenExtent;
 			continue;
 		}
 		// check that the timestamp is greater than any previously indexed timestamp,
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -758,9 +758,9 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 		if (entryLength == 0 ||
 			static_cast<uint64_t>(this->lastIndexedPosition) + TRANSACTION_LOG_ENTRY_HEADER_SIZE + entryLength > writtenExtent) {
 			// A framing break, not an in-flight append: size is bumped only after the bytes
-			// land, so a nonzero header below the written extent is a complete entry. Resume
-			// where framing does, or at the written extent so later appends are still indexed.
+			// land, so a nonzero header below the written extent is a complete entry.
 			uint32_t searchable = std::min(writtenExtent, static_cast<uint32_t>(memoryMap->mapSize));
+			bool searchedWholeExtent = searchable == writtenExtent;
 			if (searchable <= this->resyncSearchedExtent) {
 				// These bytes already failed to yield a resume for this same break; only a larger
 				// map can change the answer, and the search spans the whole corrupt gap under the
@@ -771,13 +771,16 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 #ifdef ROCKSDB_JS_NATIVE_TESTS
 			++this->resyncSearchCountForTests;
 #endif
-			uint32_t resume = findFramingResumeOffset(mappedFile, searchable, this->lastIndexedPosition + 1);
+			// A chain landing on the end of a short map lands on an arbitrary cut, not on the
+			// end of the data, so only the frame-run signal is trustworthy there.
+			uint32_t resume = findFramingResumeOffset(
+				mappedFile, searchable, this->lastIndexedPosition + 1, searchedWholeExtent);
 			if (resume != 0) {
 				this->lastIndexedPosition = resume;
 				this->resyncSearchedExtent = 0;
 				continue;
 			}
-			if (searchable < writtenExtent) {
+			if (!searchedWholeExtent) {
 				// The map stops short of the written extent, so "nothing resumes" only rules out
 				// the bytes we could read. Parking at the written extent would strand the rest
 				// permanently — a later, larger map resumes from lastIndexedPosition and would
--- a/src/binding/transaction_log/transaction_log_file.cpp
+++ b/src/binding/transaction_log/transaction_log_file.cpp
@@ -718,11 +718,6 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t 
 		if (this->lastIndexedPosition >= writtenExtent) {
 			break;
 		}
-		if (static_cast<uint64_t>(this->lastIndexedPosition) + TRANSACTION_LOG_ENTRY_HEADER_SIZE > memoryMap->mapSize) {
-			stoppedAtUnindexedTail = true;
-			break;
-		}
-		double entryTimestamp = readDoubleBE(mappedFile + this->lastIndexedPosition);
 		// The header's own timestamp slot is not an entry, so a legitimate value of exactly
 		// zero there (e.g. an unset/epoch file timestamp) must not be mistaken for the
 		// zero-padding end-of-data marker below — that heuristic only applies once we're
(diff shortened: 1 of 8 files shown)


kriszyp:
I believe this resolves the broken CI failures on harper-pro main right now.

src/binding/transaction_log/transaction_log_recovery.cpp
```diff
@@ -170,11 +220,19 @@ RecoveryScan scanTransactionLogForRecovery(
...
+			// walk resumes where framing does. A torn tail has nothing valid behind it.
+			uint32_t resume = findFramingResumeOffset(source, pos + 1);
+			if (resume == 0) {
+				return scan(RecoveryScan::Kind::TruncateTail, pos);
 			}
-			return scan(RecoveryScan::Kind::TruncateTail, pos);
+			if (firstBreak == 0) {
+				firstBreak = pos;
+			}
+			pos = resume;
```
cb1kenobi, line 233: Critical — committed reads expose a torn transaction's prefix. Resuming here lets a later LAST_FLAG move the single contiguous lastCompleteTransactionEnd beyond the break; if the broken frame ended an uncommitted batch, its intact prefix is then emitted as committed before CorruptFrameError, creating phantom writes or replication. Simplest fix: keep the committed watermark before the first break; if post-break commits must remain readable, add a torn-group skip range that committed readers exclude before advancing it.

—
_Generated by Barber AI_

src/binding/transaction_log/transaction_log_recovery.cpp
```diff
@@ -78,17 +107,23 @@ bool headerLooksLikeFrame(const char* header, uint32_t pos, uint32_t fileSize) {
...
+// offsets are served from a 64 KiB window; chain hops (HEADER+length) read a
+// 13-byte header so a large payload is not pulled in. A failed read throws — it
+// must not look like "no resume".
+uint32_t findFramingResumeOffset(ScanReader& source, uint32_t from) {
 	std::vector<char> window(RESYNC_WINDOW);
 	uint32_t windowStart = 0;
 	uint32_t windowLen = 0;
 	char headerBuf[TRANSACTION_LOG_ENTRY_HEADER_SIZE];
 
+	auto reachesWrittenExtent = [&](uint32_t pos) -> bool {
```
cb1kenobi, line 123: High — trailing-zero payloads are mistaken for a torn tail. nonzeroEnd() returns the byte after the last nonzero value, not necessarily the start of preallocation padding; when the final committed frame's payload ends in zero bytes and the resumed run has fewer than eight frames, pos lands after this value, resync returns 0, and recovery truncates valid committed entries at the break. Simplest fix: determine or persist the exact logical append extent independently of payload contents and compare against that boundary.

—
_Generated by Barber AI_

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -745,15 +753,25 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t
...
 		}
+		uint32_t entryLength = readUint32BE(mappedFile + this->lastIndexedPosition + 8);
+		if (entryLength == 0 ||
+			static_cast<uint64_t>(this->lastIndexedPosition) + TRANSACTION_LOG_ENTRY_HEADER_SIZE + entryLength > writtenExtent) {
+			// A framing break, not an in-flight append: size is bumped only after the bytes
+			// land, so a nonzero header below the written extent is a complete entry. Resume
+			// where framing does, or at the written extent so later appends are still indexed.
+			uint32_t searchable = std::min(writtenExtent, static_cast<uint32_t>(memoryMap->mapSize));
+			uint32_t resume = findFramingResumeOffset(mappedFile, searchable, this->lastIndexedPosition + 1);
+			this->lastIndexedPosition = resume != 0 ? resume : writtenExtent;
```
cb1kenobi: High — mapping lag permanently skips the post-break index tail. When writtenExtent exceeds memoryMap->mapSize and the mapped bytes contain the break but not a qualifying resumed chain, resume is 0 and this advances lastIndexedPosition across entries that were never inspected; later remaps therefore cannot index them, so timestamp seeks can falsely report past-file. Simplest fix: when searchable < writtenExtent and no resume is found, set stoppedAtUnindexedTail and leave lastIndexedPosition at the break; advance to writtenExtent only when the full extent was searchable.

—
_Generated by Barber AI_

kriszyp: Fixed in `9fc36a2b` as suggested: when `searchable < writtenExtent` the walk now stops at the break and reports an unindexed tail, and `lastIndexedPosition` is parked at the written extent only when the whole extent was searchable (a genuine torn tail). A native test (`ShortMapDoesNotSkipTheUnsearchedPostBreakTail`) covers it and reports `4294967295` on the pre-fix code. One addition on top: staying at the break would otherwise re-run the byte-wise gap scan under the store's `dataSetsMutex` on every seek, so the failed extent is memoized in `resyncSearchedExtent` and cleared by every assignment that moves `lastIndexedPosition` off the break.

— Claude Opus 5

src/binding/transaction_log/transaction_log_recovery.cpp
kriszyp: Deliberately not fixed here — this is #816, and the ruling on this task was to keep #817 scoped to the #815 recovery fix. Both suggested shapes were weighed: pinning the committed watermark before the first break restores the 39/60 amputation this PR exists to remove (a stuck sender with no error, which is worse than a loud break the forced base-copy repairs), and the torn-group skip range is the hole-aware committed read that #816 tracks — a new engine surface, and the same leak has existed for rotated files since #750. Left unresolved on purpose so it stays visible.

— Claude Opus 5

src/binding/transaction_log/transaction_log_recovery.cpp
kriszyp: Accurate, and deliberately not fixed here: doing it exactly needs the persisted logical append extent you describe, which is a format change rather than part of this red fix. The scope is narrower than the base behaviour it replaces (which truncated *every* such run, not only one whose final payload ends in zero and is shorter than eight frames), and it is Windows-only since a POSIX segment ends at its physical size. Loosening `reachesWrittenExtent` to `pos >= nonzeroEnd()` instead would accept any forged length landing anywhere in megabytes of padding, so that is not a safe cheap version. Left unresolved so the follow-up stays visible; noted for issue triage.

— Claude Opus 5

src/binding/transaction_log/transaction_log_file.cpp
```diff
@@ -745,15 +754,54 @@ uint32_t TransactionLogFile::findPositionByTimestamp(double timestamp, uint32_t
...
+				// These bytes already failed to yield a resume for this same break; only a larger
+				// map can change the answer, and the search spans the whole corrupt gap under the
+				// store's dataSetsMutex, so repeating it per seek would stall the store.
+				stoppedAtUnindexedTail = true;
+				break;
+			}
+#ifdef ROCKSDB_JS_NATIVE_TESTS
+			++this->resyncSearchCountForTests;
+#endif
+			uint32_t resume = findFramingResumeOffset(mappedFile, searchable, this->lastIndexedPosition + 1);
```
cb1kenobi: High — Partial map boundary is mistaken for EOF

Passing `searchable` as `fileSize` makes `findFramingResumeOffset()` accept a short frame chain that merely ends at the mapping boundary, even when `searchable < writtenExtent`. Bytes inside the corrupt region can therefore be indexed as a resumed frame although the reader, using the true written extent, would reject them; the resulting persistent timestamp entry can make seeks start in corruption or bypass valid entries. Simplest fix: pass separate readable and written extents, bounding reads by the former while allowing the short-chain exact-end rule only at the latter.

—
Reviewed 9fc36a2b

kriszyp: Accurate, and fixed in 7012af70.

`findFramingResumeOffset()` now takes `endIsWrittenExtent` (default true, so the recovery scan — which reads the real file — is unchanged), and the index walk passes `searchable == writtenExtent`. When the map stops short the landing rule is dropped rather than re-pointed at the true extent, because reads are already bounded by `searchable`: a chain can only ever land at or below the cut, so the rule could not fire truthfully there — it could only fire *on* the cut, which is the false positive you describe. `nonzeroEnd()` is derived from the same short region, so the same flag has to disable it too. A short map is then left with the `RESYNC_MIN_FRAMES` run as its only signal, which is evidence rather than coincidence; the walk stays parked at the break (memoized by `resyncSearchedExtent`) until a larger map makes the rest searchable.

Covered by two new native tests, both proven to fail with `endIsWrittenExtent` forced true: `TransactionLogResumeOffset.ACutIsNotTheWrittenExtent` and `TransactionLogTimestampIndex.ShortMapCutIsNotTreatedAsTheWrittenExtent` (the latter resumes at the garbage run instead of holding at the break).

— Claude Opus 5

cb1kenobi (approved):
LGTM!

PR #820, opened 2026-09-03
Release a purged transaction-log segment's mapping instead of refusing to read it

kriszyp:
`purgeLogs()` removes a retired segment's pathname, but an iterator that already holds its immutable mapping can safely finish reading that committed history. This PR keeps that behavior while making the mapping collectible and ensuring readers resume at the next real survivor after a purged run.

## What changed

- `TransactionLog._currentLogBuffer` is weak, matching the per-segment cache, so a long-lived iterator no longer pins a deleted mapping after it advances.
- [`nextReadableLogBuffer()` walks the registered successor set](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-453de70d96caada6355361ee1c44c7cdbaaa7cd276faab6b49150f4310b802e6R448) instead of deriving a jump from the newest contiguous run. It crosses multiple holes without skipping an intervening survivor, and stops rather than skipping a registered segment that is only temporarily unmappable.
- Read probes use [`TransactionLogFile::openExisting()`](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R151), backed by `O_RDWR` without `O_CREAT` on POSIX and `OPEN_EXISTING` on Windows. A segment removed between discovery and the OS open therefore remains absent instead of being recreated as a header-only ghost. [`TransactionLogStore::load()`'s two startup opens](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-42d2e6ef64957846e9ce8c84051a6fd7e8f6307406deaf7bc1bb3840938b8979R1398) (activating the surviving current segment, and the backward recovery-boundary scan) now use the same non-creating open, closing the two remaining ghost-recreate paths this round's review found.
- A mapping carries the append-owned readable extent, exposed through a type-tagged N-API getter. [`readableExtent()` requires that exact bound](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-453de70d96caada6355361ee1c44c7cdbaaa7cd276faab6b49150f4310b802e6R480), so purge cannot make a reader infer authority from frame-shaped bytes left by a failed append or from unused mapped capacity. [`publishReadableExtentLocked()`](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-70d82772a7a297690416ca9fa571592a9b9d2e0e8b91b58a2a1541ebc173ad74R778), which runs on every append, now reads the live map through a raw pointer instead of copying the `shared_ptr`, so the commit hot path pays no refcount traffic in the common case.
- An all-segment purge advances a store generation and clears old flush correlations. [`writeFlushedPosition()` reopens and verifies `txn.state` by pathname](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-42d2e6ef64957846e9ce8c84051a6fd7e8f6307406deaf7bc1bb3840938b8979R1219), preventing a pre-purge callback or an unlinked/replaced stream from publishing stale retention state.
- Purge-vs-mapping is now tested as a contract rather than a platform outcome: [a reader keeps every entry it mapped and retention reclaims the segment once nothing maps it](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-22a3c4e9b693561b0d3d6534ae81880ea0dec05f14fc12b7bbc837a5748e59e5R3020), and [an unwritable store directory forces the refused-unlink branch](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-22a3c4e9b693561b0d3d6534ae81880ea0dec05f14fc12b7bbc837a5748e59e5R3063) — the segment stays registered, the `log.warn` line is asserted end to end through the public global-event API, and the next run reclaims it. `lastRemoveError` is read through a `fileMutex`-held accessor rather than unlocked.
- [`README.md`'s memory-map section](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R1945) describes the actual ownership split: only the current file's map is held natively; a frozen (rotated) file's map is weakly cached on **POSIX** and survives solely through JS `Buffer` references, so `stats.memory.activeMaps` counts only the current file there. **On Windows the weak-for-frozen optimization does not apply** — every frozen read re-pins the mapping strongly for the file's life, and `activeMaps`/`mappedBytes` count it — which this maintenance round's review caught the README overgeneralizing past POSIX.

## For the human reviewer

1. A purged segment intentionally remains readable through a mapping acquired before purge. The mapping is immutable committed history; the fix releases it after the iterator advances rather than invalidating it mid-read. The open decision underneath it is the reader-gap policy: a subscription that falls behind retention now skips the deleted run and keeps yielding with **no** signal — no throw, no `log.warn`, no marker on the resumed entry — so a follower that lost segments 6–8 believes its history is contiguous. Stopping instead is the wedge this PR fixes, so the choice is between silent skip and a new reader-visible signal Harper would have to handle; it is cheaper to decide now than to retrofit.
2. The hole policy intentionally distinguishes absence from temporary unavailability: absent registered paths are skipped through the successor lookup, while a segment that still has a durable extent but cannot currently map causes the iterator to stop and retry later.
3. `purgeLogs({ destroy: true })` can restart segment numbering while a JavaScript buffer cache from the prior store generation survives. That pre-existing cache-identity problem is outside this PR; Harper does not use destructive log-store purge in production.
4. Windows purge-vs-mapping is unexplained, and is no longer asserted. A Windows-only test here claimed a live reader mapping refuses the unlink; it passed on all three Windows Node jobs at `9a8606a8` and failed on all three (both attempts) at `c8e790a0`, with nothing in between changing how long the mapping lives. Only the convergence is asserted cross-platform now; the first-run deletion is asserted on POSIX only. Worth a word if you know which Windows rule applies here.
5. **Real, deliberately not fixed here — a refused unlink of the *current* segment strands a reader on a stale mapping.** [`removeFileLocked()`](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-d98372e40553e26b1d451658a7b2ffd09695ece19306e042f079920d3a69172bR510) drops its strong `memoryMap` and closes the fd *before* it knows whether the unlink succeeded. If `purgeLogs({ destroy: true })` fails to delete the current segment (`EACCES`, or a Windows sharing violation) while a bound transaction keeps the store alive, the next append reopens the file and builds a *new* `MemoryMap`; a reader still holding the old buffer never sees the new entries, because its extent and overlay stop advancing. Confirmed independently by codex, cursor-grok and the harper-domain adjudicator across four rounds now (all "Scope: in-scope"), but reachable only through `purgeLogs({ destroy: true })`, which Harper does not run in production, and squarely inside the `destroy` path this PR has deliberately left untouched since an earlier scope correction — every attempt at hardening that path in earlier rounds kept surfacing a *new* destroy-path race and was reverted. Filing this as a separate issue rather than reopening that path here.
6. **Pre-existing, narrow — [`txn.state`'s 8-byte record](https://github.com/HarperFast/rocksdb-js/pull/820/changes?w=1#diff-42d2e6ef64957846e9ce8c84051a6fd7e8f6307406deaf7bc1bb3840938b8979R1288) is overwritten in place with no fsync, no temp-then-rename, and no checksum.** A crash mid-write can leave a torn `LogPosition` (e.g. a newer sequence number paired with an older offset); retention would then trust it as the flushed boundary and could delete a segment whose tail never reached RocksDB. This predates this PR (the old code overwrote the same record the same way) and the practical exposure is narrow — a torn write inside one 8-byte range is rare on filesystems with ordered/journaled metadata — but it is a real gap in the hardening this PR added around `txn.state`. Reviewed and left alone rather than restructuring that file's durability inside this PR; a proper fix is write-temp-then-atomic-rename.
7. **Rebased onto `origin/main` again as scheduled PR maintenance (round 34–35); no behavioral change.** `origin/main` had advanced 18 commits (to `dcca4ed3`) since the prior rebase, including its own new invariant 24 ("two process-wide clocks, two contracts"), which collided with this branch's purge invariant — previously also numbered 24 after the last rebase's renumbering. Resolved by keeping main's 24 and moving this branch's two invariants down: purge to 25, `databaseFlushed`/txn.state to 26 (the latter had already been 25 from the prior rebase, so it also had to move). A follow-up commit fixed three stale "invariant 24" self-references (`AGENTS.md`, `transaction-log-reader.ts`, `transaction_log_store.cpp`) that survived the conflict resolution because git had no way to flag prose that doesn't textually collide — caught by re-grepping the whole tree after the rebase completed. `git merge-tree --write-tree origin/main HEAD` matches `HEAD^{tree}`, so CI builds exactly this branch's tree; `origin/main` is a full ancestor of `HEAD` (20 commits ahead, 0 behind).
8. A full round of independent review on the rebased head (round 34) re-raised the 12 findings still open from before, all still previously-adjudicated or pre-existing, plus one genuinely new item: the README's frozen-map weak-ownership claim overgeneralized POSIX semantics to Windows (Windows re-pins every frozen read strongly, by design — see the comment above `TransactionLogFile::getMemoryMapLocked()` on that platform, and `getStats()` counts it into `activeMaps`). Fixed and confirmed resolved in round 35 (delta): the README now scopes weak ownership and `activeMaps`/`mappedBytes` accounting to POSIX and states the Windows exception.
9. `transaction-log-reader.ts:419-429` caches `logBuffer.size` after an advance **without** re-checking `latestLogId > logBuffer.logId`, unlike its sibling at `:289-296`. If the advance lands on the still-current segment, a `HEADER_SIZE`-sized snapshot is stamped onto the shared `_logBuffers` entry and every entry later written to that segment is skipped once it rotates. This is **pre-existing** — `main` has the identical shape with `getLogFileSize()` in place of `readableExtent()` — so it is left for a separate issue rather than widened into this PR; flagging it here because it is the same silent-entry-loss class this PR is about.

## Verification

- `pnpm check` — clean on the final head `ed8527d8` (type-check, oxlint, oxfmt)
- `pnpm build` + `pnpm test:native` — 243 passed
- `node --expose-gc vitest run` (transaction-log, backup-transaction-logs, destroy, transaction-log-crash-recovery) — 134 passed, 0 failed
- Full `pnpm test` on the rebased head (`96eceb38`, before the README-only follow-up) — 1028 passed / 10 skipped / 0 failed (70 files, 1 skipped)
- Merge integrity: `origin/main` is an ancestor of `HEAD`; `git merge-tree --write-tree origin/main HEAD` equals `HEAD^{tree}`, so this branch's diff against `main` is unchanged apart from the `AGENTS.md` invariant renumbering.
- Independent review: 2 rounds this maintenance pass (round 34 full, round 35 delta), codex + gemini + cursor-grok + harper-domain, `independent=true`, receipt at `ed8527d8`. One review-found defect fixed (the README POSIX/Windows overgeneralization, item 8 above); items 5–6 above reconfirmed open and deliberately left as-is.

Refs HarperFast/harper#2337

Complexity: moderate

<sub>Generated by GPT-5 Codex; maintained by Claude Sonnet 5</sub>

<details>
<summary>Origin — the dispatch brief this PR was written from</summary>

**Release a purged transaction-log segment's mapping instead of refusing to read it**

Maintain https://github.com/HarperFast/rocksdb-js/pull/820 (Release a purged transaction-log segment's mapping instead of refusing to read it) on branch fix/txnlog-purge-read-coherence. Read the current PR head, mergeability, and latest check runs before acting; the dispatch observation may be stale. First read the remote head for refs/heads/fix/txnlog-purge-read-coherence without updating local refs and require it to equal this task's observed head 5396e001119bb8e0520c7a8fcfddcc60d9c35525; if it differs, stop with needs-input because remote history changed after dispatch. Then fetch only refs/heads/main from origin into refs/remotes/origin/main. If Git reports a paused rebase for this task's exact generation branch, require that rebase's recorded original head to equal 5396e001119bb8e0520c7a8fcfddcc60d9c35525 and its recorded onto commit to equal the fetched origin/main tip; stop with needs-input on either mismatch. Only then resume it without repeating the local-HEAD ancestry check or starting another rebase. Otherwise record the remote SHA, verify it is an ancestor of local HEAD, and rebase onto the fetched base. If Context carries companion instructions, follow them exactly for every named gitlink; they override ordinary conflict handling. Preserve both sides' intent for every other conflict. Stop with needs-input on semantic conflicts. Force-with-lease is authorized only for this rebase. Immediately before pushing, re-read the PR's base ref and body plus every companion PR head/state named in Context. The base must still be main, and the declarations and submodule bindings must still match Context. If an open companion advanced, update its named gitlink to the new exact head and rerun relevant tests. If it merged, point the named gitlink at the companion repository's current default-branch tip after verifying that tip contains the merge, then rerun relevant tests. If it closed without merging, stop with needs-input and do not publish its abandoned head. If the base, declaration, binding or any other companion state changed ambiguously, stop with needs-input. Then push the rebased head with `git push --force-with-lease=refs/heads/fix/txnlog-purge-read-coherence:EXPECTED_HEAD_SHA origin HEAD:refs/heads/fix/txnlog-purge-read-coherence`, substituting the recorded SHA. If the remote head changes, stop with needs-input; never overwrite intervening remote work. Do this before evaluating CI. Then inspect CI for the resulting current PR head, not old-head failures. Wait for relevant pending checks, diagnose remaining failures, fix them, run relevant tests, and push in THIS task. Do not create a separate CI-fix task. Stop with needs-input for judgment calls or an unavailable CI result; report exactly what was verified. No merge is authorized.

</details>

<sub>Dispatch: task [`pr-maint-68bd037170be49b2a01e194607ecae40`](https://dev.engineering.harperfabric.com/dispatch/web/#/task/pr-maint-68bd037170be49b2a01e194607ecae40) · queued by automation · ran by claude/sonnet/high · worker kzyp-xps-1</sub>

<sub>Review-Coverage: authored=claude; ran=gemini,codex; adjudicated=domain; declined=cursor-grok,cursor-composer; rounds=35; full=2 @ ed8527d8cacd</sub>

<sub>Human-Review-Need: 4 (decisions: mapping-carried-extent, refused-unlink-policy, destroy-refusal-contract, gc-driven-release, txnstate-reopen-per-flush, windows-strong-frozen-maps) @ ed8527d8cacd</sub>


test/transaction-log.test.ts
```diff
@@ -2476,6 +2476,163 @@ describe('Transaction Log', () => {
...
+							await delay(20);
+						}
+						expect(deletedMappings()).toEqual([]);
+					})
+			);
+
+			// Windows cannot delete a mapped file at all, so there the reader releasing its
+			// mapping is what lets retention advance. Unverified locally (no Windows host); it
+			// runs only in the Windows CI job.
+			it.skipIf(process.platform !== 'win32' || !globalThis.gc)(
```
cb1kenobi: **Windows Node 22 CI failure on this head — Medium**

The `Test on Node.js 22 and windows-latest` job failed for 9398f6b, and it ran ~40 minutes versus ~15 for the Node 24/26 Windows jobs, which looks more like a hang than a clean assertion failure. This test is the PR's only Windows-specific, GC-timing-dependent addition: it needs V8 to collect the external mmap-backed buffer within a 10s gc/purge loop, and external-buffer GC promptness differs across Node/V8 versions — the PR description itself flags this as the spot CI can't be trusted on.

Failure mode: a red check blocks this merge, and if the convergence loop is timing-sensitive on some V8 versions it will keep flaking on future PRs.

Simplest fix: pull the failing job's log to confirm which test failed. If it's this one, make the deadline tolerant (soft-pass or skip with a logged warning when the buffer isn't collected in time, since eventual convergence can't be forced deterministically) or gate it to a single Node version; if the log shows an unrelated pre-existing flake, a rerun should confirm that before merge.

—
Reviewed 9398f6be

kriszyp: Confirmed from the failed job log: `should converge on a purge refused by a live mapping` passed in 724 ms. The two failures were unrelated pre-existing stress/race cases: `txn-close-commit-uaf.test.ts` timed out at 60s, and the `user-shared-buffer.test.ts` schema-sync child exited with Windows status `0xC0000409`. The current head’s Windows/Node 22 job then passed in 18m38s: https://github.com/HarperFast/rocksdb-js/actions/runs/33841739409/job/100936925925. No relaxation or version gate is needed for the convergence test.

src/transaction-log-reader.ts
```diff
@@ -421,6 +438,77 @@ Object.defineProperty(TransactionLog.prototype, 'query', {
...
+	const nextLogId = fromLogId + 1;
+	const logBuffer = getLogMemoryMap(transactionLog, nextLogId);
+	if (logBuffer) {
+		return logBuffer;
+	}
+	if (transactionLog.getLogFileSize(nextLogId) > 0) {
+		return;
+	}
+	FLOAT_TO_UINT32[0] = transactionLog._findPosition(0);
+	const oldestLogId = UINT32_FROM_FLOAT[1];
```
cb1kenobi: **High: purged-run skip drops a surviving segment**

When the next segment cannot be mapped and `getLogFileSize` is 0, `nextReadableLogBuffer` calls `_findPosition(0)` and maps that id. That native walk starts at the current segment and stops at the first gap, so it names the bottom of the contiguous run ending at current, not the next registered survivor. With `{1 gone, 2 gone, 3 kept, 4 gone, 5 current}`, a reader finishing 1 jumps to 5 and never yields segment 3 — silent history loss. This shape comes from `purgeLogs({all})` continuing after a failed unlink (Windows live mapping) or from out-of-band deletes. Before this change the iterator stalled at the hole instead of skipping a survivor.

Suggested fix: return the next id present in `sequenceFiles` after `fromLogId` (native successor), or refuse the jump unless every id in `[nextLogId, oldestLogId)` is gone.

—
Reviewed 20d25cf2

src/transaction-log-reader.ts
```diff
@@ -421,6 +438,77 @@ Object.defineProperty(TransactionLog.prototype, 'query', {
...
+	transactionLog: TransactionLog,
+	fromLogId: number,
+	latestLogId: number
+): LogBuffer | undefined {
+	const nextLogId = fromLogId + 1;
+	const logBuffer = getLogMemoryMap(transactionLog, nextLogId);
+	if (logBuffer) {
+		return logBuffer;
+	}
+	if (transactionLog.getLogFileSize(nextLogId) > 0) {
```
cb1kenobi: **Read probing recreates a deleted segment — High**

Failure mode: when mapping the registered next segment fails because its pathname was removed, `getLogFileSize()` opens that segment through the native file path, whose open mode includes creation. Merely advancing a reader can therefore recreate a zero-byte `.txnlog`; discovery may register that fabricated segment after restart, corrupting iteration and retention behavior.

Simplest fix: expose a non-opening native metadata/successor lookup over `sequenceFiles` and use it here instead of `getLogFileSize()`.

—
Reviewed 112c6cf2

src/transaction-log-reader.ts
```diff
@@ -421,6 +438,77 @@ Object.defineProperty(TransactionLog.prototype, 'query', {
...
+ * The bound for reading `logBuffer`. Normally the segment's written extent, from the store. Once
+ * retention has purged the segment the store reports nothing for it, and the mapping itself is
+ * the only remaining description of the file: the unlink removed one link to an inode a retired
+ * segment never changes again, and this mapping is the other, so its bytes are still exactly the
+ * committed history. Taking the store's 0 as the bound would silently drop every entry the reader
+ * had not reached before the purge — including entries appended after it last polled, which the
+ * writer's overlay made visible in this very mapping.
+ */
+function readableExtent(transactionLog: TransactionLog, logBuffer: LogBuffer): number {
+	const writtenExtent = transactionLog.getLogFileSize(logBuffer.logId);
```
cb1kenobi: **The mapping is not a committed readable extent — High**

Failure mode: after purge discards the native logical extent, `endOfEntries()` infers it from mapped bytes. A retired segment containing bytes from a partial failed append can present a structurally valid frame beyond its logical boundary, causing the reader to yield an entry that was never committed; broken framing also expands the bound to the entire mapping and permits resynchronization through unwritten overlay space.

Simplest fix: retain the immutable logical extent with the retired mapping in native metadata and use that exact bound for iteration and resynchronization; never derive authority from mapping length or frame contents.

—
Reviewed 112c6cf2

src/binding/transaction_log/transaction_log_store.cpp
```diff
@@ -1141,24 +1183,98 @@ void TransactionLogStore::databaseFlushed(rocksdb::SequenceNumber rocksSequenceN
...
+		// stream, and doClose() sets isClosing before taking flushedStateMutex, so
+		// a reopen that sees it clear is ordered before that removal. (doPurge's
+		// own remove_all of an emptied directory runs earlier, with isClosing still
+		// clear; a directory recreated in that window is removed again by the
+		// destroy, and a destroy that fails to close leaves a live store that
+		// legitimately wants its directory.)
+		if (this->isClosing.load(std::memory_order_relaxed)) {
+			return;
+		}
+		try {
```
cb1kenobi: **Flush recovery can resurrect stale retention state — High**

Failure mode: `purgeLogs({ all: true })` can remove the store directory while leaving `isClosing` false, after which a flush recreates `txn.state` with the pre-purge sequence position. A later restart begins segment numbering again at 1 but loads that stale higher position, allowing retention to classify newly written, unflushed segments as purgeable and delete them.

Simplest fix: serialize purge-all with flushed-state recreation and invalidate/reset the recorded position when all segments are removed; only recreate `txn.state` for a position belonging to the current post-purge segment generation.

—
Reviewed 112c6cf2

src/transaction-log-reader.ts
kriszyp: Confirmed and fixed. `MemoryMap` now carries an atomic append-owned `readableExtent`, published only from the validated logical file size and exposed through a type-tagged N-API buffer getter. The JS reader requires that exact extent for iteration and resynchronization; it no longer derives authority from mapped capacity or frame contents. The regression writes a frame-shaped failed-append tail and proves it is not yielded after purge.

src/binding/transaction_log/transaction_log_store.cpp
kriszyp: Confirmed and fixed. An all-segment purge now increments `flushedStateGeneration`, clears the old commit-correlation ring, closes/resets state, and resets the last-written position under the existing lock order. A flush callback carries the observed generation and refuses a stale write. `txn.state` is also reopened and read-back verified by pathname on every update, covering unlink/replacement of the state file. Native and integration regressions cover stale selection, destructive purge, unlink, and pathname replacement.

src/transaction-log-reader.ts
kriszyp: Confirmed and fixed. All read-side extent/mapping probes now call `TransactionLogFile::openExisting()`: POSIX omits `O_CREAT`, Windows uses `OPEN_EXISTING`, and neither path creates a missing parent. The native no-create test covers a missing parent/path, while the integration regression verifies `getLogFileSize()`, aggregate sizing, and timestamp positioning do not recreate an unlinked registered segment.

src/transaction-log-reader.ts
kriszyp: Confirmed and fixed. `nextReadableLogBuffer()` now advances with `_nextLogId()`, backed by `sequenceFiles.upper_bound`, and loops across absent registered successors without using `_findPosition(0)`. The regression exercises the 1/3/5 survivor layout plus a registered-but-absent segment and asserts the end-to-end result `[1, 5]`, so segment 3 cannot be silently skipped.

cb1kenobi (commented):
The new commit only renumbers AGENTS.md invariants after the rebase onto main. Prior blocking issues remain fixed: readers walk registered successors instead of jumping past a survivor, read probes cannot recreate a deleted segment, purged reads use the mapping-carried append-owned extent, and destroy invalidates stale flush correlations. No new blocking defect showed up on the changed lines.

—
Reviewed 162cf812

cb1kenobi (commented):
The new commit only corrects three invariant references in comments after the rebase. It introduces no blocking defect on changed lines.

—
Reviewed 8216e2a0

cb1kenobi (commented):
The merge of main onto this head does not reintroduce the fixed purge bugs or add a new one. Readers still walk registered successors, read probes still cannot recreate a deleted segment, and flushed-state writes still drop stale pre-purge correlations. No blocking defect showed up on the changed lines.

—
Reviewed eee417f4

Issue #821, opened 2026-09-03
WBM stall safeguard (#755) is undone on reopen: TransactionDB::Open re-derives max_write_buffer_size_to_maintain=0 to 256 MiB per CF, permanent stall on Harper prod


Devin-Holland:
## Summary

#755 drops `max_write_buffer_size_to_maintain` to `0` under a stalling WriteBufferManager so retained memtable history cannot fill the budget. RocksDB undoes that for every column family passed to `TransactionDB::Open`: `TransactionDB::PrepareWrap` rewrites `max_write_buffer_size_to_maintain == 0` to `-1`, which derives `max_write_buffer_number × write_buffer_size` = **256 MiB per CF** (`utilities/transactions/pessimistic_transaction_db.cc:296-300` in v11.8.1; `db_descriptor.cpp:1326` opens through `TransactionDB::Open`). Only CFs created *after* open keep the 0, which is why `test/write-buffer-manager-stall.test.ts` passes on a fresh DB. Every process restart reopens all CFs, so in production the safeguard is never in effect, and the permanent stall #755 was written to prevent is what took Harper's Central Manager down repeatedly on 2.8.0 (HarperFast/harper#2490, HarperFast/harper#2450).

## Production evidence (Harper 5.2.8 = rocksdb-js 2.8.0, RocksDB 11.8.1; 28 CFs across 3 DBs; WBM budget 661 MiB, `allowStall: true`)

Every 5.2.8 run logs `Options.max_write_buffer_size_to_maintain: 268435456` for **all** CFs. Memtable memory charged to the cache climbs with write volume and stops exactly at the budget; nothing can bring it back down because the excess is retained history (trimmed only *down to* the 256 MiB target, `memtable_list.cc TrimHistory`), the WBM flush trigger looks only at mutable memory (`write_buffer_manager.h:101-113`), and the stall ends only when total memory drops below the budget (`write_buffer_manager.cc:141-146`):

```
2026/09/02-17:56:31 WriteBuffer(1,    256.00 KB)
2026/09/02-19:56:31 WriteBuffer(827,  206.75 MB)
2026/09/02-22:16:31 WriteBuffer(1636, 409.00 MB)
2026/09/03-00:26:32 WriteBuffer(2236, 559.00 MB)
2026/09/03-04:56:32 WriteBuffer(2645, 661.25 MB)   <- budget reached
[...]

Issue #781, opened 2026-08-14
WriteBufferManager stall: runtime allowStall mutation + late-CF size-0 bypass the open-time history clamp


cb1kenobi:
## Summary

A Codex review on [#758](https://github.com/HarperFast/rocksdb-js/pull/758) (raised by @kriszyp) surfaced two concerns in the **write-buffer-manager / retained-history** feature (AGENTS.md invariant 10 — "Retained memtable history must fit the WriteBufferManager budget"). They are unrelated to that PR's background-error change (it only inherits these files via a `main` merge), so filing them here to track separately.

## 1. Runtime `writeBufferManagerAllowStall` mutation bypasses the open-time history clamp

`src/binding/database/db_settings.cpp` (~L209)

`buildColumnFamilyOptions` resolves the derived (`-1`) `max_write_buffer_size_to_maintain` to `0` **only when a stalling manager is configured at open time** (`resolveMaxWriteBufferSizeToMaintain`). A database can attach a WriteBufferManager while `allowStall` is `false`, so its column families keep RocksDB's derived (large) history target. Flipping `writeBufferManagerAllowStall` to `true` at runtime then enables stalling on that same manager **without revisiting those already-resolved CF targets**. With a budget below the retained history, every write to that database can stall permanently — the exact failure invariant 10 exists to prevent.

Related size-`0` mismatch for late-created families: an existing DB retains the manager, while `resolveMaxWriteBufferSizeToMaintain()` consults the *global* size and may re-derive a large history for a CF created after the global size was reset to `0`.

**Suggested direction:** enforce the invariant against the manager actually attached to each descriptor — either make `allowStall` immutable after attachment, or safely update/reject incompatible CF `max_write_buffer_size_to_maintain` targets before enabling stall.

**Coverage gap:** add tests for (a) enabling stall *after* open, and (b) creating a CF after the global size is reset to `0`.

## 2. Stall regression test can hang instead of failing

`test/write-buffer-manager-stall.test.ts` (~L41)

`db.put()` is declared `async` but calls `store.putSync()` before returning its promise. If RocksDB stalls inside that native call, the JS worker thread is blocked and Vitest's 60s timeout can't fire — the regression hangs until the outer CI job timeout rather than failing this test.

**Suggested direction:** run the stall scenario in a worker/child process the parent can terminate on a deadline, or drive it through a genuinely asynchronous native operation so the runner can still enforce its timeout.

---

[...]

PR #823, opened 2026-09-03
Keep the WriteBufferManager stall safeguard across a reopen

kriszyp:
The stall safeguard added in #755 has never survived a restart. `resolveMaxWriteBufferSizeToMaintain` dropped the derived retained-history target to `0`, and `0` is the one value RocksDB's transaction wrappers reinterpret: `OptimisticTransactionDB::Open` and `TransactionDB::PrepareWrap` each rewrite it to `-1`, which `SanitizeOptions` expands to `max_write_buffer_number * write_buffer_size` — 256MB per column family with this codebase's defaults. Only families created *after* an open, through `DB::CreateColumnFamily`, kept the `0`. So the existing fresh-database test passed while every production restart reopened all families at 256MB apiece and wedged (#821, HarperFast/harper#2490).

The fix resolves the derived target to `1`, the smallest value that survives that rewrite.

## What changed

- **The option contract is aligned across the [public README](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R98), [Store API](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-2717c7fb31c2070dd96f07394997809bc9be7f3fff03ab638db696cc480f39b6R520), [native option](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-db043ff75bbaae52decc158b2c4d47f1acfcad643506c52a02b26554a31d4581R53), [column-family builder declaration](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R43), and [maintainer invariant](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R429).** Each surface now says that an attached manager resolves the derived target to `1`, explains why `0` is unsafe, and preserves explicit positive targets.
- **Regression coverage follows both attachment paths.** The [late-column-family fixture](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-5536307e0386bb01262ce4253f3f53d8c169d1733cbc0c3c160aef5b665e695fR4) proves that clearing the global budget does not change the manager an open database still holds, while the [cache-accounting tests](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-d975ebf1f514b9e06c2b259d34746a6d61f4cf1ecfe4799d51916d9fc8ff14aeR374) retain their original intent by requesting an explicit history window.
- **[The resolver](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR91)** returns a [named `1`](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR73) instead of `0`, and an explicit `0` now takes the same path rather than being forwarded to the wrappers — passing `0` through hands the caller who asked for the least history the most. An explicit *positive* value is still honored untouched.
- **[The clamp keys off a WriteBufferManager being attached](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR101), not off it being a stalling one.** `writeBufferManagerAllowStall` is mutable at runtime — `DBSettings::Config` propagates it through `SetAllowStall` — while `max_write_buffer_size_to_maintain` is an immutable `ColumnFamilyOptions` field fixed when a column family is created. Gating the clamp on it left every family opened before the switch permanently at 256MB, with nothing able to lower it. A non-stalling manager needs the bound anyway: without `allowStall` the same history is simply never reclaimed, which is the ~3GB of retained history #821 measured on the affected deployment's hot families *while it was running with the stall disabled as a workaround*.
- **`1` is not "no history", and the docs now say so.** `MemTableListVersion::MemtableLimitExceeded` compares against `MemoryAllocatedBytesExcludingLast()`, which subtracts the entry `TrimHistory` is about to pop, so with one entry the comparison is `0 >= 1` and the newest flushed memtable is retained. It is dropped on that family's next write: `CheckMemtableFull` schedules a trim whenever `mem()->MemoryAllocatedBytes() + imm()->MemoryAllocatedBytesExcludingLast() >= target`, which any non-empty mutable memtable satisfies at `1`, and `DBImpl::PreprocessWrite` drains that scheduler *before* its `ShouldStall()` check. The bound is one flushed memtable per family instead of 256MB per family, and no positive target does better — the residual is independent of the target's magnitude.
- **[An over-budget report](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR116)** at open, on the process-global `log.warn` channel when a listener is registered, when the families known at that open already reach the budget. It covers both consequences — memory the manager never reclaims, and a permanent stall under `allowStall`. It [compares by division](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR142) rather than multiplying, because the caller's target reaches native as an unvalidated `int64`, and uses `>=` because RocksDB stalls at `memory_usage() >= buffer_size()`. Its wording states that only the families present at that open are counted, so it is a best-effort report and never a retained record or proof of safety. An actual continuous stall is still reported to `stderr` by the stall watchdog when stalls are enabled, unless `ROCKSDB_JS_WBM_STALL_WARN_MS=0` disables it.
- **The attachment it reads is the database's own, not the global setting.** `writeBufferManagerSize` is mutable too, and dropping it to `0` is documented as "no new attachments" rather than a teardown, so a database keeps whichever manager it was opened with. `buildColumnFamilyOptions` therefore takes the attachment as a parameter: the open path passes the manager it just attached, and the late-column-family path in `DBRegistry` [now passes `entry.descriptor->attachedWriteBufferManager`](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R546) — the field the descriptor already exposes for exactly this test elsewhere in the same file. Reading the global instead handed a family created after such a reconfiguration the unclamped 256MB while its history was still charged to the manager the database was holding — the same permanent stall by another route. A review round found this; the regression fails without the fix with `late: 268435456` against `late: 1`. The over-budget report takes its budget from the same attached manager's `buffer_size()`; at open those two sources agree, so this is about reading the authoritative value rather than a bug that was reachable single-threaded.
- **A second, opposite-direction regression in the same read surfaced during this rebase.** The prior fix above passed `db->GetDBOptions().write_buffer_manager != nullptr` as the attachment signal. RocksDB's `SanitizeOptions` unconditionally installs a disabled `WriteBufferManager(0)` when no manager is configured at all, so that pointer is **never null post-open** — the predicate always read "attached", clamping every late-created family to `1` byte of history even in a plain deployment that never called `RocksDatabase.config({ writeBufferManagerSize: ... })`. Fixed by reading `entry.descriptor->attachedWriteBufferManager` (a `const` pointer set once, at open, from the manager rocksdb-js actually attached) instead of RocksDB's sanitized copy — the same field the paragraph above already relies on for the open path and the size inventory (`db_registry.cpp:315`). A `--mode plan` framing recheck (triggered by the review tool's round-count escalation) confirmed this as the sound fix with no better alternative on any axis, and a new child-process regression ([`fork-wbm-unconfigured-late-cf.mts`](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-b5e734c3250cd7aac487e7f3f3db67dbe575615f1be960bcb1d12e4e768e21f6R6)) fails with `late: 1` against the old predicate and passes at `late: 268435456` with the fix.

`src/binding/database/db_settings.cpp` is untouched in the final diff. An earlier revision warned when `allowStall` was switched on at runtime; keying the clamp on the manager's presence makes that transition a non-event, so the warning went with it.

## For the human reviewer

The latest review follow-up removes the duplicate write to RocksDB's private `info_log`; this project does not expose that log as an API. The supported process-global `log.warn` event remains unchanged, and invariant 10 now describes its listener-only delivery and the independently configurable stall watchdog fallback.

**The framing changed mid-flight, on your ruling, and the planning verdict never cleared.** `Framing-Verdict: better-alternative-exists`. The planning gate accepted the value and the layer but named the `allowStall` gate as a missing approach, on the mutability fact above. That was put to you as the one open question and you chose to drop the gate, so the clamp now applies to any attached manager. The two blockers the same gate raised are also fixed: explicit `0` was still taking the broken path, and the runtime transition defeated an open-time safeguard.

**This changes behavior for the configuration Harper runs today.** With `writeBufferManagerAllowStall: false` as the current workaround, families go from a 256MB conflict-check window to `1` — more `ERR_TRY_AGAIN` retries in the forced-flush regime, in exchange for the memory bound the manager exists to provide. The measurements below show no write-path throughput cost from the target itself, but the retry regime is the part worth watching after rollout.

**`1` is a large improvement, not a proof.** The residual is one flushed memtable per family, so a budget below roughly `familyCount * writeBufferSize` can still wedge — and an idle family's retained memtable cannot be trimmed while every writer is stalled, because a write is what schedules the trim. I measured both sides: 7 families with a 1MB write buffer against an 8MB budget still stalls at target `1`, while 4 families at the shipped 16MB default against a 128MB budget writes 160MB in 660ms. The production shape (28 families, 16MB write buffer, 661MB budget) sits inside that, but the arithmetic does not close for an unbounded family count. The over-budget report therefore cannot fire for the resolved target, and the test asserting the quiet case says in its name that this is the guard's blind spot rather than codifying it as correct. Every review round raised it; the adjudication kept it at major and ruled it belongs in its own issue rather than here. The mechanism is `MemTableList::RemoveMemTable` calling `TrimHistory(to_delete, 0)` while `MemoryAllocatedBytesExcludingLast()` excludes the sole entry, so the candidates are a pinned-RocksDB patch passing the active memtable's usage instead of `0`, or an in-binding post-flush trim. Neither is something this PR can set.

**One pre-existing shape the warning leans on.** An existing database has every column family re-opened with the *first* opener's `maxWriteBufferSizeToMaintain`; unlike compression, per-family explicit values are not restored from the OPTIONS file. That is how this code already behaved and I did not change it, but the over-budget report's arithmetic assumes it. Making per-family targets persist would need the same `LoadLatestOptions` round-trip compression does.

Review points I declined, with the reason:

- Two rounds called the child-process fixtures a blocker because Node "does not natively parse TypeScript" and `process.execArgv` is not propagated. Both were adjudicated factually wrong: native type stripping is unflagged at this repo's `engines` floor, `AGENTS.md` documents `.mts` fixtures running with no flags, and all thirteen existing spawn-fixture tests use the identical form. CI confirms it on Node 22, 24 and 26 plus Bun and Deno. A later round raised the identical claim again against the new fixture added for the attachment-check fix below; declined for the same reason, re-verified directly by invoking the fixture with plain `node` (no loader flags) and observing it run.
- A standing nit across rounds is that the new comments still carry rationale `AGENTS.md` invariant 10 owns. I made a consolidation pass — the resolver docblock is now a pointer plus the three facts the code cannot state, and the duplicated restatements at the header, the `DBRegistry` call site and in the stall test are gone. What remains is deliberate.

- A leg called the child-process fixtures a blocker because Node "does not natively parse TypeScript", so spawning a `.mts` file would fail with `ERR_UNKNOWN_FILE_EXTENSION`. This repo's `engines` floor is `^22.18.0 || >=24.0.0` precisely because that is where native type stripping is unflagged, `AGENTS.md` documents the no-`tsx` rule, and `test/fixtures/fork-park-timeout.mts` already does exactly this. CI confirms it on Node 22, 24 and 26 across Linux, macOS and Windows, plus Bun and Deno.
- An earlier leg flagged the `wbmAlreadyCreated` gate on the runtime-transition warning as a proxy that is only true after the first write. It was true from the first open — the manager is created inside `DBDescriptor::open` — and the adjudication pass reached the same conclusion. That code is gone from the final diff regardless.
- Three rounds raised comment mass on the new docblocks. I kept them: each states a why or a constraint the code cannot express — the wrapper rewrite, the `MemoryAllocatedBytesExcludingLast` semantics, the overflow reason for the division, the immutability that motivates the clamp condition. That is what `AGENTS.md` asks comments to carry, and the surrounding file is written the same way.

## Verification

**The regression was watched go red on the base commit**, most recently with the resolver reverted to `origin/main` and the tests kept:

```
 × keeps the safeguard across a reopen and does not stall (optimistic) 60289ms
   → expected 'SIGKILL' to be null
 × keeps the safeguard across a reopen and does not stall (pessimistic) 60234ms
   → expected 'SIGKILL' to be null
 × clamps under a non-stalling manager too 816ms
   → expected [ 268435456, 268435456, …(2) ] to deeply equal [ 1, 1, 1, 1 ]
 × normalizes an explicitly requested 0 rather than passing it to the wrappers 60249ms
   → expected 'SIGKILL' to be null
 × warns when the known families already reach the budget, and says so 415ms
   → expected [] to have a length of 1 but got +0
 × warns when the total exactly equals the budget 274ms
   → expected [] to have a length of 1 but got +0

 Tests  6 failed | 5 passed (11)
```

`SIGKILL` rather than a red assertion is the point for the stalling cases: `put()` runs `store.putSync()` before returning its promise, so a stalled write blocks the JS thread and no in-process timer or runner timeout can fire (#781). The scenario therefore [runs in a child process the parent kills on a deadline](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-95a5b55c8462612b5a1aacbefc07ff676d7cff00dafa1d892642cec70b5ef8f3R151), so a regression fails bounded instead of hanging the job. The [non-stalling case](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-95a5b55c8462612b5a1aacbefc07ff676d7cff00dafa1d892642cec70b5ef8f3R237) cannot stall by construction, so it fails as a plain value assertion — which is what pins the new clamp condition.

A [fourth case](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-95a5b55c8462612b5a1aacbefc07ff676d7cff00dafa1d892642cec70b5ef8f3R253) covers the attachment path: it opens a database under a manager, clears the global budget, then creates a column family and reads both families' targets out of the `LOG` after the child exits, because RocksDB's info logger is buffered and only flushes the newest block on close.

The [fixture](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-0007e6ae235ee482609fdc40c926a7326d44a2ad1e5827d26bb4bc8cd611ea58R27) runs at the **shipped 16MB `writeBufferSize` default** with a 128MB budget over four families, so the pre-fix target it faces is the real `16 * 16MB` rather than a scaled-down stand-in. Both transaction modes get their own case, because the rewrite is duplicated in each wrapper. The assertion reads the per-family `max_write_buffer_size_to_maintain` back out of the database `LOG`, which RocksDB rewrites on every open, and requires every family to be at `1` — the same observation the issue's repro used.

**Measured, because the #755 benchmark cited in the issue measured the old effective `0` and not `1`.** Interleaved arms, fresh database per rep, 4 families, 1MB write buffer, 32MB written, WriteBufferManager off so every target is viable, median of 5:

| target | organic flushing | flush every 2MB |
|---|---|---|
| 0 (temporary local patch) | 205ms | 582ms |
| 1 | 170ms | 592ms |
| 16MB | 174ms | 593ms |
| 256MB | 161ms | 619ms |

No measurable write-path cost from the positive target in either regime, including the forced-flush case where the extra history bookkeeping was predicted to show up.

**Three existing tests changed meaning, and all were corrected rather than weakened.**

The conflicting-commit case asserted `ERR_TRY_AGAIN`, and with a retained memtable the check can now find the conflict and report `ERR_BUSY`. It writes twice after the flush so the trim is scheduled and then drained, which puts it back on the minimal-history path the test exists to cover. Both codes refuse the commit, so this was never a safety change, but the specific code is what proves the scenario is the intended one.

`write-buffer-manager.test.ts`'s two memory-reclamation cases observe what a *retained window* charges to the block cache, and under a configured manager that window is no longer the default. They now [ask for an explicit 256MB target](https://github.com/HarperFast/rocksdb-js/pull/823/changes?w=1#diff-d975ebf1f514b9e06c2b259d34746a6d61f4cf1ecfe4799d51916d9fc8ff14aeR378), which is what the default used to derive to there, so they keep testing WriteBufferManager charge accounting rather than the new default.

Separately, a latent cross-file hazard surfaced. The `WriteBufferManager` is a native process-global and Vitest's `threads` pool runs every file in one process, so the two WriteBufferManager test files were racing to create the singleton and only worked because of the file order. Growing the stall file changed that order and exposed it: `costToCache` disagreed between them and `config()` refuses to change it after creation. Both files now agree on it, and the stall file resets the manager when it finishes.

**Suites.** Rebased again onto current `origin/main` (252 commits past the prior rebase's base) and re-verified there: full Vitest suite green locally, 915 passed, 9 skipped, 0 failed, including the reopen and attachment-check regressions above. `pnpm build` and `pnpm check` clean.

CI on this rebase's head: 23/24 green (native suite on all three platforms, Bun/Deno on all three, Node 22/24/26 on Linux and macOS, Node 24/26 on Windows). `Test on Node.js 22 and windows-latest` failed once in `test/user-shared-buffer.test.ts`'s unrelated shared-buffer schema-sync repro — a child process exiting `3221226505` (`0xC0000409`, Windows `STATUS_STACK_BUFFER_OVERRUN`) rather than `0` — in code this diff does not touch (no WriteBufferManager, column-family creation, or shared-buffer paths in that repro), matching the same Windows-alternate-runtime flake pattern the prior rebase's CI note already describes for two other jobs. Re-run in progress at the time of writing.

No data-format change and no migration: this only changes an in-memory RocksDB option, so rolling back is mechanically safe and simply restores the permanent-stall defect. It does need rebuilt native prebuilds on every supported platform to take effect.

Refs #821

<sub>Review-Coverage: authored=codex; ran=gemini,claude; adjudicated=domain; declined=cursor-grok,cursor-composer; rounds=2 @ e52f86bd9d4e</sub>

<sub>Human-Review-Need: 3 (decisions: clamp-every-attached-manager, explicit-zero-semantics, unsafe-explicit-target-policy, warning-threshold-model, warning-delivery-surface) @ e52f86bd9d4e</sub>


src/binding/database/db_descriptor.cpp
```diff
@@ -66,43 +70,111 @@ static void applyCompression(
...
+		<< "Only the column families known at this open are counted; families created later add to "
+		<< "the same budget, so a database that does not warn is not proven safe.";
+	const std::string text = msg.str();
+	DEBUG_LOG("DBDescriptor::open WARNING: %s\n", text.c_str());
+	// Also to the database's own LOG: a deployment that registers no JS listener still needs a
+	// durable record of the configuration alongside the options block it applies to.
+	if (infoLog) {
+		rocksdb::Warn(infoLog, "%s", text.c_str());
+	}
+	if (GlobalEvents::hasListeners()) {
```
cb1kenobi: We don't have an API to surface the info log, so skip recording it.
```suggestion
```

src/binding/database/db_descriptor.cpp
```diff
@@ -66,43 +70,111 @@ static void applyCompression(
...
+			: ". ")
+		<< "Only the column families known at this open are counted; families created later add to "
+		<< "the same budget, so a database that does not warn is not proven safe.";
+	const std::string text = msg.str();
+	DEBUG_LOG("DBDescriptor::open WARNING: %s\n", text.c_str());
+	// Also to the database's own LOG: a deployment that registers no JS listener still needs a
+	// durable record of the configuration alongside the options block it applies to.
+	if (infoLog) {
+		rocksdb::Warn(infoLog, "%s", text.c_str());
+	}
```
cb1kenobi: We don't have an API to surface the info log, so no point logging it.
```suggestion
```

cb1kenobi (changes requested):
Just one small nit.

src/binding/database/db_descriptor.cpp
kriszyp: Addressed in `e52f86bd`: removed the unexposed RocksDB `info_log` sink and the `GetDBOptions()` copy it required, while retaining the supported process-global `log.warn` event. Invariant 10 now states that the open-time report is listener-only and separately qualifies the stderr stall-watchdog fallback. Verified with `pnpm build`, the 13-test focused stall suite, the full 916-test Vitest suite, all 194 native tests, and `pnpm check`.

cb1kenobi (approved):
Great!

Issue #822, opened 2026-09-03
WriteBufferManager stall is unobservable and unrecoverable: no counter moves, no log line, only a restart clears it


kriszyp:
## Summary

A WriteBufferManager write stall is invisible to every counter and log surface this library exposes, and nothing ever ends one. On a wedged Harper Central Manager node (5.2.8 = rocksdb-js 2.8.0, 28 CFs, 661 MiB budget, `allowStall: true`) all three of the obvious signals read zero for the entire eight-hour wedge:

| Signal | Value during the stall |
|---|---|
| `rocksdb.stall.micros` | 0 — a WBM stall is not counted there |
| `rocksdb.flush.reason.write_buffer_manager` | 0 |
| `write-buffer-manager-limit-stops` (LOG) | 0 |

Meanwhile every writer thread sat in `WBMStallInterface::Block ← DBImpl::WriteBufferManagerStallWrites ← PreprocessWrite ← WriteImpl ← Put ← rocksdb_js::Database::PutSync`, the flush pool was idle, and reads kept working — so from outside the process the node looks healthy while no write can complete. Diagnosis required gdb on production. That is the part this issue is about; the sizing defect that filled the budget is #821 and the Harper default that enables stalling is HarperFast/harper#2490.

Two things are missing:

**1. No way to see it.** `WriteBufferManager` exposes `memory_usage()`, `mutable_memtable_memory_usage()`, `buffer_size()` and `IsStallActive()`, none of which are surfaced. A stall that has been active for minutes is indistinguishable from an idle database.

**2. No way out of it.** The stall ends only when total memtable memory drops below the budget, and `ShouldFlush()` requires `mutable >= buffer_size / 2` — so when the budget is held by retained history or by memtables already awaiting flush, no flush is triggered, and no stalled writer can switch a memtable to cause one. The only exit is a process restart.

## Requested

1. **Expose the manager's live state.** `memoryUsage`, `mutableMemoryUsage`, `bufferSize`, `stallActive` on the existing stats/settings surface, so a stall is visible in Harper's `system_information` rather than in a debugger.
2. **Watchdog log.** When a stall has been active longer than some threshold (a few seconds), log once at `warn`: budget, memory usage, mutable share, live CF count, and the per-CF `max_write_buffer_size_to_maintain` in effect. Every one of those was needed to explain the production wedge, and every one required reading the RocksDB `LOG` or attaching a debugger.
[...]

PR #824, opened 2026-09-04
Make a WriteBufferManager write stall observable

kriszyp:
A `WriteBufferManager` stall is a second, entirely separate stall mechanism from the one RocksDB reports on. `DBImpl::WriteBufferManagerStallWrites` parks writers on the manager's own queue without touching the `WriteController`, so `rocksdb.stall.micros`, the `WRITE_STALL` histogram and `OnStallConditionsChanged` — and therefore `db.isWriteStalled()` and the `'writeStall'` event — all read `0` for its entire duration. Through an eight-hour production wedge every one of them did, while reads kept working and diagnosis needed gdb.

This makes the condition visible. It is additive: no write-path behavior changes, and nothing about `allowStall`, WriteBufferManager sizing, or the retained-history resolution (#821) is touched.

**1. The manager's live state, on two surfaces.** `db.getStats()` / `db.getStat()` gain five keys under a `writeBufferManager.` prefix — `bufferSize`, `memoryUsage`, `mutableMemoryUsage`, `stallActive`, `stallActiveMs` — added unconditionally, the same treatment the existing `txnlog.*` and `commitPipeline.*` keys get, so they appear with `enableStats: false`. They are routed through [the same prefix branch in `DBHandle::getStat`](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-927a214e5d5cca0234f56ddcab411eac454d4ee098051bde49cd6cd923828bb4R192) those two use, because a single-key read that fell through to the RocksDB statistics path would throw "Statistics are not enabled" during exactly the incident an operator is reaching for it in. The top-level `getWriteBufferManagerStats()` export returns the same values plus the manager's configuration and its live column-family inventory. Both read the manager through an [`std::atomic<WriteBufferManager*>` published at construction](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-44bc4a676f65953959ecb869ef849f6d2924dcc6120e8102870b0dd2fa395ca1R83), so a metrics scrape takes no lock and never materializes the manager as a side effect.

These are process-wide values coming back from a per-database call — the manager is a singleton shared by every database in the process, `worker_threads` included — which is stated in the docs, the JSDoc, and the stat-name reference.

**2. A watchdog that logs one warn line per stall episode**, carrying the budget, usage, mutable share, live column-family count and the [*effective* per-CF `max_write_buffer_size_to_maintain`](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R872) — effective, not requested, because #821's whole finding is that `TransactionDB::Open` rewrites a requested `0` into 256 MiB per column family, so the requested value would hide the fact the line exists to expose. Every one of those was needed to explain the wedge and every one required the RocksDB `LOG` or a debugger.

It owns a thread because every other tick in the process is blocked by the condition it reports: `CommitWorker` parks in `db->Write()` (the `rocksdb-commit` lanes are in #822's backtrace), `logWorker` is event-driven off commits the stall prevents, `ParkTimeoutRegistry`'s thread is per-descriptor and only exists after a VT conflict, RocksDB's stall callbacks never fire, and a JS timer cannot run on a thread parked in `store.putSync()`. One thread per process, [started lazily](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-44bc4a676f65953959ecb869ef849f6d2924dcc6120e8102870b0dd2fa395ca1R87) only while a manager exists *with* `allowStall` (`ShouldStall()` short-circuits otherwise, so no stall is reachable and no thread is started), sampling one relaxed atomic per second. Four lifecycle constraints on that thread are the places to look hardest, each of which the review found the hard way:

- **Lock order is `databasesMutex -> writeBufferManagerMutex -> watchdogMutex`**, because `DBRegistry::OpenDB` holds the first across `DBDescriptor::open`. So the start path [never joins a retiring thread](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-44bc4a676f65953959ecb869ef849f6d2924dcc6120e8102870b0dd2fa395ca1R87) — the thread it would join may be waiting for exactly that lock — and a per-start generation counter retires stale threads instead.
- **The inventory walk [try_locks `databasesMutex`](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R307)** and reports `inventoryAvailable: false` rather than waiting: `PurgeAll` holds that lock across a close whose flush waits out a write stall (AGENTS.md note 16), so blocking there would silence the alarm and hang `getWriteBufferManagerStats()` during the incident both exist to report. A dropped column family keeps charging the manager until its last handle closes, so it moves to a [`weak_ptr` plus a copy of its retention target](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R298) and stays counted until then. Only `expired()` is ever called on that reference — `lock()`ing it would make the sampling thread the potential last releaser, running `~ColumnFamilyDescriptor` under both inventory locks.
- **Stop and join both follow the database flush** — [`shutdown()` snapshots the arm-request generation, closes databases, then retires the watchdog](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-5bdfd0ac49ea126864c1cfbff17a606f1aa339fffd20472611f6fc68082ca199R46), so a shutdown wedged behind a stalled writer can still emit its `stderr` warning. The [`allowStall` falling edge](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-44bc4a676f65953959ecb869ef849f6d2924dcc6120e8102870b0dd2fa395ca1R119) only *parks* the thread, so a report blocked on a full pipe holds no lock any database open needs. Destruction order comes from `DBStats`'s constructor materializing `DBSettings` and `GlobalEvents` first, so reverse static destruction joins the watchdog before anything it reads dies.
- **One joiner owns each retirement.** In [`joinWriteBufferManagerWatchdog`](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-44bc4a676f65953959ecb869ef849f6d2924dcc6120e8102870b0dd2fa395ca1R148), callers wait for an existing owner, return without touching stop state when there is no thread, and decline to retire a replacement when a concurrent reopen advanced the arm-request generation. [The deterministic fixture](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-f2a47b19c8f19999e44f93b9ccecf5922480137a912de052809cbe7df0fab9abR69) delays an older joiner until a database has reopened, then proves the replacement remains monitored; it also runs four rounds of gated concurrent shutdowns.

The decision FSM itself is [Node-free](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-dfe95546fb4a17f54377f5f142145418b9fe0471eb97b92e304e7a6cd881cd54R33) and GoogleTest-covered, so the sampling logic is proved without threads or RocksDB.

## For the human reviewer

The step-6 planning gate ran before any code and cleared: `Framing-Verdict: chosen-approach-sound`. Many pre-push review rounds followed across the original implementation and this rebase (codex + gemini + harper-domain); the decisions below are the ones a reviewer should weigh rather than defects left open.

**A recurring pattern worth knowing about before reading the round history below:** Gemini runs with no session continuity (a fresh model instance each round), so it re-derives the whole diff from scratch every time and has repeatedly raised the same handful of shutdown/teardown "blocker" claims about code paths it apparently doesn't fully trace before writing the finding. Four instances, all checked directly against the code and all wrong:
- **`DBRegistry::instance` torn down under the watchdog** (raised 3 times across rounds 20-26): `instance` is a `std::unique_ptr<DBRegistry>` assigned exactly once, in `Init()` (`db_registry.cpp:348`); grepping the whole file shows no other assignment to it anywhere — `Shutdown()` and `PurgeAll()` both only ever do `if (instance) { ... }` and use the object through it, never reset or release it.
- **`emitGlobalEvent`-vs-`GlobalEvents::Shutdown` use-after-free** (raised twice): `GlobalEvents::Shutdown()` is `releaseAll()` on a function-local-static `EventEmitter` (`global_events.cpp:93-96`) — nothing is freed. `EventEmitter::notify`'s fast path returns on `listenerCount == 0` before touching anything else (`event_emitter.cpp:296-303`), and every listener-mutating path (including `releaseAll`) takes the same mutex `notify` holds across its calls, so a watchdog emit landing after shutdown safely no-ops.
- **Mixed-`allowRearm` join races unloading the module with a thread still running**: the last-env cleanup hook's `join(false)` only runs at `moduleRefCount == 0`, and the env executing the JS `shutdown()` call holds a ref until that call returns — the two joins the claim depends on overlapping cannot overlap. A thread re-armed by the JS path is joined by `~DBStats` at static destruction regardless, which is the backstop for every path.
- **`emitGlobalEvent` from a `std::thread` violates V8 thread-safety** (this round): `EventEmitter::notify` calls `napi_call_threadsafe_function` (`event_emitter.cpp:354` onward) — the standard N-API mechanism for exactly this, not a bare synchronous call into V8. This is also the same pattern the transaction-log worker already uses to reach JS from a non-JS thread.

None of these needed a code change; each is recorded here so a re-read of the raw Gemini lens text in review history doesn't get re-litigated.

Two findings from the last rounds were adjudicated and declined with evidence, both raised more than once: a claimed TOCTOU/use-after-free on `DBRegistry::instance` during shutdown, and a claimed `emitGlobalEvent`-vs-`GlobalEvents::Shutdown` race (see above — these are the same two, from earlier rounds).

**Decisions taken, each reversible:**

- **The warn line goes to `stderr` as well as the `'log.warn'` event.** Harper registers no `log.warn` listener today, so an event-only line reaches nothing without a Harper change. The cost is unstructured output a log pipeline cannot format; route the event and ignore `stderr` if you would rather have it structured. @kriszyp ruled on this, and corrected the rationale I had offered for it: WBM stalls have not been observed freezing Harper's JS threads (no `putSync` on active paths; the observed symptom is every thread complaining about slow commits), so the load-bearing argument for `stderr` is the missing listener, not a blocked event loop. A direct consumer of this library's synchronous write path can still block its own loop, which is why the *sampler* still cannot be a JS timer.
- **The process-wide keys live inside per-database `getStats()`.** That is what makes a stall visible in Harper's `system_information` with no Harper change. A scraper summing them across databases will N-count; the alternative was leaving them only on the static accessor, where nothing scrapes them.
- **The threshold is env-only** (`ROCKSDB_JS_WBM_STALL_WARN_MS`, read once per process), not a `config()` knob like every other WriteBufferManager setting. Adding the knob later is additive.
- **`0` disables the watchdog thread entirely**, and with it `stallActiveMs` — silencing the log also silences that gauge, because only the thread populates it.
- **An out-of-range threshold falls back to 5 s rather than clamping**, erring toward alarming earlier than asked, never later. It now says so on `stderr` instead of doing it silently.
- **The inventory counts every descriptor that attached this manager, read-only included** (WAL recovery can retain charged memtables), plus dropped families still held by a live handle. So `columnFamilies` will not reconcile with a count of open handles. Descriptors that never attached the manager — opened before it was configured, or while its size was 0 — are excluded, and skip the tracking entirely.
- **The end-to-end stall test lives in the default suite** (~9 s, and it hard-fails rather than skips if no stall is reached). It is the only proof the wiring works end to end.
- **`costToCache` in the stats/report surfaces reflects the requested setting, not whether a live cache actually got attached.** Only diverges with the self-contradictory `blockCacheSize: 0` + `writeBufferManagerCostToCache: true` combination — a pre-existing `config()` gap, not something introduced here. The natural fix is `config()` rejecting that combination outright, which is out of scope for an observability PR.

**Accepted residuals, all raised by the review and deliberately not fixed:**

- The end-to-end test does not exercise stall *recovery* or prove at the process level (rather than the Node-free FSM-unit level) that recovery re-arms a second episode or that `disable()` mid-stall zeroes `stallActiveMs`. The recovery FSM is covered deterministically in `test/native/wbm_stall_watchdog_test.cc`; reaching a recovering stall end to end needs a scenario whose retained history can drain, which this one deliberately cannot, and adding another child-process fixture purely for the re-arm/disable path is disproportionate to a nit-level gap.
- **Nothing exercises the inventory's fail-fast `try_lock`s under real contention.** Both could be reverted to blocking locks and every test would still pass. The only holder long enough to observe is `PurgeAll`'s close-with-flush, and `DBRegistry::Shutdown` deliberately closes every descriptor *outside* the lock first, so that window is already collapsed; making it deterministic needs a test-only delay seam inside the `databasesMutex` critical section — a branch on the open path to test a fail-fast property. The concurrent-shutdown fixture does now run four envs through `Shutdown()` while the watchdog samples, which exercises the path without asserting on it.
- **An open that races the module's last-env teardown or true process exit does not arm the watchdog.** `ensure` returns early while the stop latch is held, so a database opened in that narrow window gets no alarm until the *next* open arms it. This is fixed for the explicit, JS-callable `shutdown()` (see "Rebase onto main" below) — only the ambiguous last-env-cleanup-hook / `~DBStats()` path still declines to re-arm, unchanged from the original reasoning: that path cannot tell "shutdown, may reopen" from "process is ending", and a thread spawned there would never be joined again.
- **A refused `std::thread` is not logged**, for the same lock-order reason. It is visible as `watchdogRunning: false` while `enabled` and `allowStall` are true, and the next open retries.
- Recurring comment-narration nits raised across several rounds (`db_stats.cpp`, `binding.cpp`, `db_descriptor.cpp`/`.h`, `db_registry.cpp`/`.h`) were re-checked each time against the WHY-not-WHAT bar; the ones that survive explain a non-obvious lock-order or lifecycle constraint the code cannot say on its own and are kept. A handful that had genuinely become redundant with a newer comment nearby (introduced by this rebase's own fixes) were trimmed.

**Not implemented, by instruction:** issue item 3 (`SetAllowStall(false)` after N seconds to release queued writers). The hook is now obvious — the watchdog is the only thing in the process that knows how long a stall has lasted, so a bounded escape hatch is a second threshold in the same FSM — but it converts a hang into unbounded memtable growth, which is a maintainer's call.

**A finding the task anticipated that did not hold.** The context expected that with `atomic_flush` (which this library always sets) a WBM-pressure flush is recorded under `ATOMIC_FLUSH_REQUEST_REASON_WRITE_BUFFER_MANAGER` *instead of* `FLUSH_REASON_WRITE_BUFFER_MANAGER`, and asked for a `docs/stats.md` correction if so. Measured here, both counters move: 13/13 with one column family, roughly 1:2 with two to six, and requests can outnumber executed flushes when several databases share a manager. So `rocksdb.flush.reason.write_buffer_manager` is not the wrong counter, and the production `0` was genuine — no manager-pressure flush was ever *requested*, because the trigger looks only at mutable memory. The docs carry the measured relationship rather than the anticipated correction; writing the anticipated note would have shipped a false statement.

## Rebase onto main (2026-09-08)

Rebased onto `origin/main` (47 commits, including #814 secondary-open and #829 `db.use()`). One merge conflict, in `DBDescriptor`'s constructor and its call in `open()`: main added `identityPath`/`secondaryLockToken` for secondary-open support, this branch added `attachedWriteBufferManager`. Resolved by taking both — the header had already merged cleanly with both parameters, so the fix was ordering the constructor call and the `DBDescriptor(...)` definition to match it. Also renumbered a duplicate `AGENTS.md` invariant 19 (main and this branch each independently appended a new one).

Five independent review rounds (codex + gemini) immediately after the rebase surfaced and fixed real issues in the watchdog lifecycle, none touching write-path behavior:

- **Major (gemini): a `db.open()`/`config()` racing `shutdown()` permanently loses its stall watchdog.** `DBRegistry::Shutdown()` releases `databasesMutex` before closing descriptors (deliberately, per the lock-order note above), so a concurrent open during that window is real, not hypothetical. `ensureWriteBufferManagerWatchdog()` bailing on `watchdogStopRequested` was already documented above as an accepted residual — but the fix needed to reconcile with the *reason* it was accepted (see next point), not just remove the gap.
- **Reconciling with the residual above.** The straightforward fix — replay a queued `ensure()` once the stop resolves — re-triggers exactly the "resurrect a thread during final teardown" concern this PR's body already recorded as the reason for *not* doing that. Resolved by making `joinWriteBufferManagerWatchdog()` caller-aware (`allowRearm`): the explicit, JS-callable `shutdown()` — already documented and tested (`shutdown.test.ts`) to support reopening — passes `true`; the module's last-env cleanup hook and `~DBStats()`, which cannot distinguish "may reopen" from "process is ending", pass `false` and drop a queued replay exactly as before this ever tracked one.
- **Two follow-ups from codex's delta review:** an explicit `disableWriteBufferManagerWatchdog()` landing after a queued replay could have the replay silently re-arm a watchdog the caller had just turned off (fixed: disable now cancels a pending replay); and a failing `std::thread::join()` (rare, but `std::system_error` is documented as possible) left the thread object still joinable, so its destructor would call `std::terminate()` even though the flag/wedge state was already fixed by the retirement guard (fixed then, later fully removed — see below).
- **TOCTOU (gemini): a disable racing a live sample could have its `stallActiveMs` reset overwritten by a stale in-flight value.** `sampleWriteBufferManagerStall()` checked-then-stored without holding `watchdogMutex`, while `disableWriteBufferManagerWatchdog()` clears the same field under that lock. Fixed by moving the check-then-store under the lock.

**Declined at that stage, with evidence:** a repeated gemini "blocker" that `std::erase_if(vector)` (pre-existing, not part of this rebase's changes) needs C++20 and this repo might target C++17 — `binding.gyp` pins `-std=c++20` / `/std:c++20` / `CLANG_CXX_LANGUAGE_STANDARD=c++20` on every platform. A repeated codex minor that the join-failure handling was still not fully airtight (a genuinely unreachable trigger, since `join()` is only ever called on a private, non-shared, non-self-joining local `std::thread` here) — later resolved by deleting the handling entirely rather than continuing to defend it (see below). Two gemini minors — a "hot path" string-allocation claim against `getStats()` (an administrative/telemetry surface, not a per-operation path, later fixed anyway — see below) and a claimed unconditional watchdog spawn when `writeBufferManagerSize` is 0 (a plain misread; the code has two size==0 early returns before that call).

### CI repairs, and post-rebase review continued (2026-09-08)

Two unrelated CI regressions surfaced after the rebase. First, pnpm 12's built-in `rebuild` took precedence over the same-named repository script, so automation now uses the unambiguous `pnpm run rebuild`. Then the floating `pnpm/action-setup` input moved again: failing run 34209077677 upgraded its bootstrap pnpm to 12.3.4 and produced a dead Windows shim before repository code ran, while green run 33831487630 had resolved 11.25.0. The repository now [pins pnpm 11.25.0 in `packageManager`](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R97), and all 12 action sites defer to that single pin instead of `latest`.

The same failing matrix exposed a separate macOS-only path spelling regression from the rebase: `/var/...` was correctly canonicalized to `/private/var/...` for transaction-log I/O, but the public `TransactionLog.path` getter leaked that identity. The getter now [returns the store's caller-spelled `displayPath`](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-6a5fb51459adc2a017bf9767a4dc899c0681885c1fb96431af114466605d3448R339), while all I/O stays on the resolved path; [the symlink regression covers both discovered and newly created stores](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-22a3c4e9b693561b0d3d6534ae81880ea0dec05f14fc12b7bbc837a5748e59e5R2246).

Six more review rounds (codex + gemini, forcing harper-domain back in once to adjudicate two Gemini findings that a narrow-delta round had auto-pruned it for) found and fixed:

- **A degraded stall report on inventory contention.** `sampleWriteBufferManagerStall()` marked the episode reported as soon as the alarm line itself was written, even when the retention-histogram inventory collection lost its `try_lock` race (e.g. `DBRegistry::OpenDB` holding `databasesMutex` for WAL recovery, which the review cited at "hundreds of milliseconds"). Since this is the one report a stall episode ever gets, that left the exact diagnostic #821 exists for permanently missing whenever the two windows overlapped. Now retries the inventory collection up to 10 times, 50ms apart (~450ms worst case) before giving up and reporting degraded — still non-blocking (`try_lock` each attempt), bounded, and cheap since it runs once per episode, not per sample.
- **Dead, and actively wrong if reached, join-failure handling.** The `catch (std::system_error) + detach()` guard around `toJoin.join()` (added in an earlier round) turned out to be unreachable — `toJoin` is a local moved out under `watchdogMutex`, so exactly one caller ever holds it, and neither the JS thread nor the static destructor is the watchdog thread itself, so `join()` can neither self-join nor race a concurrent join — and domain adjudication noted that if the branch were somehow reached, the compensating detach is actively wrong (`RetireGuard` would resurrect the detached, still-parked thread). Deleted.
- **A string-keyed round-trip on every `db.getStats()` call.** `setWriteBufferManagerStatsOnObject()` looked up five known keys through the generic string-keyed `lookupWriteBufferManagerStat()`, allocating temporaries and running compare chains for values already known at the call site. Now assigns the five values directly; the generic lookup still backs the single-name `getWriteBufferManagerStat(name)` path, which legitimately needs it.
- **A lock-order violation in the malformed-threshold diagnostic.** `writeBufferManagerStallWarnMs()`'s rare, one-time `fprintf(stderr, ...)` for a malformed `ROCKSDB_JS_WBM_STALL_WARN_MS` ran inside a lazily-initialized function-local static — so its first call, reachable from `ensureWriteBufferManagerWatchdog()` under `databasesMutex -> writeBufferManagerMutex` via `DBDescriptor::open()`, could in principle block those callers on a wedged `stderr` pipe, directly contradicting this PR's own established "no stderr under those locks" rule. Moved the resolution into `DBStats::DBStats()`, which `DBStats::Init()` already materializes once at module load on the JS thread with no locks held (the same pattern already used for the `DBStats` singleton itself); both call sites now read a cached `const` member.
- **Two README/JSDoc mismatches**, in `README.md` and the mirrored `WriteBufferManagerStats` doc comment in `src/load-binding.ts`: "every other value is 0/false when unconfigured" is wrong for `inventoryAvailable` (defaults `true` — nothing to fail to collect) and for `allowStall`/`costToCache` (they reflect the configured setting even before a manager exists, since a manager additionally needs `writeBufferManagerSize`); and "the first five values are also in `db.getStats()`" relied on declaration order matching the five actually-mirrored keys, which it didn't (`enabled` is listed first but isn't one of the five). Both now name the fields explicitly.

The CI-repair review then found and fixed two shutdown-lifecycle gaps: the watchdog had been stopped before database teardown, suppressing the only diagnostic when teardown itself waited on a WBM stall, and a delayed concurrent shutdown could retire a watchdog that a newer open had armed. Retirement now happens after the database flush, is bound to the caller's arm-request generation, and no-thread callers leave shared stop state untouched. The review also replaced an exact cross-surface `stallActiveMs` equality with monotonic assertions because the one-second sampler may tick between separate reads.

The next cross-runtime matrix separated three timing assumptions from product failures. The transaction-log backup warning test now [waits for asynchronous event delivery](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-9f2dd4646e38cd6a757b8ca49223179f96d6981088310017d76a1e77f2cd0bdaR139), the coordinated-retry close regression accepts either rejection or the native `RETRY_NOW` result because both prove the park drained, and the pre-existing Deno/macOS worker-teardown abort in #746 uses the suite's established one-retry policy. Windows then exposed the same identity/display distinction in malformed-segment warning text: backup snapshots now [compose warnings from the store's display path and the segment filename](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-42d2e6ef64957846e9ce8c84051a6fd7e8f6307406deaf7bc1bb3840938b8979R462), while filesystem access remains on the canonical identity path.

Finally, Node 22/macOS exposed that a close-during-backup promise could settle before `AsyncBackupState` released its copied descriptor pin, allowing an immediate `registryStatus()` or subsequent `shutdown()` to observe the closing database. The complete callback now [releases the pin and retries the deferred purge before resolve/reject](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-07add81e0e72aab506258935de0b08132dc52c05c2df8109bef01b257c7bf1cdR372); the destructor remains an idempotent fallback for earlier exits.

**Full local suite green** on the final pushed head: 65 files passed, 1 skipped, 906 tests passed, 9 skipped, 0 failed (`pnpm test`, Node 26 / Linux), plus 194/194 native GoogleTest. The two final production deltas also passed incremental release binding rebuilds, their focused backup/transaction-log/shutdown regressions, and `pnpm check`.

## Verification

- **The concurrent-join regression is proved by reverting the fix**: the fixture hung to its deadline on 1 of 3 runs without it, and passes 4/4 with it.
- **Concurrent reopen is deterministic**: a test seam delays the second shutdown join for five seconds, the database reopens after the first caller returns, and the delayed caller leaves the replacement watchdog running.
- **Stalled teardown stays diagnosable**: [the real-stall shutdown regression](https://github.com/HarperFast/rocksdb-js/pull/824/changes?w=1#diff-2d0db2138b6a78af721ce4096bb065a01879c61b11ac60d0f52a0cbd683b3b43R209) blocks `shutdown()` behind a parked writer and observes the watchdog warning before the parent kills the child.
- **Native GoogleTest**: 194 tests pass (`pnpm test:native`) on the current pushed head, including the watchdog FSM (rising edge, threshold crossing, one-shot, retry-until-acknowledged, recovery re-arm, flapping, threshold `0`), the env-var parse (clamp, reject, out-of-range) and the report formatter.
- **The end-to-end stall test reaches a real stall** and asserts exactly one warn line across several times the threshold, its full payload, the `'log.warn'` event carrying the same line once, and — from the child's main thread while a worker is blocked in `putSync` — `stallActive`, a rising `stallActiveMs`, and agreement between `getWriteBufferManagerStats()`, `getStats()` and `getStat()`.
- **`process.exit()` with a live watchdog exits 0**, verified directly; without the join it is `std::terminate()` by construction.
- **The `allowStall` runtime edges** start and stop the watchdog, verified by test and by hand (`on → off → back on`).
- **Transaction-log display path**: the discovered/created symlink test fails on `origin/main` with the resolved real path and passes on this head with the caller spelling.
- **CI toolchain audit**: local pnpm resolves to 11.25.0; all 12 `pnpm/action-setup` sites use the manifest pin, with no pnpm `version: latest` input remaining. Run 34225580031 passed every Windows Node/native/Bun/Deno job.
- **Cross-runtime stabilization**: run 34225580031 passed every Linux job and every macOS Bun/Deno/Node 24/Node 26 job; its sole failure, Node 22/macOS, was the backup completion-order regression fixed in the final delta.
- **Full Vitest suite**: 65 files passed, 1 skipped, 906 tests passed, 9 skipped, 0 failed (`pnpm test`, Node 26 / Linux) on final HEAD `4e308c5f`.
- **Backup completion regression**: Node 22 ran the full backup plus shutdown suites (37 passed, 1 skipped), then the close-during-backup test passed five additional consecutive runs.
- **Final GitHub CI**: [all 19 Test workflow jobs passed](https://github.com/HarperFast/rocksdb-js/actions/runs/34227730170) on `4e308c5f`, covering Node 22/24/26, Bun, Deno, and native GoogleTest across Linux, macOS, and Windows; [Benchmark](https://github.com/HarperFast/rocksdb-js/actions/runs/34227730173), [Stress](https://github.com/HarperFast/rocksdb-js/actions/runs/34227730180), caller-workflow validation, and both Socket checks also passed.
- `pnpm check` (type-check, lint, format) clean.

Refs #822

https://claude.ai/code/session_01LHT9magoijhyikugJcB3Wu

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=claude,gemini; declined=cursor-grok,cursor-composer,domain; rounds=12 @ 4e308c5f7bf9</sub>

<sub>Human-Review-Need: 3 @ 4e308c5f7bf9</sub>

### DIFF
--- a/src/binding/database/db_settings.h
+++ b/src/binding/database/db_settings.h
@@ -35,6 +35,12 @@ struct WriteBufferManagerStats final {
 	bool watchdogRunning = false;
 	uint64_t columnFamilies = 0;
 	std::map<int64_t, uint64_t> maxWriteBufferSizeToMaintain;
+	/**
+	 * False when the registry lock was held by a close that is itself wedged on
+	 * the stall being diagnosed, so the two fields above are empty rather than
+	 * measured. A diagnostic that blocks on that lock answers nothing at all.
+	 */
+	bool inventoryAvailable = true;
 };
 
 /**
@@ -98,9 +104,14 @@ private:
 	std::condition_variable watchdogCv;
 	bool watchdogStarted = false;
 	bool watchdogStopRequested = false;
+	// Bumped on every start. A thread whose generation is stale exits even if a
+	// new start has already cleared `watchdogStopRequested`: the joiner releases
+	// watchdogMutex before `join()`, so without this the retiring thread can
+	// observe the reset flag and loop forever with its joiner blocked on it.
+	uint64_t watchdogGeneration = 0;
 
 	void ensureWriteBufferManagerWatchdog();
-	void runWriteBufferManagerWatchdog();
+	void runWriteBufferManagerWatchdog(uint64_t generation);
 	void sampleWriteBufferManagerStall(WbmStallWatchdogState& state, uint64_t thresholdMs);
 
 	bool compactOnClose;
@@ -176,6 +187,14 @@ public:
 
 	/** Joins the stall watchdog. Idempotent, and restartable afterwards. */
 	void joinWriteBufferManagerWatchdog();
+
+	/**
+	 * Joins the watchdog if the module's cleanup hook never ran. `process.exit()`
+	 * skips N-API env cleanup, and destroying a joinable `std::thread` calls
+	 * `std::terminate()` — an observability feature must not turn a clean exit
+	 * into SIGABRT. Mirrors `~CommitWorker`.
+	 */
+	~DBSettings();
 
 	inline bool getCompactOnClose() const {
 		return compactOnClose;
--- a/src/binding/database/db_settings.h
+++ b/src/binding/database/db_settings.h
@@ -1,14 +1,41 @@
 #ifndef __DB_SETTINGS_H__
 #define __DB_SETTINGS_H__
 
+#include <atomic>
+#include <condition_variable>
+#include <map>
 #include <memory>
 #include <mutex>
 #include <node_api.h>
+#include <thread>
 #include "rocksdb/cache.h"
 #include "rocksdb/write_buffer_manager.h"
 #include "core/verification_table.h"
+#include "core/wbm_stall_watchdog.h"
 
 namespace rocksdb_js {
+
+/**
+ * Live state of the process-wide `WriteBufferManager`, as reported by
+ * `db.getStats()` / `db.getStat()` and `RocksDatabase.getWriteBufferManagerStats()`.
+ *
+ * The column-family inventory is filled in only when the caller asks for it: it
+ * walks the database registry, while everything above it is a handful of atomic
+ * loads. `getStats()` is a scrape path and takes the cheap half.
+ */
+struct WriteBufferManagerStats final {
+	bool enabled = false;
+	uint64_t bufferSize = 0;
+	uint64_t memoryUsage = 0;
+	uint64_t mutableMemoryUsage = 0;
+	bool allowStall = false;
+	bool costToCache = false;
+	bool stallActive = false;
+	uint64_t stallActiveMs = 0;
+	bool watchdogRunning = false;
+	uint64_t columnFamilies = 0;
+	std::map<int64_t, uint64_t> maxWriteBufferSizeToMaintain;
+};
 
 /**
  * Stores the global settings for RocksDB databases as well as various global
@@ -47,6 +74,34 @@ private:
 
 	std::shared_ptr<rocksdb::WriteBufferManager> writeBufferManager;
 	std::mutex writeBufferManagerMutex;
+
+	// The manager, published with release ordering once it is fully constructed
+	// and never reset — a runtime `writeBufferManagerSize: 0` is a "no new
+	// attachments" signal, not a teardown, so the raw pointer stays valid for the
+	// process lifetime. Read paths take one acquire load instead of
+	// writeBufferManagerMutex: a metrics scrape must not contend with config()
+	// or a database open.
+	std::atomic<rocksdb::WriteBufferManager*> writeBufferManagerPtr{nullptr};
+
+	// Published by the watchdog thread each sample; 0 when no stall is active.
+	std::atomic<uint64_t> writeBufferManagerStallActiveMs{0};
+	std::atomic<bool> writeBufferManagerWatchdogRunning{false};
+
+	// Watchdog lifecycle. The start path runs under databasesMutex ->
+	// writeBufferManagerMutex (DBRegistry::OpenDB holds the former across
+	// DBDescriptor::open), so watchdogMutex is the innermost lock and the
+	// watchdog thread must never hold it while taking either of the other two —
+	// it drops watchdogMutex before every sample, and the sample's inventory walk
+	// is the one place it takes databasesMutex.
+	std::thread watchdogThread;
+	std::mutex watchdogMutex;
+	std::condition_variable watchdogCv;
+	bool watchdogStarted = false;
+	bool watchdogStopRequested = false;
+
+	void ensureWriteBufferManagerWatchdog();
+	void runWriteBufferManagerWatchdog();
+	void sampleWriteBufferManagerStall(WbmStallWatchdogState& state, uint64_t thresholdMs);
 
 	bool compactOnClose;
 
@@ -94,6 +149,34 @@ public:
 
 	std::shared_ptr<rocksdb::WriteBufferManager> getWriteBufferManager();
 
+	/**
+	 * Returns the manager if one has already been created, without creating it.
+	 * Lock-free and safe from any thread — a stats read must never materialize
+	 * the manager as a side effect (the peer of `getVerificationTableRaw()`,
+	 * without its mutex).
+	 */
+	rocksdb::WriteBufferManager* getWriteBufferManagerRaw() const {
+		return this->writeBufferManagerPtr.load(std::memory_order_acquire);
+	}
+
+	/**
+	 * Samples the manager's live state. `includeColumnFamilies` additionally walks
+	 * the database registry for the attached, writable column-family inventory,
+	 * which is O(open column families) — leave it off on scrape paths.
+	 */
+	WriteBufferManagerStats getWriteBufferManagerStats(bool includeColumnFamilies);
+
+	/**
+	 * Signals the stall watchdog to stop without waiting for it. Split from the
+	 * join so teardown can shut databases down (and flush) while the watchdog is
+	 * still winding down: its warn line goes to stderr, which can block on a full
+	 * pipe, and a logging stall must never sit in front of the flush path.
+	 */
+	void requestWriteBufferManagerWatchdogStop();
+
+	/** Joins the stall watchdog. Idempotent, and restartable afterwards. */
+	void joinWriteBufferManagerWatchdog();
+
 	inline bool getCompactOnClose() const {
 		return compactOnClose;
 	}
@@ -116,6 +199,8 @@ public:
 	VerificationTable* getVerificationTableRaw();
 
 	static napi_value Config(napi_env env, napi_callback_info info);
+
+	static napi_value GetWriteBufferManagerStats(napi_env env, napi_callback_info info);
 
 	static void Init(napi_env env, napi_value exports);
 };
--- a/src/binding/database/db_settings.h
+++ b/src/binding/database/db_settings.h
@@ -113,6 +113,7 @@ private:
 	// watchdogMutex before `join()`, so without this the retiring thread can
 	// observe the reset flag and loop forever with its joiner blocked on it.
 	uint64_t watchdogGeneration = 0;
+	bool watchdogAtexitRegistered = false;
 
 	void ensureWriteBufferManagerWatchdog();
 	void runWriteBufferManagerWatchdog(uint64_t generation);
@@ -193,10 +194,11 @@ public:
 	void joinWriteBufferManagerWatchdog();
 
 	/**
-	 * Joins the watchdog if the module's cleanup hook never ran. `process.exit()`
-	 * skips N-API env cleanup, and destroying a joinable `std::thread` calls
-	 * `std::terminate()` — an observability feature must not turn a clean exit
-	 * into SIGABRT. Mirrors `~CommitWorker`.
+	 * Joins the watchdog if neither the module's cleanup hook nor the `atexit`
+	 * handler registered at watchdog start ran. `process.exit()` skips N-API env
+	 * cleanup, and destroying a joinable `std::thread` calls `std::terminate()` —
+	 * an observability feature must not turn a clean exit into SIGABRT. Mirrors
+	 * `~CommitWorker`.
 	 */
 	~DBSettings();
 
--- a/src/binding/database/db_settings.h
+++ b/src/binding/database/db_settings.h
@@ -89,7 +89,6 @@ private:
 	// or a database open.
 	std::atomic<rocksdb::WriteBufferManager*> writeBufferManagerPtr{nullptr};
 
-	// Published by the watchdog thread each sample; 0 when no stall is active.
 	std::atomic<uint64_t> writeBufferManagerStallActiveMs{0};
 	std::atomic<bool> writeBufferManagerWatchdogRunning{false};
 
--- a/src/binding/database/db_settings.h
+++ b/src/binding/database/db_settings.h
@@ -91,6 +91,11 @@ private:
 
 	std::atomic<uint64_t> writeBufferManagerStallActiveMs{0};
 	std::atomic<bool> writeBufferManagerWatchdogRunning{false};
+	// Lock-free mirror of watchdogStopRequested, so the sample path can abandon a
+	// report before writing it. The report's stderr write is unbounded on a full
+	// pipe and a joiner waits behind it; this bounds that wait to a write already
+	// in progress rather than one about to start.
+	std::atomic<bool> writeBufferManagerWatchdogStopping{false};
 
 	// Watchdog lifecycle. The start path runs under databasesMutex ->
 	// writeBufferManagerMutex (DBRegistry::OpenDB holds the former across
--- a/src/binding/database/db_settings.h
+++ b/src/binding/database/db_settings.h
@@ -2,46 +2,14 @@
 #define __DB_SETTINGS_H__
 
 #include <atomic>
-#include <condition_variable>
-#include <map>
 #include <memory>
 #include <mutex>
 #include <node_api.h>
-#include <thread>
 #include "rocksdb/cache.h"
 #include "rocksdb/write_buffer_manager.h"
 #include "core/verification_table.h"
-#include "core/wbm_stall_watchdog.h"
 
 namespace rocksdb_js {
-
-/**
- * Live state of the process-wide `WriteBufferManager`, as reported by
- * `db.getStats()` / `db.getStat()` and `RocksDatabase.getWriteBufferManagerStats()`.
- *
- * The column-family inventory is filled in only when the caller asks for it: it
- * walks the database registry, while everything above it is a handful of atomic
- * loads. `getStats()` is a scrape path and takes the cheap half.
- */
-struct WriteBufferManagerStats final {
-	bool enabled = false;
-	uint64_t bufferSize = 0;
-	uint64_t memoryUsage = 0;
-	uint64_t mutableMemoryUsage = 0;
-	bool allowStall = false;
-	bool costToCache = false;
-	bool stallActive = false;
-	uint64_t stallActiveMs = 0;
-	bool watchdogRunning = false;
-	uint64_t columnFamilies = 0;
-	std::map<int64_t, uint64_t> maxWriteBufferSizeToMaintain;
-	/**
-	 * False when the registry lock was held by a close that is itself wedged on
-	 * the stall being diagnosed, so the two fields above are empty rather than
-	 * measured. A diagnostic that blocks on that lock answers nothing at all.
-	 */
-	bool inventoryAvailable = true;
-};
 
 /**
  * Stores the global settings for RocksDB databases as well as various global
@@ -80,44 +48,6 @@ private:
 
 	std::shared_ptr<rocksdb::WriteBufferManager> writeBufferManager;
 	std::mutex writeBufferManagerMutex;
-
-	// The manager, published with release ordering once it is fully constructed
-	// and never reset — a runtime `writeBufferManagerSize: 0` is a "no new
-	// attachments" signal, not a teardown, so the raw pointer stays valid for the
-	// process lifetime. Read paths take one acquire load instead of
-	// writeBufferManagerMutex: a metrics scrape must not contend with config()
-	// or a database open.
-	std::atomic<rocksdb::WriteBufferManager*> writeBufferManagerPtr{nullptr};
-
-	std::atomic<uint64_t> writeBufferManagerStallActiveMs{0};
-	std::atomic<bool> writeBufferManagerWatchdogRunning{false};
-	// Lock-free mirror of watchdogStopRequested, so the sample path can abandon a
-	// report before writing it. The report's stderr write is unbounded on a full
-	// pipe and a joiner waits behind it; this bounds that wait to a write already
-	// in progress rather than one about to start.
-	std::atomic<bool> writeBufferManagerWatchdogStopping{false};
-
-	// Watchdog lifecycle. The start path runs under databasesMutex ->
-	// writeBufferManagerMutex (DBRegistry::OpenDB holds the former across
-	// DBDescriptor::open), so watchdogMutex is the innermost lock and the
-	// watchdog thread must never hold it while taking either of the other two —
-	// it drops watchdogMutex before every sample, and the sample's inventory walk
-	// is the one place it takes databasesMutex.
-	std::thread watchdogThread;
-	std::mutex watchdogMutex;
-	std::condition_variable watchdogCv;
-	bool watchdogStarted = false;
-	bool watchdogStopRequested = false;
-	// Bumped on every start. A thread whose generation is stale exits even if a
-	// new start has already cleared `watchdogStopRequested`: the joiner releases
-	// watchdogMutex before `join()`, so without this the retiring thread can
-	// observe the reset flag and loop forever with its joiner blocked on it.
-	uint64_t watchdogGeneration = 0;
-	bool watchdogAtexitRegistered = false;
-
-	void ensureWriteBufferManagerWatchdog();
-	void runWriteBufferManagerWatchdog(uint64_t generation);
-	void sampleWriteBufferManagerStall(WbmStallWatchdogState& state, uint64_t thresholdMs);
 
 	bool compactOnClose;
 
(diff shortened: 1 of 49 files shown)


src/database.ts
```diff
@@ -342,6 +344,35 @@ export class RocksDatabase extends DBI<DBITransactional> {
...
+	 *
+	 * @example
+	 * ```typescript
+	 * const wbm = RocksDatabase.getWriteBufferManagerStats();
+	 * if (wbm.stallActive) {
+	 * 	log.warn(`writes stalled for ${wbm.stallActiveMs}ms: ${wbm.memoryUsage}/${wbm.bufferSize}`);
+	 * }
+	 * ```
+	 */
+	static getWriteBufferManagerStats(): WriteBufferManagerStats {
```
cb1kenobi: This exposes a second surface for the same API.

```
import { RocksDatabase, getWriteBufferManagerStats } from '@harperfast/rocksdb-js';
const wbm1 = RocksDatabase.getWriteBufferManagerStats();
const wbm2 = getWriteBufferManagerStats();
```

I think we should remove this static method version and be more like `registryStatus` and only have a top-level export.

src/index.ts
```diff
@@ -26,6 +26,7 @@ export {
 	coolTransactionLogs,
 	currentThreadId,
 	fileLockRelease,
+	getWriteBufferManagerStats,
```
cb1kenobi, line 29: `getWriteBufferManagerStats()` is the only top-level export with the "get" prefix, yet provides similar readonly view as `registryStatus()` does. For consistency, we should either drop the "get" or add the "get" to `registryStatus()`. I'm thinking we do the latter. I kinda of like the "get" and I should probably have done it in the first place. Ticket created: https://github.com/HarperFast/rocksdb-js/issues/840.

src/binding/database/db_settings.h
```diff
@@ -1,15 +1,48 @@
...
 
+/**
+ * Live state of the process-wide `WriteBufferManager`, as reported by
+ * `db.getStats()` / `db.getStat()` and `RocksDatabase.getWriteBufferManagerStats()`.
+ *
+ * The column-family inventory is filled in only when the caller asks for it: it
+ * walks the database registry, while everything above it is a handful of atomic
+ * loads. `getStats()` is a scrape path and takes the cheap half.
+ */
+struct WriteBufferManagerStats final {
```
cb1kenobi: I don't love putting the `WriteBufferManagerStats` logic in `db_settings`. `db_settings` is for incoming database settings, not APIs returning data and wiring up watchdogs. I think all of this write buffer stats stuff should go in a `database/db_stats.*` (or whatever) files with it's own singleton and `Init()`.

cb1kenobi (changes requested):

src/database.ts
kriszyp: Done — the static `RocksDatabase.getWriteBufferManagerStats()` is gone; the top-level export is the only surface, matching `registryStatus`.

— Claude Opus

src/index.ts
kriszyp: Agreed, and I'm leaving `getWriteBufferManagerStats` as-is here so the rename lands as one change: HarperFast/rocksdb-js#840 adds the `get` prefix to `registryStatus`, and doing half of it in this PR would just make that one a partial revert.

— Claude Opus

src/binding/database/db_settings.h
kriszyp: Done — `src/binding/database/db_stats.{h,cpp}` now owns the `WriteBufferManagerStats` snapshot, the `getWriteBufferManagerStats` export (its own `DBStats::Init()`), the stall-watchdog thread and its whole lifecycle. `DBSettings` is back to incoming settings: it publishes the manager pointer when it lazily creates one and calls `ensure`/`disable` on the `allowStall` edges, and that's all it knows about stats.

The singleton is a magic static like `DBSettings`'s, and its constructor materializes `DBSettings` and `GlobalEvents` first so reverse static destruction joins the watchdog before either dependency it reads is destroyed — that ordering also let the old `atexit` handler and the deliberate event-emitter leak go away.

— Claude Opus

cb1kenobi (approved):
Looks good, nice work!

PR #829, opened 2026-09-04
feat(database): add `db.use()` to access multiple column families

### LINKED ISSUE #828: Add `db.use(name)` to access multiple column families from a single database instance
## Summary

Add `RocksDatabase.use(name, options?)` so a single `db` instance can interact with multiple column
families, instead of requiring a separate `new RocksDatabase(path, { name })` per column family.

## Motivation

Today each column family is reached through its own `RocksDatabase`:

```js
const db = new RocksDatabase('/path/to/db');            // default CF
const events = new RocksDatabase('/path/to/db', { name: 'events' }); // events CF
```

Internally these share a process-global, path-keyed `DBDescriptor` (which already opens every column
family), and each `RocksDatabase` pins one column family via its own `DBHandle`. The ergonomics don't
reflect that shared backing: callers manage independent instances and paths by hand.

## Proposal

A factory method — create-and-open sugar, modeled on `useLog`:

```js
const db = RocksDatabase.open('/path/to/db');
const events = db.use('events');   // RocksDatabase bound to the 'events' CF

await events.put('e1', payload);
await db.put('k', 'v');            // default CF, unaffected
```

Semantics:

- Returns a `RocksDatabase` bound to `name`, opening (and creating, if missing) the column family on
  first use. It shares the same underlying database, so a transaction, backup, or checkpoint still
  spans every column family.
- **Get-or-create, weakly cached:** `db.use('events') === db.use('events')` while the view is
  referenced and open; a closed or garbage-collected view is transparently recreated. The cache
  holds only a `WeakRef` (+ a `FinalizationRegistry` to reclaim map entries), so views are never
  pinned for the database's lifetime.
- Views are **independent handles**: closing the parent does not close them (and vice versa); the
  underlying database stays open until every handle is closed or collected.
- Views inherit the options the database was constructed with; `options` overrides them per CF.
- `use()` works regardless of the parent's open state (a discarded parent still yields a valid,
  opened view).

## Scope / non-goals

- No native/C++ change — the `DBDescriptor`/`columns` machinery already supports this; the change is
  TypeScript-only.
- Not in scope (possible follow-ups): a narrower return type than `RocksDatabase` that hides
  whole-database verbs (`backup`/`checkpoint`) from a single-CF view; and cross-column-family atomic
  writes through a single `Transaction`.

cb1kenobi:
## What

Adds `RocksDatabase.use(name, options?)` so a single `db` instance can interact with multiple column
families, instead of constructing a separate `new RocksDatabase(path, { name })` per column family.

Closes #828

## API

```js
const db = RocksDatabase.open('/path/to/db');
const events = db.use('events');   // RocksDatabase bound to the 'events' CF

await events.put('e1', payload);
await db.put('k', 'v');            // default CF, unaffected
```

Create-and-open sugar, modeled on `useLog`. The `DBDescriptor`/`columns` machinery already opens
every column family and pins one per `DBHandle`, so this is **TypeScript-only** — no native change.

## Semantics

- Returns a `RocksDatabase` bound to `name`, opening (and creating, if missing) the CF on first use.
  It shares the same underlying database, so a transaction, backup, or checkpoint still spans every
  column family.
- **Get-or-create, weakly cached:** `db.use('events') === db.use('events')` while the view is
  referenced *and open*; a closed or garbage-collected view is transparently recreated. The cache
  holds only a `WeakRef` (plus a `FinalizationRegistry` that reclaims map entries), so views are
  never pinned for the database's lifetime.
- Views are **independent handles**: closing the parent does not close them (and vice versa); the
  underlying database stays open until every handle is closed or collected.
- The own-column-family name returns `this` (opened); `#name` derives from the store, so
  `Store`-based construction binds the correct CF.
- Works regardless of the parent's open state — a discarded parent still yields a valid, opened view.

## Tests

`test/column-families.test.ts` covers: data isolation, cache identity, `this`-return (open + unopened
parent), `Store`-name derivation, cross-handle data sharing, parent/view close independence,
recreation of an explicitly closed view, per-view transactions, and input validation. GC-collection
is intentionally *not* asserted (unreliable under the runner per AGENTS.md — verified separately that
`use()`-created views are collectible).

## Review

Ran a cross-model review (Codex + Gemini) on the change; all surfaced findings were resolved before
this PR — notably the own-name shortcut now honors create-and-open (returns `this.open()`),
`Store`-based construction derives `#name` from the store, the `WeakRef` map is reclaimed via a
`FinalizationRegistry`, and the docs scope the identity guarantee to referenced-and-open views.

## Follow-ups (not in scope)

- A narrower return type than `RocksDatabase` that hides whole-database verbs (`backup`/`checkpoint`)
  from a single-CF view.
- Cross-column-family atomic writes through a single `Transaction`.

<sub>Review-Coverage: authored=claude; ran=codex; declined=gemini,cursor-grok,cursor-composer,domain; rounds=3 @ 73385e4baee3</sub>

<sub>Human-Review-Need: 4 @ 73385e4baee3</sub>


src/database.ts
```diff
@@ -1225,6 +1253,66 @@ export class RocksDatabase extends DBI<DBITransactional> {
...
+	 * ```
+	 */
+	use(name: string, options?: RocksDatabaseOptions): RocksDatabase {
+		if (typeof name !== 'string' || name === '') {
+			throw new TypeError('Column family name must be a non-empty string');
+		}
+		// This database is already bound to its own column family; return it
+		// (opened, honoring the create-and-open contract). `options` are ignored
+		// here — a bound column family cannot be reconfigured through `use()`.
+		if (name === this.#name) {
```
kriszyp: The `options` contract says these values override the inherited configuration, but this shortcut—and the cached-view branch below—returns an existing handle without validating or applying them. For example, `new RocksDatabase(path).use('default', { readOnly: true })` opens the original writable store, so writes still succeed; a later encoding or compression override for a cached view is likewise silently ignored. Please either remove `options` from `use()` (the simplest API), or compare effective configurations and reject incompatible reuse/create an appropriately configured handle. Please cover both own-name and cached-view calls with differing `readOnly`, encoding, and compression options.

src/database.ts
```diff
@@ -1225,6 +1253,66 @@ export class RocksDatabase extends DBI<DBITransactional> {
...
+			return this.open();
+		}
+
+		const cached = this.#columnFamilies.get(name)?.deref();
+		if (cached !== undefined && cached.isOpen()) {
+			return cached;
+		}
+
+		const columnFamily = new RocksDatabase(this.path, { ...this.#options, ...options, name });
```
kriszyp: This shallow spread shares reference-valued `encoder` and `decoder` instances between the parent and its supposedly independent view. Opening the child then mutates the shared encoder's `name` at `src/database.ts:986`, and clearing either view can reset its shared `structures` at `src/database.ts:161`, changing the other view's codec state and potentially making existing records undecodable. Please move column-family derivation behind a Store-owned factory/fork that creates independent codec state, or require codec factories/per-view instances instead of inheriting mutable codec objects. The retained configuration should also be snapshotted rather than remaining coupled to later mutations of the caller's options object.

src/database.ts
```diff
@@ -1225,6 +1255,71 @@ export class RocksDatabase extends DBI<DBITransactional> {
...
+		if (cached !== undefined && cached.isOpen()) {
+			return cached;
+		}
+
+		// Build the view's store from the parent store's class, not the base
+		// `Store`, so a custom `Store` subclass's overridden behavior carries over.
+		// `mergedOptions` is also passed to the view so a nested `use()` inherits.
+		const mergedOptions = { ...this.#options, ...options, name };
+		const StoreClass = this.store.constructor as typeof Store;
+		const columnFamily = new RocksDatabase(new StoreClass(this.path, mergedOptions), mergedOptions);
```
kriszyp: This shallow spread shares reference-valued `encoder` and `decoder` instances between the parent and its supposedly independent view. Opening the child then mutates the shared encoder's `name` at `src/database.ts:986`, and clearing either view can reset its shared `structures` at `src/database.ts:163`, changing the other view's codec state and potentially making existing records undecodable. Please move column-family derivation behind a Store-owned factory/fork that creates independent codec state, or require codec factories/per-view instances instead of inheriting mutable codec objects. The retained configuration should also be snapshotted rather than remaining coupled to later mutations of the caller's options object.

kriszyp: `this.store.constructor` does not guarantee that the subclass can recreate an equivalent instance from `(path, options)`; the cast only suppresses that mismatch. A valid custom Store may require injected dependencies, retain runtime configuration, or consume a second constructor argument without forwarding it to `super()`. In those cases `db.use('events')` can throw, lose required state, or construct a Store whose native name remains `default`, so the returned view targets the wrong column family. The new test exercises only a subclass with the inherited base constructor and a field initializer. Please make derivation an explicit Store hook/factory (for example, `createColumnFamilyStore(name, options)`) that custom stores can override, or reject custom stores that do not provide one. Add coverage for a subclass with a custom constructor/runtime dependency and assert both `events.name` and data isolation.

kriszyp (commented):
Neat, this is a cool API. A few things to check on from comments, but I like it.

src/database.ts
```diff
@@ -1225,6 +1248,72 @@ export class RocksDatabase extends DBI<DBITransactional> {
...
+	 * ```
+	 */
+	use(name: string, options?: RocksDatabaseOptions): RocksDatabase {
+		if (typeof name !== 'string' || name === '') {
+			throw new TypeError('Column family name must be a non-empty string');
+		}
+		// This database is already bound to its own column family; return it
+		// (opened, honoring the create-and-open contract). `options` are ignored
+		// here — a bound column family cannot be reconfigured through `use()`.
+		if (name === this.#name) {
```
kriszyp: The `options` contract says these values override the inherited configuration, but this shortcut—and the cached-view branch below—returns an existing handle without validating or applying them. For example, `new RocksDatabase(path).use('default', { readOnly: true })` opens the original writable store, so writes still succeed; a later encoding or compression override for a cached view is likewise silently ignored. Please either remove `options` from `use()` (the simplest API), or compare effective configurations and reject incompatible reuse/create an appropriately configured handle. Please cover both own-name and cached-view calls with differing `readOnly`, encoding, and compression options.

src/store.ts
```diff
@@ -626,6 +649,47 @@ export class Store {
...
+	 * to control how a view's codec state is built.
+	 *
+	 * @param name - The column family name for the derived store.
+	 * @param options - Options overriding this store's, for the derived store.
+	 * @returns A new, independent `Store` bound to `name`.
+	 */
+	createColumnFamilyStore(name: string, options?: StoreOptions): Store {
+		const derived: StoreOptions = { ...this.#columnFamilyOptions, ...options, name };
+
+		if (sharesMutableCodec(derived.encoder) || sharesMutableCodec(derived.decoder)) {
```
cb1kenobi, line 675: **Medium: per-view encoder instance is rejected**

`sharesMutableCodec` runs on the merged options, so `db.use('events', { encoder: { encode, decode } })` throws even when the parent has no encoder (or a factory) and this instance would live only on the new store. Opening then assigns it to the view alone; the parent's codec is a different object.

Suggested fix: reject (or fork) only when `derived.encoder === this.#columnFamilyOptions?.encoder` (same for decoder). A caller-supplied per-view instance is already independent.

cb1kenobi (commented):
Reviewed 73385e4

src/database.ts
cb1kenobi: 73385e4 moves derivation behind `Store#createColumnFamilyStore` and snapshots construction options.

A naive `{...encoder}` clone is not safe here: msgpackr's `encode` is an own bound method, so a spread still mutates the original instance's `offset`/`structures`. The factory / named-`encoding` path already constructs a fresh encoder per view; a pre-constructed instance is rejected instead of shared.

One remaining hole (also on the new review): the reject runs on the merged options, so `use('events', { encoder })` throws even when that instance is exclusive to the view.

cb1kenobi: Addressed in 73385e4: derivation is now `Store#createColumnFamilyStore`, which custom stores override. The base implementation still reconstructs via `(path, options)` and asserts `store.name === name`, so a subclass that drops options cannot silently bind the default CF. Coverage is in the injected-dependency override test (name + data isolation) and the non-forwarding-subclass guard.

kriszyp (approved):
Perfect. Great work.
<sub>🤖 Reviewed (poorly) with Codex</sub>

PR #842, opened 2026-09-07
Add a native RocksDB storage lease for derived indexes

kylebernhardy:
Adds a build-gated [native storage lease](https://github.com/HarperFast/rocksdb-js/pull/842/changes?w=1#diff-ed1057f222d170d80d2529da82efb82352d25facb1ff6a5a6473852f7e3d6a0eR95) so another Node addon can safely use a caller-owned rocksdb-js database and column family without receiving RocksDB pointers or creating a second RocksDB owner. The lease provides bounded point reads, prefix scans, atomic batches, durability selection, lifecycle fencing, stale-column detection, and provider-owned buffer management.

This advances [“Epic: Provide the minimum native storage capabilities required by fulltext”](https://github.com/HarperFast/rocksdb-js/issues/831). Phase 0 remains hidden behind `ROCKSDB_JS_NATIVE_STORAGE_LEASE=1`; ordinary builds do not expose the lease method.

## For the human reviewer

1. The boundary exposes a versioned C callback table instead of `rocksdb::DB*` or C++ layouts. This adds a little ABI machinery, but isolates consumers from rocksdb-js's embedded RocksDB version and ownership model. Reversing this choice later would require replacing the addon-to-addon contract.
2. The shared [operation admission and close gate](https://github.com/HarperFast/rocksdb-js/pull/842/changes?w=1#diff-5ea8d049b841c98e052711ee247dc0a2b0cf833841c562c224501f6869f867a4R50) replaces the existing in-flight counter for ordinary operations as well as native leases. This keeps one close/drain invariant across JavaScript operations, checkpoints, backups, iterators, drops, and foreign native work; review the admission-versus-close ordering most closely.
3. [Native leases and Verification Table handles are mutually exclusive](https://github.com/HarperFast/rocksdb-js/pull/842/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R777) on one column family. Fulltext is expected to use dedicated derived-index storage, avoiding two independent concurrency protocols over the same keys.

## Verification

End-to-end route: the companion fulltext Phase 0 consumer exercised the lease against a real rocksdb-js database through Tantivy lifecycle tests. `pnpm check` passed; `pnpm test:native` passed 161 tests with 3 skipped, including the [operation-gate concurrency cases](https://github.com/HarperFast/rocksdb-js/pull/842/changes?w=1#diff-05a45207035012e3c2ed06ba2e94586de0ce59e6e788a4b973da17cb9e18f739R54); `pnpm test` passed 866 tests with 8 skipped; the feature-enabled `pnpm rebuild` and C11 header syntax check passed.

Comment generated by kAIle (GPT-5)

Complexity: complicated

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 6ec5c3e82afc</sub>

<sub>Human-Review-Need: 4 @ 6ec5c3e82afc</sub>


src/binding/core/operation_gate.h
```diff
@@ -0,0 +1,141 @@
...
+inline OperationClaim OperationClaim::acquireBorrowed(OperationGate& gate) {
+	return gate.tryAcquire() ? OperationClaim(&gate, {}) : OperationClaim();
+}
+
+inline OperationClaim OperationClaim::acquireShared(std::shared_ptr<OperationGate> gate) {
+	if (!gate || !gate->tryAcquire()) {
+		return {};
+	}
+	return OperationClaim(gate.get(), std::move(gate));
+}
```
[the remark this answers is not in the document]
kylebernhardy: Addressed in 6ec5c3e8 by capturing the borrowed gate pointer before moving the shared owner, so compiler argument-evaluation order cannot null the claim pointer. Comment generated by kAIle (GPT-5)

src/binding/native_storage/native_storage_lease.cpp
```diff
@@ -0,0 +1,592 @@
...
+	}
+	if (!retainContext(context)) {
+		return fail(context->state.get(), ROCKSDB_JS_STORAGE_CLOSED, status, "storage lease is closed");
+	}
+	uint8_t* copy = new (std::nothrow) uint8_t[static_cast<size_t>(length)];
+	if (length != 0 && !copy) {
+		releaseContext(context);
+		return fail(context->state.get(), ROCKSDB_JS_STORAGE_LIMIT, status, "owned value allocation failed");
+	}
+	if (length != 0) std::memcpy(copy, source, static_cast<size_t>(length));
```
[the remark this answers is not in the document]
kylebernhardy: Addressed in 6ec5c3e8 with a zero-length fast path that returns the provider release contract without retaining the lease or allocating byte storage. The companion fulltext test now covers an empty value. Comment generated by kAIle (GPT-5)

kriszyp:
What part of https://github.com/HarperFast/harper/issues/2489 indicates it should use direct interaction with Rocksdb-js? I thought the protocol was pretty explicitly in terms of JS APIs and events, wasn't it? Defining the harper encoding for another native add-on is not something we have committed to supporting.

kylebernhardy:
Thanks, Kris. This is not intended to be a requirement inferred from HarperFast/harper#2489, nor are we trying to establish Harper encoding as a generally supported contract for native add-ons.

The intent is to let the fulltext library use rocksdb-js directly as an optional data source alongside the native Tantivy filesystem. That gives us a way to benchmark the RocksDB-backed implementation directly and compare its raw performance with Tantivy native storage, independent of Harper integration.

If you feel adding and maintaining this native surface in rocksdb-js is not worthwhile for that use case, we can skip it.

Comment generated by kAIle (GPT-5)

src/binding/database/db_handle.cpp
```diff
@@ -344,6 +348,21 @@ void DBHandle::open(const std::string& path, const DBOptions& options) {
...
 	this->enableVerificationTable = options.verificationTable;
+	if (this->enableVerificationTable) {
+		auto registration = this->columnDescriptor->registerVerificationTableHandle();
+		if (registration != VerificationTableRegistration::Registered) {
+			DBKey key = descriptorKey(*this->descriptor);
+			this->columnDescriptor.reset();
+			this->descriptor.reset();
+			this->enableVerificationTable = false;
+			DBRegistry::PurgeIfUnreferenced(key);
+			if (registration == VerificationTableRegistration::Revoked) {
```
kriszyp, line 359: Folding the `revoked` check into VT registration adds an open-time throw that is unrelated to the lease feature and is inconsistent with how a non-VT open behaves on the same column family.

`ColumnFamilyDescriptor::revoke()` and `descriptor->unregisterColumnFamily(name)` are two separate statements in both `Database::Drop` and `Database::DropSync`. In the window between them, another worker thread calling `db.open()` on that name still gets handed the (now revoked) descriptor from `descriptor->columns`. With `verificationTable: true` that open now throws `Column family was dropped`; with `verificationTable: false` the identical open succeeds and proceeds exactly as it did before this PR. Harper opens record CFs with the VT enabled, and the drop path is explicitly designed around a broadcast where several worker handles all react to the same drop (see the `isColumnFamilyAlreadyDropped` comment just above `Drop`) — so this is precisely the shape of race the surrounding code already went out of its way to make idempotent.

The lease's own protection does not depend on this: `acquire()` re-checks `column->revoked` on every call, so a lease admitted in that window still fails fast. Suggest dropping `VerificationTableRegistration::Revoked` and having `registerVerificationTableHandle()` return `Registered` for a revoked family, leaving the existing downstream dropped-CF handling (write batches erroring, `clear()` treating `IsColumnFamilyDropped` as a no-op) to deal with it as it does today. That keeps the new registry focused on the one exclusion it exists for — VT handles vs. native leases.

If the throw is intentional, it would be worth stating in the design doc, since it changes `db.open()` semantics for the default Harper configuration.

src/binding/native_storage/native_storage_lease.cpp
```diff
@@ -0,0 +1,600 @@
...
+	if (!descriptor) return fail(state, ROCKSDB_JS_STORAGE_CLOSED, status, "storage lease is closed");
+
+	auto claim = OperationClaim::acquireShared(state->gate);
+	if (!claim) return fail(state, ROCKSDB_JS_STORAGE_CLOSED, status, "database is closing");
+	if (state->descriptor.load(std::memory_order_acquire) != descriptor || state->revoked.load(std::memory_order_acquire)) {
+		return fail(state, ROCKSDB_JS_STORAGE_CLOSED, status, "storage lease was revoked");
+	}
+
+	auto column = state->column.lock();
+	if (!column || column->revoked.load() || column->incarnation != state->columnIncarnation) {
```
kriszyp, line 161: These two identity comparisons can never evaluate true, so they read like stale-incarnation detection while providing none.

`state->column` is a `weak_ptr` to the exact `ColumnFamilyDescriptor` whose `incarnation` was copied in the constructor, and `incarnation` is `const`. So if `lock()` succeeds it returns that same object and `column->incarnation != state->columnIncarnation` is always false — the real protection is the `weak_ptr` expiring, plus the `revoked` flag. The same applies on line 164: `descriptor->vtEpoch` is `const`, `state->descriptor` is only ever mutated to `nullptr`, and a recycled `DBDescriptor` address is unreachable because the gate claim is what proves the object is still alive. Both checks are unconditionally-false branches.

That is not a bug — the invariants they assert do hold — but the design doc leans on them ("A stale lease fails by identity/lifetime validation; it never follows a reused pointer or numeric ID"), and a future reader could reasonably conclude that identity is what keeps this safe and then weaken the gate or the `weak_ptr` on that basis. Worth either dropping them, or keeping them with a comment saying they are assertions of an invariant established elsewhere (`incarnation` and `vtEpoch` remain genuinely useful as the ABI's diagnostic tokens either way).

src/binding/native_storage/native_storage_lease.cpp
```diff
@@ -0,0 +1,600 @@
...
+			if (remaining < 16 || keySize > remaining - 16 || valueSize > remaining - 16 - keySize) {
+				if (entries == 0) return fail(context->state.get(), ROCKSDB_JS_STORAGE_LIMIT, status, "scan entry exceeds page limit");
+				break;
+			}
+			appendU64(bytes, iterator->key().size());
+			appendU64(bytes, iterator->value().size());
+			bytes.insert(bytes.end(), iterator->key().data(), iterator->key().data() + iterator->key().size());
+			bytes.insert(bytes.end(), iterator->value().data(), iterator->value().data() + iterator->value().size());
+			++entries;
+			if ((entries & 63) == 0 && context->state->gate->isClosing()) {
```
kriszyp, line 462: Two things about this cancellation probe:

1. `(entries & 63) == 0` is evaluated after `++entries`, so the first check lands at entry 64. Any caller passing `entryLimit <= 63` — which is a perfectly reasonable page size for the recovery/GC scans this is meant for — never checks `isClosing()` at all. `entries % 64 == 0` on the pre-increment value, or simply checking every iteration (an `isClosing()` load is far cheaper than the `iterator->Next()` around it), would cover every page size.

2. When it does fire, the already-copied `bytes` are thrown away and the caller gets `CLOSED`. Since the gate claim is still held, the safe-and-useful alternative is to `break` and return the partial page — exactly what the byte-limit path does — leaving the consumer to discover the closure on its next call or via `poll_state()`. That matches the doc's own framing that "a non-empty partial page is returned if the read deadline expires after making progress": a close request is the same category of event, and discarding a page of work makes a clean shutdown more expensive than a timeout.

Neither affects the close guarantee itself (the claim bounds close regardless), but together they mean the cooperative-cancellation behaviour the doc describes is weaker than intended for small pages.

src/binding/transaction_log/transaction_log.cpp
```diff
@@ -336,8 +336,9 @@ napi_value TransactionLog::GetPath(napi_env env, napi_callback_info info) {
 	UNWRAP_TRANSACTION_LOG_HANDLE("GetPath");
 	auto store = (*txnLogHandle)->store.lock();
 	if (store) {
+		const auto path = store->displayPath.string();
```
kriszyp, line 339: This changes what `TransactionLog.getPath()` returns — from `store->path` (the resolved identity path) to `store->displayPath` (the caller's original spelling) — which is a user-visible behaviour change to a public API, arriving in a PR about native storage leases with no test and no mention in the description.

The change itself looks defensible: `AGENTS.md` is explicit that the identity path is captured at open specifically so a later `chdir()` or repointed symlink cannot re-resolve it, and handing a caller a canonicalized path they never supplied is arguably the bug. `TransactionLogStoreStats.path` (`transaction_log_store.cpp:519`) already uses `displayPath`, so this makes `getPath()` consistent with `getStats()`. Hoisting the `string()` into a local is also a strict improvement over calling it twice.

The ask is just to separate it: split it into its own commit or PR with a test that opens a store under a custom `transactionLogsPath` (relative, or via a symlink) and asserts `getPath()` returns the requested spelling. Bundled here it is easy to miss in review and hard to find later when someone bisects a path-comparison regression in a consumer.

src/binding/native_storage/native_storage_lease.cpp
```diff
@@ -0,0 +1,600 @@
...
+			if (mutation.struct_size < sizeof(rocksdb_js_storage_mutation) || mutation.key.length == 0 ||
+				(mutation.key.length != 0 && mutation.key.data == nullptr) ||
+				(mutation.value.length != 0 && mutation.value.data == nullptr) ||
+				(mutation.kind == ROCKSDB_JS_STORAGE_DELETE && mutation.value.length != 0) ||
+				(mutation.kind != ROCKSDB_JS_STORAGE_PUT && mutation.kind != ROCKSDB_JS_STORAGE_DELETE)
+			) {
+				return fail(context->state.get(), ROCKSDB_JS_STORAGE_INVALID_ARGUMENT, status, "invalid mutation");
+			}
+			if (mutation.key.length > kMaxKeyBytes || mutation.value.length > kMaxOwnedBytes ||
+				mutation.key.length > kMaxScanEntryPayloadBytes ||
```
kriszyp, line 362: `kMaxScanEntryPayloadBytes` is derived from the scan-page frame (`kMaxOwnedBytes` minus the page header and one entry's two length fields) and does not describe anything about a write batch. Using it here to bound mutation sizes is harmless today — `key.length` is already capped at `kMaxKeyBytes` on the same line, so the subtraction on the next line cannot underflow — but it couples the write limits to a constant that will move if the scan framing ever changes, and it makes the batch's actual limits (`kMaxKeyBytes`, `kMaxOwnedBytes`, `kMaxBatchBytes`) harder to read off.

The two `kMaxBatchBytes - totalBytes` checks below already enforce the real per-batch budget. Suggest dropping the two `kMaxScanEntryPayloadBytes` terms, or introducing a separate `kMaxMutationBytes` if a per-mutation cap distinct from the batch cap is actually wanted.

src/binding/database/db_descriptor.h
```diff
@@ -832,12 +838,44 @@ struct ColumnFamilyDescriptor final {
...
+		++this->nativeStorageLeases;
+		return true;
+	}
+
+	void unregisterNativeStorageLease() {
+		std::lock_guard lock(this->nativeUsageMutex);
+		if (this->nativeStorageLeases != 0) --this->nativeStorageLeases;
+	}
+
+	void revoke() { this->revoked.store(true); }
```
kriszyp, line 877: `revoke()` stores outside `nativeUsageMutex`, while both registration functions read `revoked` under it. A lease or VT registration can therefore observe `revoked == false`, and a concurrent `Database::Drop` can set it, with the registration still succeeding — the mutex gives no ordering against the store.

The practical impact is small: `acquire()` re-checks `column->revoked` per operation, so a lease registered across that race fails on first use, and a VT handle registered across it is in the same position it was before this PR. But since this registry exists precisely to make an exclusion decision, taking the lock in `revoke()` (`std::lock_guard lock(this->nativeUsageMutex); this->revoked.store(true);`) makes the decision actually atomic with respect to the two readers, at zero cost on a path that runs once per drop. The atomic type is still worth keeping for the lock-free per-operation reads in `acquire()` and `pollState()`.

.github/workflows/pr.yml
```diff
@@ -42,6 +42,12 @@ jobs:
           GH_TOKEN: ${{ secrets.GH_TOKEN }}
         run: pnpm build
 
+      - name: Build native storage lease
```
kriszyp, line 45: Two small things about this step:

1. Placing it before `Check` leaves `build/Release/*.node` as the lease-enabled binary for the rest of the job. `pnpm check` is TS-only so nothing is affected today, but if a native-dependent step is ever added to this job it will silently run against a non-default build. Moving this step after `Check` (or re-running `pnpm rebuild` without the variable afterwards) makes the compile-guard intent obvious and keeps the job's artifact in its default configuration.

2. This is a compile check on ubuntu only. `native_storage_lease.cpp` is never compiled by MSVC or AppleClang in CI, so the first Windows/macOS build of the experiment will be on whoever runs it locally. Adding the variable to one non-Linux matrix leg, or accepting the gap explicitly, would be worth deciding now rather than at promotion.

Separately: `docs/native-storage-lease-phase0.md` commits to "Release/prebuild/package workflows leave it off and assert the export is absent from every prebuild." I do not see that assertion in this PR or in the release workflow. Since the whole scope-containment argument for merging an unfinished ABI rests on it, it would be good to know whether that guard is in this PR's sequence or a follow-up.

AGENTS.md
```diff
@@ -731,6 +732,12 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     the transaction remains pending. rocksdb-js does not define record value layouts: a producer
     that copies `getTimestamp()` into record bytes must call `setTimestamp()` first.
 
+20. **Operation admission**: `OperationGate` uses sequentially consistent acquire/close ordering.
+    Every RocksDB access must hold a claim, and no RocksDB-owned pointer or slice may outlive it.
+    Borrowed claims require their caller to retain the owning descriptor until claim destruction;
+    async and foreign-thread work uses a shared claim so the gate survives its final release and
```
kriszyp, line 738: The clause "async and foreign-thread work uses a shared claim so the gate survives its final release and notification **without affecting `DBRegistry` descriptor reference counts**" reads as if it applies to all shared-claim users, but it is only true of the lease. `AsyncCheckpointState`, `AsyncBackupStreamState` and `AsyncCatchUpState` all hold a `std::shared_ptr<DBDescriptor>` for the duration of their work and therefore very much do affect the registry refcount — that is why each of their destructors has to retry `DBRegistry::PurgeIfUnreferenced()` after dropping it.

Suggest splitting the two ideas, roughly: a shared claim keeps the gate alive across its own final release and `notify_all()` even when the closer destroys the descriptor on observing zero; and separately, the native storage lease uses a shared claim *instead of* a `shared_ptr<DBDescriptor>` specifically so it does not defer the registry purge — a foreign thread has no safe point at which to retry it.

While here, it may be worth adding the destruction-order rule to this invariant, since it is the part most likely to be silently broken by a later edit: a claim must be declared before anything holding RocksDB-owned state (column-family handles, `PinnableSlice`, `WriteBatch`, `Iterator`) so those unwind first. `Admission` gets this right by member order and has a one-line comment, but the rule belongs here too.

kriszyp (commented):
At this point, this can probably be closed, right?

<sub>🤖 Reviewed with Claude</sub>

Issue #806, opened 2026-08-28
Optimistic commit racing DropColumnFamily latches a fatal background error under the default parallel OCC validation


kriszyp:
## What

An **optimistic** transaction whose commit races a `DropColumnFamily` from another thread does not fail at conflict validation as [#725](https://github.com/HarperFast/rocksdb-js/pull/725) assumed. Under RocksDB's default `OccValidationPolicy::kValidateParallel` the commit is admitted past validation, fails inside the memtable inserter with `Invalid argument: Invalid column family specified in write batch`, and `DBImpl::HandleMemTableInsertFailure` latches that as a **fatal background error on the whole environment**. From that instant every write to every column family on the path fails with the same message until the database is closed and reopened (or, on 2.8.0, `resume()`d).

#725 fixed this for non-transactional `putSync`/`removeSync` (`ignore_missing_column_families`) and kept transactions out of scope on the premise that

> Optimistic mode (the default, and the only mode Harper uses) already fails safely: conflict validation rejects a commit naming a dropped family early, with `Invalid argument: Could not access column family <id>`

That is only true of `kValidateSerial`. [#726](https://github.com/HarperFast/rocksdb-js/issues/726) tracks the same poisoning for pessimistic mode; this is the optimistic-mode counterpart, and it is the one Harper hits.

## Why

`OptimisticTransactionDB::Open(dbOptions, path, cfDescriptors, ...)` (`db_descriptor.cpp`, the overload without `OptimisticTransactionDBOptions`) leaves `validate_policy` at its default, `kValidateParallel`. `OptimisticTransaction::Commit()` then takes `CommitWithParallelValidate()`:

```cpp
  Status s = TransactionUtil::CheckKeysForConflicts(db_impl, *tracked_locks_, true /* cache_only */);
  if (!s.ok()) return s;                       // "Could not access column family N" only if already dropped HERE
  s = db_impl->Write(write_options_, GetWriteBatch()->GetWriteBatch());   // plain Write, no callback
```

[...]

PR #843, opened 2026-09-08
Gate transaction commits against concurrent column-family drops

kriszyp:
Refs HarperFast/rocksdb-js#806, HarperFast/rocksdb-js#726. A transaction commit racing another thread's `dropSync()`/`drop()` of the same column family used to reach RocksDB's memtable inserter naming the dropped family, and RocksDB latched that as a fatal background error on the whole database — every later write on every column family failed with `Invalid column family specified in write batch` until reopen. Under the default optimistic `kValidateParallel` policy conflict validation runs before the write thread and cannot see the drop; pessimistic commits have no validation at all. The binding now interlocks the two itself with a [per-column-family admission gate](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-3b2c4c36b470b912640968914e1dfee1839047c2239480149659aa402be06e9fR53): a commit takes admission on every family its batch names [immediately around `txn->Commit()`](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR435), and a drop [closes the gate before it waits](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R766) for admitted commits to finish, so the losing side of the race is always the commit, refused whole with a contained, attributable `ERR_COLUMN_FAMILY_DROPPED` (`column family "<name>" is being dropped` / `was dropped`) before it touches RocksDB, while the winning side's drop simply waits for the commit that is already inside RocksDB. Unrelated column families keep writing, `getLastError()` stays `null`, and the dropped name can be recreated immediately. Staging stays lock- and allocation-free for up to eight distinct families per transaction; the interlock costs two atomic read-modify-writes per distinct family per commit. The rules are recorded as [AGENTS.md invariant 23](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R892) (renumbered twice by later rebases onto `main`: 20→22 when `main` independently added invariants 20–21, then 22→23 when `main` added its own new invariant 22). Harper-side changes (bumping the pin, unskipping the harper#2456 reproducer) are a separate follow-up.

## For the human reviewer

1. **Admission is taken at commit time, not at staging.** The issue body's original sketch (a touch count taken on transactional `Put`/`Delete`) was rejected in the issue's own planning comment and again here: Harper calls `dropSync()` from inside a synchronous exclusive schema section while transactions staged on that thread are still open, so a staging-time hold would deadlock the drop against the only event loop that could release it. A staged, idle transaction therefore holds nothing and its later commit is what fails. A "no" here costs the Harper `dropTable()` path.
2. **[Overflow past eight distinct families allocates a vector at staging](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-3b2c4c36b470b912640968914e1dfee1839047c2239480149659aa402be06e9fR143)** (Gemini's round-1 alternative), replacing the planning review's preferred design of deriving the family set from the write batch at commit. The batch-derived path also allocated at commit (id and token vectors), took `columnsMutex` on the commit lane, and re-parsed every record of the batch on the lane that serializes every commit on the database; one vector allocation at the ninth distinct family of a transaction is cheaper on every axis, and `txn->Put` already allocates on every call, so staging was never allocation-free in the strict sense the objection assumed.
3. **Transactions pin gate tokens, never column-family descriptors.** A token is an atomic word plus a name, so a transaction that outlives its family — or, on the drain-timeout leak path in `TransactionHandle::close()`, the whole database — touches no RocksDB state when the token goes. This is what makes the lifetime story hold on every early exit without new ordering rules.
4. **A failed drop leaves the gate closed and is repaired by the retry, not by reopening.** RocksDB persists the OPTIONS file *after* removing the family, so a drop reported as failed may have succeeded; reopening the gate would race a second concurrent dropper. Instead the by-name registry entry is retired by whichever call next lands on RocksDB's already-dropped path, through an [identity-checked `unregisterColumnFamily(name, dropped)`](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1752) that can never erase a recreated same-name family. Harper's own semantics for a failed drop are tombstone-and-retry. Where to look hardest: [`dropColumnFamilyGated`](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R761) in database.cpp.
5. **Both `drop()` and `dropSync()` block the calling thread for commits already inside `Commit()`** — the same bound RocksDB's own `DropColumnFamilyImpl` imposes through the write thread — and [`drop()` now also holds the operations lock](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R814) like `dropSync()` always did. A commit stalled in a write stall (invariant 16) therefore stalls the drop with it; making `drop()` genuinely asynchronous is left as a follow-up.
6. **The log stage runs an [advisory `isDropping()` pre-check](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR372) before writing the transaction-log batch**, taken from the round-1 decision ledger: in the common drop-wins case the caller now sees the refusal itself rather than a `TransactionAbandonedError` over log bytes that were written for a doomed commit. It is advisory only — admission around `Commit()` remains the authoritative check — and a transaction that does lose after its log batch was written still surfaces as an abandonment, with the refusal [attached as `cause`](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-724f334a2b47a5e7cbcab252ce04fc232071d0bc058c96ed6fd4ac96e5dff308R1249).
7. **A refused commit is a hard error; `db.transaction()` does not retry it.** Correct for Harper's drop semantics (the old generation's data is gone), but a caller using drop-then-recreate as a truncate will see failures rather than convergence onto the new family. Reversible in TypeScript only.
8. **Test seams ship in the production binary** (`setCommitHoldForTesting`, `getCommitGateCountersForTesting`, `forceDropFailureForTesting`), consistent with the existing `forceTryAgainForTesting`. They are inert until armed by a test — a production commit pays one relaxed load — but `setCommitHoldForTesting(true)` parks every commit in the process for up to 30 s each, so it is as sharp as the existing seams.
9. **An `open()` by name during a drop's wait, or after a failed drop, binds to the dying generation.** The registry entry is retired only when a drop completes, so a handle opened in that window refuses every commit with `was dropped` until it is reopened; the [failed-drop retry test](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-16392a3a3806860a77f11bda437f184b06f47dc216aacb34fe587c7ff50dee73R592) pins this. Strictly better than the ghost-table behavior it replaces, and Harper reopens after its drop broadcast settles, but the only signal such a handle gets is the error code. Reversible by making `OpenDB` wait on `isDropping()` and create a fresh family.
10. **The build script now runs the native build as `pnpm run rebuild`** (package.json, the ASan workflow, and the two docs that spell the command). `pnpm/action-setup` with `version: latest` began installing pnpm ≥ 11 today, whose built-in `rebuild` command shadows this package's `rebuild` script, so CI's `pnpm build` silently skipped node-gyp and every runtime job tested the published 2.8.0 prebuild — this PR's first CI run shows it (`RocksDB/?` in the Vitest banner, `catchUpWithPrimary is not a function`, my gate tests seeing the old `ERR_INVALID_ARGUMENT`). One word, no behavior change under pnpm 10; the same fix is already on another open branch. *(The Windows `pnpm/action-setup` breakage described here was transient to that afternoon; a later rebase onto current `main` — see item 13 — no longer reproduces it, and every Windows job is green on the current head.)*
11. **Declined from the final review round (Gemini, advisory):** a null guard on `column->gate` inside `dropColumnFamilyGated` for the default family — both callers route the default family to `clear()` before reaching it, and a second check would be a second home for the same rule; and replacing the test seams' function-local static atomics with inline globals — they follow the existing `forceTryAgainCounter` pattern, and the guard is one predicted byte load per commit.
12. **Overruled from the planning review:** allocation-failure fault injection on the commit lane (the round-1 adjudicator also dropped it as immaterial). Admission runs inside an exception boundary whose fallback status carries no message and so cannot throw, but the lane already allocates unguarded on every path, so a dedicated `bad_alloc` test would exercise a failure class this change does not introduce.
13. **Rebased onto `main` (40 commits) and fixed two things the rebase and CI then surfaced.** `main` had independently added invariants 20–21, `ColumnFamilyDescriptor::maxWriteBufferSizeToMaintain`, and `droppedColumns` stall-inventory tracking in the same functions this PR touches (`unregisterColumnFamily`, `ColumnFamilyDescriptor`'s constructor, `test_seam.h`, `binding.cpp`'s N-API registrations) — merged to keep both sides' additions, renumbering this PR's invariant to 22 (item 1 above). The independent pre-push review then caught four stale `AGENTS.md invariant 20` comment cross-references the renumbering itself missed. Separately, CI on the rebased head reproduced a persistent Windows+Deno `EBUSY` on the new "worker terminated with its commit admitted" test's `rmSync` teardown — widening the retry budget up to `maxRetries:10` did not help (identical failure three times running), so it now follows the identical established precedent in `test/concurrent-teardown.test.ts` (retry, then swallow a persistent EBUSY — the test's actual assertions already passed by that point). No production code changed for either fix.
14. **Rebased onto `main` again (2 more commits: harper#849, a lock-callback env-cleanup fix) and renumbered again.** `main`'s own new commit independently claimed invariant number 22 (queued unlock callbacks belong to their env), so this PR's drop-gate invariant moved to 23 (item 1 above); the same five in-code `AGENTS.md invariant 22` cross-references the prior rebase had just fixed needed the same one-number bump. No production code changed — purely comment/doc renumbering, verified by a full local rebuild + `pnpm test`/`test:native`/`check` and a fresh full independent review round on the pushed head.
15. **New from that review round (domain-adjudicated minor, not fixed here):** `dropColumnFamilyGated` (database.cpp:761) closes the gate via `beginDrop()` before calling RocksDB's `DropColumnFamily`, and on any failure other than "already dropped" it returns without reopening the gate — per the existing "gate never reopens" design (item 4 above, [AGENTS.md invariant 23](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R892)). The concrete case the review surfaced: if RocksDB's `WriteOptionsFile`/VersionEdit persistence fails (e.g. transient `ENOSPC`) *before* the family is actually removed, the family stays live and registered but every future transactional commit to it is refused with `ERR_COLUMN_FAMILY_DROPPED` for the life of the process — no reopen path short of a full database restart. This is the same fail-closed-forever tradeoff the decision ledger already accepted for the "drop may have actually succeeded" case, just triggered from the opposite direction (drop did *not* succeed). Left as-is rather than redesigned in this rebase task — undoing it needs a drop-generation counter, which is a bigger change than a rebase's scope.

## Verification

- **Fails on base, passes with the fix.** The issue's reproducer (worker commits a 4 KB value while the main thread drops) against this worktree's build at the merge-base: optimistic poisons 8 of 15 runs (`ENV POISONED -> Put failed: Invalid argument: Invalid column family specified in write batch`), pessimistic 15 of 15. With the fix: 0 of 15 in both modes; every run's commit error is `Transaction commit failed: Column family dropped: column family "table" is being dropped` and the unrelated write succeeds.
- **New tests**: `test/native/column_family_gate_test.cc` (GoogleTest: gate protocol, concurrent admitters vs dropper, all-or-nothing unwind, inline and overflow tracking with the failed-write undo, generations) and `test/drop-commit-gate.test.ts` (Vitest) covering both modes × sync/async commit × sync/async drop for the drop-wins ordering; the [commit-wins ordering with observed barriers](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-16392a3a3806860a77f11bda437f184b06f47dc216aacb34fe587c7ff50dee73R344) (process-global admitted/drop-begun counters and a hold, no sleeps): the drop waits for the admitted commit, a second worker's `commitSync` submitted during the wait is refused *before* admission, and an unrelated family's write during the wait succeeds; multi-family all-or-nothing (live half never applied); a [staging write RocksDB itself refuses](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-16392a3a3806860a77f11bda437f184b06f47dc216aacb34fe587c7ff50dee73R229) (pessimistic, stale handle) [leaves no gate token behind](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-3b2c4c36b470b912640968914e1dfee1839047c2239480149659aa402be06e9fR166) so the live half still commits; override-only batch after its untouched home family is dropped; staged-but-idle transaction on the dropping thread; >8-family overflow including a recreated same-name family; a worker terminated with its commit admitted; two handles sharing one generation dropping concurrently; abort/close release; verification-table intents released after a refusal (a coordinated-retry writer does not park); a logged transaction in the drop-wins case is refused before any log bytes are written (`hasLog` false), and an abandoned logged transaction (forced `TryAgain` with no retry budget) carries its commit failure as `cause`; failed-drop retry retires the name; same-thread `dropSync()` with the thread's own admitted commit held under all three commit execution modes (`ROCKSDB_JS_COMMIT_THREAD` 0 / default / 2) in a watchdog-bounded child process; and, in two-lane mode, the [authoritative refusal after a multi-family transaction's log batch was already persisted](https://github.com/HarperFast/rocksdb-js/pull/843/changes?w=1#diff-1aee84855c9143937de38695c88d2a423613c46ed0ddb11f2b683a5b1ee42065R134) (abandonment with the dropped-family `cause`, live half not applied, committed-read watermark released).
- `pnpm test` on the pushed head (5f2cd1cd): 65 files, 928 passed, 9 skipped, 0 failed (Node 26, Linux). `pnpm test:native` (GoogleTest): 190 passed. `pnpm check` (type-check, lint, fmt:check): clean. Five pre-push review rounds (codex graded + Gemini, with Harper-domain adjudication on the four that touched native code): round 1 full, round 2 full after the overflow redesign, rounds 3–5 delta (5 covers only the build-script commit); every surviving finding was fixed in the following commit, and the ones the adjudicator dropped as factually wrong are noted in the dispatch log rather than here.
- **After the rebase onto `main` (item 13) and its two follow-up fixes**, re-ran locally: `pnpm build` clean, `pnpm test:native` 205/205, `pnpm test` 975 passed / 9 skipped (984, Node 26 Linux, including all 47 drop-commit-gate/drop tests), `pnpm check` clean. Four more pre-push review rounds ran across the rebase and its fixes (rounds 6–9), each fixing or verifying what the previous round found; two pre-existing-scope minors (a by-name-registry reopen race, an unsynchronized double read in `transaction_handle.cpp:727`) and one comment-narration nit were left for you rather than fixed here — see the dispatch log for the evidence trail, including a Gemini "major" use-after-free claim on `ColumnFamilyAdmission`'s raw pointers that a direct code trace refutes (`TransactionHandle::close()`'s drain-timeout path returns before reaching `stagedColumns.clear()`).
- **After the second rebase (item 14)**, re-ran locally: `pnpm build` clean, `pnpm test:native` 205/205, `pnpm test` 976 passed / 9 skipped (985, Node 26 Linux — one more test than the prior round's count, from `main`'s own rebased-in lock-callback-teardown test), `pnpm check` clean, and `test/drop-commit-gate.test.ts` alone (35/35). One more full independent review round ran on the pushed head (round 10 on this branch): every surviving finding was either already adjudicated in an earlier round and kept at its prior severity, an already-documented design tradeoff (item 15), or the pre-existing log-write-ordering item from item 6/the decision ledger — no new code change resulted.
- **CI is green on every platform on the current head** (native GoogleTest, Vitest on Node 22/24/26 × Ubuntu/macOS/Windows, Bun, Deno — including `Test on Deno (windows-latest)`, which needed item 13's teardown fix). The stale account below (first two pushes, before the rebase) is kept for the historical build-script diagnosis in item 10; it no longer describes the current head's CI state.
- **Benchmarks** (worker-thread harness, same toolchain, merge-base build vs this branch, plus the published 2.8.0 prebuild): the box was shared with other agents' review processes throughout, so the multi-worker numbers bound the effect rather than resolve it (round-to-round spreads of 2–6× on identical configs). Staging is the metric the change could regress; the only per-put work added is one pointer compare against the transaction's inline family set.

| measurement | base → new | Δ |
|---|---:|---:|
| single thread, 1 family, 1000 puts/txn, same JS bundle, native swapped (min of 8 interleaved rounds) | 1133 → 1177 ns/put | +3.9% |
| single thread, 3 families, same setup | 1433 → 1486 ns/put | +3.7% |
| control: same setup, a build with the per-put `note()` removed | 1177 → 1184 / 1437 → 1487 ns/put | +0.6% / +3.5% |
| single thread, 200 puts/txn, min of 6 interleaved rounds: 1 / 3 / 9 families | 1104→1127 / 1372→1414 / 1388→1425 ns/put | +2.1% / +3.1% / +2.7% |
| same, `commitSync` throughput (max): 1 / 3 / 9 families | 1962→1885 / 1648→1618 / 1650→1606 txns/s | −3.9% / −1.8% / −2.7% |
| 16 workers, 100 puts/txn, staging (median of 3): 1 / 3 / 8 / 9 families | 2782→3205 / 4622→4665 / 5167→5328 / 5339→5051 ns/put | +15% / +0.9% / +3.1% / −5.4% |
| 64 workers, 100 puts/txn, staging (median of 3): 1 / 3 families | 13393→14427 / 22165→25672 ns/put | +7.7% / +16% |
| 16 / 64 workers, async commit txns/s (median of 3), 1 family | 5516→6839 / 7049→7667 | +24% / +8.8% |
| 16 / 64 workers, `commitSync` txns/s (median of 3), 3 families | 15404→15202 / 13080→12467 | −1.3% / −4.7% |
| published 2.8.0 prebuild, single thread, 1 family (min) | 1114 ns/put | — |

The control row is the decisive one: removing the per-put addition entirely leaves the same +0.6–3.5% spread, so the residual is binary-to-binary layout variance and box noise, not the gate. The 8- vs 9-family rows cover the inline boundary and the vector overflow.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=cursor-composer,gemini,codex; adjudicated=domain; declined=cursor-grok; rounds=10; full=2 @ 350fce64f4d8</sub>

<sub>Human-Review-Need: 4 (decisions: drop-blocks-the-js-thread, admission-after-the-log-write, gate-never-reopens, staged-set-vs-batch-derivation, test-seams-as-shipped-exports) @ 350fce64f4d8</sub>


kriszyp:
### Heads up: this PR may be superseded by #850

Main's tip commit — [#850, "Defer physical column-family drops behind admitted commits"](https://github.com/HarperFast/rocksdb-js/pull/850), merged 2026-09-16 — is a complete alternative fix for the same bug this PR addresses (#806, #726). Its own new `AGENTS.md` invariant states it "subsumes the commit-time admission gate of PR #843." Every core file this PR touches (`database.cpp`, `db_descriptor.{h,cpp}`, `transaction.cpp`, `transaction_handle.cpp`, `test/drop.test.ts`) now conflicts with #850 at a design level — not additive, mutually exclusive mechanisms (this PR: commit-side gate + immediate physical drop on `drop()`; #850: admission claim taken earlier + physical drop deferred to the last releasing commit).

During routine rebase maintenance, an independent advisor review reached the same conclusion and recommended closing this PR as superseded, optionally porting just the "transaction spans more column families than the inline tracking set" overflow-path test to #850's implementation — the one behavioral case #850's own test suite doesn't already cover.

Maintenance has held off rebasing or closing this since that's your call, not a maintenance pass's. Let me know whether to close this as superseded, or if there's something here worth reconciling into main instead.

— Claude (automated PR maintenance)

PR #847, opened 2026-09-10
Throw when a range iterator fails instead of ending the range

### LINKED ISSUE #846: Range iteration ends silently when RocksDB's iterator fails (e.g. block checksum mismatch)
**Version:** @harperfast/rocksdb-js 2.8.0 (RocksDB 11.8.1), darwin-arm64 prebuild, Node 24.21.0. The 2.9.0 notes do not mention a change here.

**What happens**

When the underlying `rocksdb::Iterator` becomes invalid because of an error, rather than because the range is exhausted, `getRange()` ends as though the data had run out. No error reaches JavaScript, and `db.getLastError()` stays `null`.

A point read of the same corrupted block does throw (`Get failed: Corruption: block checksum mismatch …`), so the corruption is detected. The range path is the only one that drops it.

**Reproduction** (20,000 keys flushed to one SST; 128 bytes flipped in a data block):

| corruption location | `[...getRange({})].length` | `for await` length | error |
|---|---|---|---|
| first data block | 0 of 20,000 | 0 | none |
| middle of the file | 10,164 of 20,000 | 10,164 | none |

```js
import { RocksDatabase } from "@harperfast/rocksdb-js";
import { readdirSync, readFileSync, writeFileSync } from "node:fs";
const path = "/tmp/iter-status";
let db = new RocksDatabase(path, { keyEncoding: "binary", encoding: "binary" });
db.open();
await db.transaction((t) => { for (let i = 0; i < 20000; i++) t.putSync(Buffer.from(`k${String(i).padStart(6, "0")}`), Buffer.alloc(40, i & 0xff)); });
db.flushSync(); db.close();
const sst = readdirSync(path).find((f) => f.endsWith(".sst"));
const bytes = readFileSync(`${path}/${sst}`); for (let i = 64; i < 192; i++) bytes[i] ^= 0xff; writeFileSync(`${path}/${sst}`, bytes);
db = new RocksDatabase(path, { keyEncoding: "binary", encoding: "binary" }); db.open();
console.log([...db.getRange({})].length);   // 0, no error
db.getSync(Buffer.from("k000000"));          // throws Corruption
```

**Cause**

`src/binding/iterator/db_iterator.cpp`, `DBIterator::Next`:

```cpp
if (!it->valid()) {
    if (it->iterator && !it->iterator->status().ok()) {
        DEBUG_LOG("%p DBIterator::Next iterator not valid/ok: %s\n", ...);
    } else {
        DEBUG_LOG("%p DBIterator::Next iterator no keys found in range\n", ...);
    }
    NAPI_STATUS_THROWS(::napi_create_uint32(env, ITERATOR_RESULT_DONE, &result));
    return result;
}
```

The non-OK status is detected but only logged in debug builds. Then the same `ITERATOR_RESULT_DONE` is returned as for normal exhaustion. RocksDB's contract is that `!Valid()` means "check `status()`", and a non-OK status is an error, not the end of the range.

**Expected**

`Next` throws (or rejects, on the async path) with the status message, as `get` does. For comparison, the discontinued `rocksdb` (leveldown-style) binding surfaces the same corruption as an error from `iterator.next()`.

**Why it matters**

A consumer that scans, such as an index rebuild, a backup, or a query over a range, gets a silently truncated result from a damaged file. That is indistinguishable from a smaller dataset, so the damage propagates instead of being caught.

ivanjeremic:
Fixes #846.

## Problem

When the underlying `rocksdb::Iterator` becomes invalid because it failed, not because the range ran out, `DBIterator::Next` returned `ITERATOR_RESULT_DONE`, the same signal as normal exhaustion. The non-OK `status()` was detected, but only written to the debug log. A block checksum mismatch therefore made a range read as if the data had ended: 0 of 20,000 entries with the damaged block at the start of the SST, and 10,164 with it mid-file. No error was raised, while `get()` on the same block does throw `Corruption`.

## Change

- **`src/binding/iterator/db_iterator.cpp`:** when the iterator is invalid and `status()` isn't OK, `Next` throws through the existing `ROCKSDB_STATUS_THROWS_ERROR_LIKE` macro. The error carries the usual mapped `code` (`ERR_CORRUPTION`, `ERR_IO_ERROR`, …) and the message `Iterator failed: <status>`. Normal exhaustion, and the end of a bounded range, still return `DONE`.
- **`src/dbi-iterator.ts`:** if the native `next()` throws, the native iterator is released before the error is rethrown. A `for...of` loop doesn't call `return()` when `next()` throws, so without this a failed iterator, and the snapshot it holds, would stay alive until garbage collection.

## Tests

`test/iterator-status.test.ts` writes 20,000 keys, flushes them into one SST and damages a data block:

- A synchronous range over a corrupted first block throws `ERR_CORRUPTION`.
- `for await` over the same range rejects the same way.
- With the damage mid-file, the entries before it are yielded, then the range throws instead of ending early.
- The native iterator's `return()` runs when a step fails.

Against the unpatched build, all four new tests fail; a range over the damaged first block "ended after 0 of 20000 entries without an error". With the change, all four pass.

- `pnpm test`: 67 files, 931 tests passed, 8 skipped.
- `pnpm check`: clean.
- The downstream project that found this ([Zelavis](https://github.com/zelavis/zelavis), which is evaluating rocksdb-js as its RocksDB engine) was run against this branch. Its adapter, key-value conformance, SIGKILL-durability, truncated-WAL and concurrency suites pass. Its test for this exact case, "a scan over a corrupted block fails rather than ending early", passes where it had been marked as a known gap.

Tested on macOS arm64, Node 24.21.0.


kriszyp (approved):
This looks great to me, good catch (pun intended)! (It's always hard to remember if this should be iterator.return() or iterator.throw(), but I think return is probably right and does the job).

cb1kenobi (approved):
This is great! Thank you for the contribution!

kriszyp:
And Zelavis looks cool, btw! Thanks again for the contribution, hope this project works well for you!

ivanjeremic:
> And Zelavis looks cool, btw! Thanks again for the contribution, hope this project works well for you!

Thank you!

PR #849, opened 2026-09-11
Release a dying env's queued unlock callbacks from its cleanup hook

### LINKED ISSUE #848: tryLock callback queued by a terminated worker aborts the process on unlock (Node 22)
## Summary

A `tryLock(key, callback)` waiter registered from a worker thread that is later `worker.terminate()`d makes the **whole process SIGABRT** when another thread calls `unlock(key)`. Node 22.23.1 aborts; Node 24.18.0 survives. No Harper code involved.

## Repro

```js
// node repro.js lock      → Node 22: "Aborted (core dumped)" after "terminated"; Node 24: SURVIVED
// node repro.js buffer    → survives on both (control: getUserSharedBuffer listener + notify)
const { Worker, isMainThread, parentPort, workerData } = require('node:worker_threads');
const path = require('node:path'), fs = require('node:fs'), os = require('node:os');
const { RocksDatabase } = require('@harperfast/rocksdb-js');
if (isMainThread) {
	const dbPath = path.join(fs.mkdtempSync(path.join(os.tmpdir(), 'rlock-')), 'db');
	const db = new RocksDatabase(dbPath).open();
	if (!db.tryLock('k', () => {})) throw new Error('main should own');
	const w = new Worker(__filename, { workerData: { dbPath, mode: process.argv[2] || 'lock' } });
	w.once('message', async (m) => {
		console.log('worker:', m); await w.terminate(); console.log('terminated');
		if (workerData?.mode === 'buffer' || process.argv[2] === 'buffer') { db.getUserSharedBuffer('r', new ArrayBuffer(8), {}).notify(); console.log('notified'); }
		else { db.unlock('k'); console.log('unlocked'); }          // ← abort here on Node 22
		await new Promise((r) => setTimeout(r, 500)); db.close(); console.log('SURVIVED'); process.exit(0);
	});
} else {
	const db = new RocksDatabase(workerData.dbPath).open();
	if (workerData.mode === 'buffer') { globalThis.keep = db.getUserSharedBuffer('r', new ArrayBuffer(8), { callback: () => {} }); parentPort.postMessage('buffer callback registered'); }
	else { db.tryLock('k', () => parentPort.postMessage('cb')); parentPort.postMessage('pending unlock callback registered'); }
	setInterval(() => {}, 1000);
}
```

Observed with `@harperfast/rocksdb-js` as pinned by harper `main` (`6010ea9fd`), macOS arm64 and GitHub Actions ubuntu (harper Unit Test Node 22 legs, runs 34637871368 / 34637971903).

## Cause

`DBDescriptor::lockReleaseByKey` (`src/binding/database/db_descriptor.cpp` ~1185) pops the queued `LockCallback`s and calls `napi_call_threadsafe_function(cb, nullptr, napi_tsfn_blocking)`. The callback's tsfn belongs to the terminated worker's env. Node 24 returns `napi_closing`, which the loop skips; Node 22 aborts inside the call. Queued lock callbacks have no per-env cleanup — `ParkTimeoutRegistry` has `releaseByEnv(napi_env)` for exactly this (`db_descriptor.h:130`), lock callbacks do not. `lockReleaseByOwner` covers `db.close()` but a terminated worker never closes.

## Impact

Harper `manageThreads.js` calls `worker.terminate()` on worker restart, so any pending lock waiter in the dying worker (derived-index runner, `recordLock.ts` contenders) can take down the main process on Node 22 when the holder unlocks. Harper's derived-index runtime is switching to `tryLock(key)` without a callback and waking through `getUserSharedBuffer` `notify()` (harper PR pending); `recordLock.ts` still relies on the callback.

## Suggested fix

Register queued lock callbacks with an env-scoped registry and an `napi_add_env_cleanup_hook` that, on env teardown, marks/removes that env's `LockCallback`s and releases their tsfns under the same mutex `lockReleaseByKey` takes — the `ParkTimeoutRegistry::releaseByEnv` pattern. A regression test in the shape of `notify-teardown-uaf.test.ts` (child process, expect exit 0 / no signal) with the repro above.

kriszyp:
Fixes #848.

## Defect

A `tryLock(key, callback)` waiter registered from a worker thread that is later `worker.terminate()`d leaves its callback — a threadsafe function bound to the dead env — queued on the shared `LockHandle`. When another env calls `unlock(key)`, `DBDescriptor::lockReleaseByKey` calls that tsfn: Node 24 returns `napi_closing` (skipped), **Node 22 aborts the process**. Queued lock callbacks had no per-env cleanup; parks already do (`ParkTimeoutRegistry::releaseByEnv`), and so do commit completions and event listeners.

Found when harper#2430 landed: its derived-index runtime keeps a standing lock waiter on every non-owner worker, Harper's thread manager terminates workers on restart, and every Node 22 unit-test run on harper `main` aborted (`Aborted (core dumped)`, exit 134). `recordLock.ts` uses the same primitive, so the exposure predates that runtime.

## Fix

- `LockCallback` records the `napi_env` that queued it.
- `DBDescriptor::releaseLockCallbacksByEnv(env)` drops a dying env's callbacks from every lock's queue (releasing their tsfns, never calling them), under `locksMutex`.
- `DBRegistry::ReleaseLockCallbacksByEnv` snapshots descriptors and calls it; wired into the module env-cleanup hook in `binding.cpp` next to `ReleaseParkTimeoutsByEnv`.
- `lockReleaseByKey` / `lockReleaseByOwner` now hold `locksMutex` **across** the tsfn calls instead of moving the queue out and calling unlocked. That is the same closure as `EventEmitter::notify` (harper#1370): the cleanup hook takes the same mutex, so it either removes a callback before the call or waits until the call has returned, and Node cannot free the tsfn under `napi_call_threadsafe_function`. Calling a tsfn only enqueues onto its env's loop, so nothing re-enters the mutex.
- AGENTS.md invariant 22 records the rule.

## Verification

- `test/lock-teardown-abort.test.ts` (child-process repro, same shape as `notify-teardown-uaf`): 3 rounds of queue → `terminate()` → `unlock()`, then a live waiter that must still be woken.
  - **Unfixed binding, Node 22.23.1: exit 134** (control).
  - Fixed binding: exit 0 on Node 22.23.1 and 24.18.0.
- `lock.test.ts`, `lock-tracker.test.ts`, `notify-teardown-uaf.test.ts`, `lock-teardown-abort.test.ts`: 30 passing (Node 24).
- Harper: `unitTests/resources/vectorIndexPlane.test.js` (the suite that aborted harper `main`'s Node 22 CI) against this binary with **no Harper change**: 22 passing on Node 22.23.1.
- `oxfmt --check`, `oxlint` clean. `node-gyp rebuild` clean on macOS arm64.

Deno/Bun skip the new test for the same reason `notify-teardown-uaf` does: the guarantee under test is Node's env-cleanup ordering.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -872,6 +872,24 @@ sufficient (env teardown does not honor tsfn acquire counts); see
     a child process the parent kills on a deadline, because the stalled writer blocks the JS thread
     and the runner's own timeout cannot fire (#781 item 2).
 
+22. **A queued unlock callback belongs to its env and is released by that env's cleanup hook**:
+    `tryLock(key, callback)` on a held key queues the callback as a threadsafe function of the
+    caller's env on the shared `LockHandle` (`DBDescriptor::lockEnqueueCallback`), and only the
+    holder's `unlock()` (`lockReleaseByKey`) or `db.close()` (`lockReleaseByOwner`) ever calls it. A
+    `worker_threads` env that is `terminate()`d never closes its handles in order, so its callback
+    stayed queued on a lock another env held; that env's later `unlock()` then called a tsfn Node had
+    already freed — Node 24 returns `napi_closing`, Node 22 aborts the process (rocksdb-js#848;
+    harper's derived-index runner kept exactly such a waiter on every non-owner worker, and Harper's
+    thread manager terminates workers on restart). Every `LockCallback` now records its `napi_env`,
+    and `DBRegistry::ReleaseLockCallbacksByEnv` → `DBDescriptor::releaseLockCallbacksByEnv`, wired
+    into the module env-cleanup hook beside `ReleaseParkTimeoutsByEnv`, releases (never calls) the
+    dying env's callbacks. The release paths hold `locksMutex` **across** their tsfn calls for the
+    same reason `EventEmitter::notify` holds its mutex (harper#1370): the cleanup hook takes that
+    mutex too, so it either removes a callback before the call or waits until the call has returned,
+    and Node cannot free the tsfn under `napi_call_threadsafe_function`. Calling a tsfn only enqueues
+    onto its env's loop, so holding the mutex across it cannot re-enter. `test/lock-teardown-abort.test.ts`
+    is the child-process repro; it also proves a live waiter is still woken.
+
 ## Debugging native heap corruption
 
 AddressSanitizer is the first choice (`ROCKSDB_ASAN=1 node-gyp rebuild` toggles `-fsanitize=address`
--- a/src/binding/binding.cpp
+++ b/src/binding/binding.cpp
@@ -232,6 +232,9 @@ NAPI_MODULE_INIT() {
 		// tsfns, so the descriptor's park-timeout thread never fires into a
 		// torn-down env.
 		rocksdb_js::DBRegistry::ReleaseParkTimeoutsByEnv(dyingEnv);
+		// And this env's queued unlock callbacks: a lock another env holds would
+		// otherwise call them after Node freed their tsfns (rocksdb-js#848).
+		rocksdb_js::DBRegistry::ReleaseLockCallbacksByEnv(dyingEnv);
 
 		int32_t newRefCount = --moduleRefCount;
 		if (newRefCount == 0) {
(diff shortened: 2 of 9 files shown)


kriszyp:
Harper side: #2574 (harper) is the temporary workaround that unblocks `main` today; HarperFast/harper#2576 tracks reverting it once a release containing this PR is pinned. — Claude Fable 5.1

cb1kenobi (approved):
Good catch!

Issue #784, opened 2026-08-14
Make TransactionHandle close exclude new work and fully drain in-flight operations


kriszyp:
Parent: #741

## Problem

`TransactionHandle::close()` and transaction entry points do not share one lifecycle gate:

- `closed.exchange(true)` is separate from async-work registration, so close can observe zero work and begin teardown immediately before another operation registers.
- The current wait is bounded to five seconds, after which close mutates transaction-log state, releases VT intents, and deletes `txn` even if native execution is still using it.
- `state`, `txn`, `lockedVTSlots`, and `heldTrackers` have cross-thread access without one complete synchronization contract.
- Synchronous `getSync`, `putSync`, `removeSync`, `CommitSync`, and `getCount` remain outside the async-work drain mechanism.

PR #780 must make its new environment-cleanup reap safe for already-running native execution. This issue tracks the complete cross-environment close invariant, including entry points outside the narrow #780 fix.

## Required invariant

After close begins, no new operation may acquire transaction state. Every operation that acquired it earlier must finish before `state`, `txn`, VT intents, log state, or the owning database handle can be mutated or destroyed.

## Acceptance criteria

- Make the closed check and work/resource acquisition atomic with the start of close.
- Reject new operations once close has claimed the handle.
- Do not destroy native state after a timeout while work remains active; defer destruction or wait for definitive native completion.
- Synchronize transaction state and VT intent-vector ownership across close, commit, abort, reset, and release paths.
- Define and enforce the same safety contract for synchronous entry points without imposing an unjustified hot-path cost.
- Add deterministic regressions for:
  - environment teardown during a native commit lasting longer than five seconds;
  - close racing async work registration;
  - close racing representative synchronous reads and writes.

## Relationship

- Umbrella: #741
- #780 should resolve the already-running-work blocker in its new cleanup path.
- Previous partial implementation/reference: #745
- Related teardown symptom: #746

— KrAIs (Codex), on behalf of @kriszyp

PR #850, opened 2026-09-13
Defer physical column-family drops behind admitted commits

kriszyp:
Refs HarperFast/rocksdb-js#806, HarperFast/rocksdb-js#726. A transaction commit already inside RocksDB naming a column family that another thread dropped failed in the memtable inserter with `Invalid column family specified in write batch`, and RocksDB latched that as a fatal background error on the whole database: every later write on every column family failed until reopen (harper#1381). `drop()`/`dropSync()` now [**retire the family logically and return**](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1769): the name leaves the registry at once and a later `open()` of it creates a fresh family, while the physical `DropColumnFamily` runs only when no commit holds a claim on that generation. A commit [**claims** every family its batch names](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR290) once, at admission before the transaction-log write, and [releases right after `txn->Commit()`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR608); the last releaser of a retired generation [runs the drop](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1945). With no admitted commit (the common case) the drop is physical before the call returns, exactly as before. The protocol is Node-free in [`core/column_family_lifetime.h`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-3bca5afc069b910c20cc86a2f2526f8716c04974faf9aff82bcf811d70ced659R24) and GoogleTest-covered; the rules are [AGENTS.md invariant 23](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R947).

Consulted with Kris before implementing (dispatch task `rocksdb-js-deferred-cf-reclamation`): rulings A1 (retired family refuses transactional staging and commit with `ERR_COLUMN_FAMILY_DROPPED`; reads through retained handles continue; non-transactional writes keep #725's silent discard), B1 (commit-duration admission rather than staging-lifetime pins), C1 (`open()` of a still-claimed name waits, bounded, then creates the fresh family), D1 (#843 is subsumed and left untouched), E1 (the cross-restart window is documented, no tombstone in rocksdb-js).

**Framing-Verdict: chosen-approach-sound** (planning round 2, codex). Round 1 returned `better-alternative-exists` against the first draft, which pinned generations from staging until the transaction closed; its alternative, commit-duration admission, was adopted because the disqualifiers hold in this checkout (see item 1).

## For the human reviewer

1. **Pins are taken at commit admission, not at staging, and staging records touched families weakly.** The task's literal shape ("staged transactions pin the family until they release") was rejected in planning: an idle, abandoned, or drain-timeout-leaked (#784) transaction would block reclamation and same-name recreate, permanently on the leak path; `resetTransaction()` reruns the coordinated-retry callback, which may touch a different family set, so staging pins go stale; and an advisory pre-log check plus an authoritative pre-commit check yields a logged-then-refused commit surfacing as `ERR_TRANSACTION_ABANDONED`. On the first attempt, admission is [one decision before any log byte](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR437), held through `txn->Commit()`. The existing `txn-close-commit-uaf` fixture then found (SIGSEGV, 3/3) that even the *touched* set must be weak: a strong reference let an aborted transaction JS still referenced destroy the RocksDB column-family handle after `finishClose()` had destroyed the database. The commit locks the weak references only for its claim ([`noteTouchedColumnFamily`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-5d0063e928e32a18b23d6ce43f80fbd6fee68e69c9bccd4c792307bae0e9d4fdR717) records them); a lock that fails is a dropped-and-reclaimed generation and refuses the commit like a retired one.
2. **Only write batches pin; reads, iterators, async gets and user shared buffers do not.** Measured on the pinned build: after a physical drop a retained handle still serves `get`, iteration and counts for every key (RocksDB's handle refcount keeps the dropped `ColumnFamilyData` readable), a non-transactional write is discarded (and, from round 2, is skipped rather than written while the physical drop is still deferred, so it cannot come back with a resurrected family), a transaction staged afterwards fails at validation. The only hazard is a batch naming the id entering the write thread after `SetDropped()`. An "every owner pins" design would also block `open()` of the dropped name until every worker's JS handle closes or is collected (RocksDB cannot hold two families of one name), which breaks the existing "free the name immediately" test and ties correctness to GC.
3. **`open()` of a name whose previous generation is still claimed waits** (20 ms slices, `databasesMutex` released during the wait, bounded by `ROCKSDB_JS_CF_RECLAIM_WAIT_MS`, default 30 s, armed on the first such wait, then a retryable error). Deadlock-free because a claim is only ever held by a commit inside RocksDB on a lane, a libuv thread, or another thread's `commitSync`, never parked on the opener's event loop. The polling slice exists because the reclaimer notifies without `databasesMutex`, so a notify between the predicate check and the wait would otherwise be lost. The cost you chose in ruling C1, and the one both review rounds flagged: a commit slowed by a write stall turns this into an event-loop stall of up to the bound on the opening worker, where an open never waited before.
4. **A failed physical drop is never silently lost, but the caller-visible report depends on who ran it.** The drop that `drop()`/`dropSync()` runs itself rejects/throws with RocksDB's status and the name stays retired; a deferred drop that fails on a commit lane is reported through the global `log.warn` event and `columnFamily.pendingReclaims`. Both are retried on the next drop on the database (before that drop's own attempt, so a failure it reports stays visible until the next retry point), the next `open()` of that name (inline, `columnsMutex` released; the open throws only when its *own* attempt fails, and waits when another thread holds the retry), and `finishClose()` (review round 2 caught that the close-time retry stood down on the closing flag it had just set; it now runs past that gate, and the test proves a real drop on close). Idempotent because RocksDB removes the family before it persists OPTIONS: a post-MANIFEST failure retries as "already dropped". A `forceDropFailureForTesting(0|1|2)` seam ships in the production binary like `forceTryAgainForTesting`; mode 2 performs the real drop and then reports failure to prove that path.
5. **Cross-restart is documented, not solved.** A process that exits while a physical drop is pending (only while a commit admitted before the drop is still inside RocksDB) or after one failed leaves the family on disk under its name, and the next open lists it as live; a backup taken inside that window copies it. The same is true of a concurrent reader while the window is open: a read-only or secondary open is a separate registry key with its own descriptor, and it enumerates the on-disk families, so it lists the retired generation as live until the physical drop runs (`retireColumnFamily` returns `NotSupported` on a read-only descriptor, so such a handle never retires one itself). The immediate drop had the same exposure after a failure and none during the window, which did not exist. `rocksdb-js` deliberately does not provide a durable tombstone. Downstream integration must retain and recover its own catalog tombstone until physical reclamation completes; current Harper removes its catalog rows immediately after `dropSync()`, so that integration follow-up is still required before crash durability can be claimed.
6. **PR #843's commit gate is subsumed and its branch untouched.** Every physical drop now goes through `reclaimColumnFamily` at `admitted == 0`; this is #843's admission taken once, earlier, with its blocking wait replaced by deferral to the last releaser, so no immediate-drop path remains for the gate to protect. Whether #843 lands, is rebased onto this, or is closed is Kris's call — the tour's last section has the measured comparison.
7. **`OpenDB` decides under `columnsMutex`, and asks the reclaim gate rather than a parallel state.** Review round 1 (cursor-grok, adjudicated blocker) found the first cut deciding on a sample taken before the create's critical section, so a drop landing in between could make the open physically drop a generation still claimed by an admitted commit. The [decision and the create now share one `columnsMutex` section](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R556), and the open calls [`reclaimColumnFamily`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1838) and branches on whether it actually ran the drop; `reclaimColumnFamily` itself re-checks `admitted == 0` under its claim, so no path can drop a claimed generation.
8. **Teardown and retry are close-safe without leaking the process-global descriptor.** A queued commit registers `operationsInFlight` before leaving the handle's async-work admission. The descriptor-owned commit workers then use a raw descriptor pointer; `finishClose()` waits for that operation and drains and joins the log and commit lanes before destroying the descriptor. The legacy libuv path instead retains a strong descriptor for its independent worker and explicitly retries the registry purge after releasing it. Normal JS completion also performs the authoritative purge retry, while an env-gone completion never touches N-API or re-pins the descriptor. The child matrix covers close-after-admission and forced retry-after-close in modes 0, 1, and 2, and a sole-worker-env termination test proves the committed value reaches disk, the registry entry is removed, and an independent process can take RocksDB's lock.
9. **Retained handles cannot mutate a retired generation.** `putSync`/`removeSync` are discarded (as #725 already did after the physical drop, now also inside the deferral window) and `clear()`/`clearSync()` are no-ops, so a generation that comes back after a crash or a failed drop comes back as it was at the drop, never partially cleared or with post-drop writes.
10. **Review follow-ups incorporated.** Admission refusal now abandons writes and releases verification-table intents; a failed pessimistic `Put`/`Delete` removes the touched-family entry it reserved; synchronous log failures release claims through the live descriptor; async drop errors use the normal coded RocksDB error; touched/claimed vectors use two inline slots; all three asynchronous commit modes retain safe close context and retry the registry purge; `InlineVector` guards empty removal; and `RegistryStatus` snapshots the column map and user-buffer count under their owning mutexes. An admitted commit is intentionally successful even though its retiring-family data is then discarded, and its transaction-log entries remain published; downstream consumers must order the schema drop after those entries. The remaining design tradeoffs are the approved C1 event-loop wait, the E1 cross-restart gap, and the separate read-only/secondary descriptor visibility window.
11. **Dropped-family errors are consistent across entry points.** Non-transactional `putSync`/`removeSync` errors now use `createRocksDBError`, so a transactional write routed through `db.put(key, value, { transaction })` reports `ERR_COLUMN_FAMILY_DROPPED` like `Transaction.putSync` does. Staging-time and commit-time messages now also follow the repository convention by beginning with `Column family`.
12. **The `retiring` entry is a list of generations, not a state machine — deliberately, after the fact.** An earlier revision of this branch gave each entry a `Pending`/`Reclaiming`/`Failed` enum and a `lastError` string. `Reclaiming` was exactly `lifetime.reclaimClaimed` and `Pending`-with-no-claim was exactly `admitted == 0`, so the enum was a second copy of atomic state kept in step by hand under `columnsMutex`, with three readers each restating the predicate; `lastError` was never read. Asking `reclaimColumnFamily` — which already reads those atomics and reports whether it ran the drop — removed 58 source lines and one latent bug (the open's "another thread holds the claim" wait had no deadline at all; it is now under `ROCKSDB_JS_CF_RECLAIM_WAIT_MS` like every other wait). The tradeoff the adjudicator named and I accepted: `columnFamily.pendingReclaims` no longer distinguishes "deferred behind a live commit" from "drop failed and keeps failing" — the `log.warn` event is the only signal for the second. [AGENTS.md invariant 23](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R999) records that a status enum must not come back.

13. **The reclaim re-check loop has no iteration bound, and does not need one.** [`reclaimColumnFamily`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1853) re-claims after unclaiming when `admitted` fell to 0 in between, because a retirer can observe the transient increment of an admission that is about to be refused, and returning there would strand the generation with nobody to drop it. The last review round read that as an unbounded hot spin on Harper's synchronous schema thread under heavy commit load. The load it describes drives the *exit*, not the spin: a sustained non-zero `admitted` takes the `return` at the second load and leaves the drop to the last releaser, and the loop only re-arms when a transient increment **and** its matching decrement both land between two loads a few nanoseconds apart — an independent coincidence per iteration, at a rate bounded by how fast a commit can reach admission. Two threads in the loop cannot ping-pong either: `claimReclaim()` is a CAS, so the loser returns at once. Left as written rather than adding a yield whose only justification would be that coincidence chain.

14. **Post-approval round: the default commit lane now rechecks `isClosing()` after publishing its operation.** The pre-push review found the gap this PR's own new accounting left open: `registerCommitCompletion` and `registerDescriptorOperation` ran back to back with no recheck between them, and `commitCompletionsClosed` is set only *after* `finishClose()` has passed its drain gate and stopped both lanes — so a close racing another env could pass that gate in the window, and `CommitWorker::enqueue` runs a task inline once the lane is stopped (`src/binding/database/commit_worker.h`, unchanged by this PR). The [recheck](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR993) is the same publish-then-check shape the legacy path 50 lines below already used, so teardown either waits for the operation or the commit observes the close and rejects. It also balances `registerCommitCompletion` (nothing else would unref that tsfn once no dispatch follows) and releases the operation while the local descriptor pin is still alive — the follow-up round caught that leaving it to `~TransactionCommitState` decremented `operationsInFlight` through a pin destroyed one declaration earlier. **What is not here:** a discriminating deterministic test. Reaching the window needs a fourth production test seam between the two registrations, and the same review is already carrying a standing nit about the per-write seam load; the existing close fixtures delay only *after* the operation is registered, so they exercise the rejection's cleanup but not the race that selects it. Worth your judgment on whether the seam is worth adding.

### What `drop()` promises when it returns
Should a drop report success once the name is unusable, or only once the bytes are gone?

This PR splits one operation into two. **Logical retirement** is what the caller waits for: the name leaves `db.columns`, a same-name `open()` yields a fresh empty family, and every transactional write to the old generation is refused. **Physical deletion** is what RocksDB does, and it now happens on whichever thread last stops using the generation — usually the dropping thread itself, because in the common case nothing else is using it.

| Before | After |
| --- | --- |
| `dropSync()` called `DropColumnFamily` immediately.<br>A commit already inside RocksDB naming that family failed in the memtable inserter, and RocksDB latched a **fatal background error on the whole database** — every later write on every family failed until reopen.<br>Data was gone when the call returned. | `dropSync()` retires the name and returns.<br>A commit admitted before the retire lands normally, linearized before the drop; one that arrives after is refused whole with `ERR_COLUMN_FAMILY_DROPPED`.<br>Data is gone when the call returns **unless** a commit was in flight, in which case that commit deletes it on its way out. |

> The name is free and the old generation unwritable before `drop()` returns. The bytes are not guaranteed to be gone, and a process that exits inside that window leaves the family on disk under its name — the next open lists it as live, with its rows. A backup or checkpoint taken inside the window copies it.
> Callers that need a drop to survive a crash record their own durable tombstone before acknowledging it. Current Harper removes the catalog rows immediately after `dropSync()`, so it still needs a companion change that retains and recovers that tombstone until physical reclamation completes.
> A commit admitted before the drop still succeeds and publishes its transaction-log entries, even though the retiring generation is then physically deleted. Consumers must order the schema drop after those entries.

- [The caller-facing contract](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/README.md#L595-L640) — what a user reads, including the successful-commit/log-ordering consequence and the crash gap stated as a non-guarantee.
- [`retireColumnFamily`](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/src/binding/database/db_descriptor.cpp#L1769-L1828) — the logical half: identity-checked erase from `columns`, `retired = true`, entry in `retiring`, then the physical drop only when nothing holds a claim.

### Who owns the deletion
A drop and a commit can each discover the other first. Which one deletes the family, and how is it exactly one?

Both sides are sequentially consistent and two-phase: a retire stores `retired` then reads `admitted`, an admission increments `admitted` then reads `retired`. At least one side therefore observes the other, and both can conclude "reclaim now" — so a separate one-shot `claimReclaim()` decides which of them actually calls `DropColumnFamily`. Staging only *records* the families a transaction touched (weakly, 2 inline slots); it holds no claim, so an idle or abandoned transaction can never block a drop.

#### A drop landing while a commit is inside RocksDB
The drop returns as soon as the name is retired. The commit that was already admitted finishes normally and, as the last holder, performs the physical deletion.

```mermaid
sequenceDiagram
    participant Commit as Commit lane
    participant Gen as Generation
    participant Drop as drop() caller
    participant DB as RocksDB
    Commit->>Gen: admit (claim)
    Gen-->>Commit: admitted, not retired
    Drop->>Gen: retire
    Gen-->>Drop: a claim is held
    Drop-->>Drop: return, name already free
    Commit->>DB: write log batch, then Commit
    Commit->>Gen: release (last claim)
    Gen-->>Commit: you own the drop
    Commit->>DB: DropColumnFamily
```

#### The four points a physical drop can run from
1. **The drop itself** — nothing was admitted, so `dropSync()` deletes before returning. The common case, and identical to the old behavior.
2. **The last releasing commit** — a commit admitted before the retire finishes `txn->Commit()` and finds itself the last holder.
3. **A later drop or a same-name `open()`** — picks up a generation whose physical drop failed, or one a commit torn out of its pipeline abandoned.
4. **`finishClose()`** — drops whatever is still listed and destroys those RocksDB handles ahead of the database.

#### Example: a transaction that staged a write before the drop
A worker stages `users.putSync('k', v)` inside a transaction and then yields. Another thread drops `users`. The worker's transaction is still pending and holds no claim, so the drop is not delayed by it. The worker then commits.

**Outcome:** the commit is refused whole with `ERR_COLUMN_FAMILY_DROPPED` (`Column family "users" was dropped`), decided at admission before any transaction-log byte is written — so it surfaces as that error rather than as an abandoned transaction over log bytes already on disk. Any other family the same transaction wrote is not applied either.

- [The lifetime protocol](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/src/binding/core/column_family_lifetime.h#L9-L64) — Node-free and GoogleTest-covered: `admit`, `release`, `retire`, and the one-shot `claimReclaim`.
- [The commit's claim](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/src/binding/transaction/transaction.cpp#L280-L350) — taken once for every touched family, holding no reference to the descriptor, released after the commit returns.
- [Staging records, never claims](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/src/binding/transaction/transaction_handle.cpp#L716-L744) — a write to an already-retired family is refused here; everything else is one pointer compare.
- [The exactly-once gate](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/src/binding/database/db_descriptor.cpp#L1838-L1864) — the re-check loop that keeps a transient claim from leaving the generation to nobody.

### Reopening the same name
A drop frees the name immediately, but RocksDB cannot hold two families of one name. What does an `open()` inside the window do?

It waits for the old generation, then creates the fresh family. The wait polls in 20 ms slices with `databasesMutex` released, bounded by `ROCKSDB_JS_CF_RECLAIM_WAIT_MS` (default 30 s), after which it throws a retryable error. It is deadlock-free because the only thing it can be waiting on is a commit already inside RocksDB — on a commit lane, a libuv thread, or another thread's `commitSync` — never something parked on the opener's own event loop.

> This is a real cost, chosen knowingly: a commit slowed by a write stall turns a same-name `open()` into an event-loop stall of up to the bound on the opening worker, where an open never waited before. The alternative — create the fresh family under a different physical name — was not taken, because it moves naming policy into the binding.

- [The open's wait](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/src/binding/database/db_registry.cpp#L549-L581) — finds the retiring generation under `columnsMutex`, asks the reclaim gate to run the drop, and waits only when it reports that it could not.
- [The bound](https://github.com/HarperFast/rocksdb-js/blob/e454c78e2682778a139c04eeb4db987c3b22f555/src/binding/database/db_registry.cpp#L345-L359) — `ROCKSDB_JS_CF_RECLAIM_WAIT_MS`, read once per process, malformed or non-positive falling back to 30 s.

### What this costs a write, and the alternative that was measured and rejected
Does a per-family claim slow ordinary commits enough to justify a cheaper interlock?

No. The per-write addition is a relaxed `retired` load plus a pointer compare against the most-recently-touched slot; the per-commit addition is a `weak_ptr::lock` and two atomic read-modify-writes per distinct family. Measured in isolation against the real `ColumnFamilyLifetime`, the whole admission is **24 ns for a one-family commit and 145 ns for an eight-family one**, against a commit that takes 7.7 µs and 28 µs respectively — 0.3% to 0.5%. End-to-end benchmarks cannot resolve an effect that size; their run-to-run spread here was ±5% single-threaded and ±30% multi-worker, and a control build with the admission compiled out landed on both sides of this branch.

The obvious cheaper design — **one admission counter per database** instead of one per column family — was prototyped and measured rather than argued about. Uncontended it saves 13 ns of that 24 ns. Under concurrency it is worse, and worse in the shape Harper has: with each worker committing to its own family, per-family counters shard naturally and stay flat (28 ns at 1 thread, 36 ns at 12), while a single database-wide counter is one shared cache line and degrades to 1.2 µs per commit at 12 threads. It is also *more* machinery, not less: a database-wide counter records that commits are in flight but not which families they name, so the drop side must either block every write on the database while it drains — which is exactly the blocking this PR removed — or wait for the counter to reach zero, which on a busy database may never happen and can starve the deletion indefinitely. Per-family claims always drain, because that family's own commits always finish.

| Concurrent committers, each on its own family | Per-family claim (this PR) | One gate per database (rejected) |
| --- | --- | --- |
| 1 | 28 ns/commit | 11 ns/commit |
| 4 | 29 ns | 152 ns |
| 8 | 31 ns | 479 ns |
| 12 | 36 ns | 1191 ns |

### Open questions for the human
These are product and architecture calls, not correctness checks.

- Is "the name is free, the bytes may not be yet" the right contract for `drop()`, given that the crash window it creates is covered only by a caller-side tombstone? The alternative is #843's gate, where `dropSync()` returns only after the deletion — at the cost of blocking the calling thread, unboundedly under a write stall.
- Is a bounded wait the right behavior for a same-name `open()` inside the window, or should it fail fast and let the caller retry? The wait bound is process-global (`ROCKSDB_JS_CF_RECLAIM_WAIT_MS`) rather than a per-`open()` option.
- Should `columnFamily.pendingReclaims` distinguish "deferred behind a live commit" from "drop failed and keeps failing"? Collapsing the state machine made both read as 1, leaving the transient `log.warn` as the only signal for the second.
- Should #843 be closed, or rebased onto this? Its gate is subsumed by construction — same admission, taken once and earlier, with the blocking wait replaced by deferral.

## File coverage

- Native lifetime: [`ColumnFamilyDescriptor` gains `name`, `droppable` and `lifetime`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R956), and [`DBDescriptor` gains the `retiring` list](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R325) plus [`retireColumnFamily`/`reclaimColumnFamily`/`retryPendingReclaims`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R681); [`Database::Drop`/`DropSync` route through `dropColumnFamily`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R749), which retries earlier failures and retires the handle's generation, and their non-transactional `putSync`/`removeSync` skip a retired generation; [`DBHandle::clear()` is a no-op on one](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-927a214e5d5cca0234f56ddcab411eac454d4ee098051bde49cd6cd923828bb4R90) and [`columnFamily.pendingReclaims`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-927a214e5d5cca0234f56ddcab411eac454d4ee098051bde49cd6cd923828bb4R17) is served by `getStat`/`getStats`; [`DBRegistry::OpenDB` waits out a retiring generation of the name](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R556).
- Transactions: [`ColumnFamilySet` and `InlineVector`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-c49d956d0a464b270f42df08ca24d605ef2187bd043c9a6d69ba850e1d6401c7R101) hold the weak touched set inline (2 entries), recorded by [`noteTouchedColumnFamily`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-5d0063e928e32a18b23d6ce43f80fbd6fee68e69c9bccd4c792307bae0e9d4fdR717) and claimed by [`ColumnFamilyCommitClaim`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR290).
- Test seam: [`forceDropFailureForTesting(0|1|2)`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-9ea8fc3125cffd2aeaf8fdb203506abb4afe2ad0f10f16c1278576470fd6fda4R48), [exported from the binding](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-5bdfd0ac49ea126864c1cfbff17a606f1aa339fffd20472611f6fc68082ca199R338) and [typed in `load-binding.ts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-e5d300a7cae786782650a8e3f024c70312c0c5de94447dded589990def224845R860).
- Docs and types: [README "Dropping column families"](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R595) states the contract, transaction-log ordering consequence, and cross-restart gap; [AGENTS.md invariant 23](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R947) and [the `ROCKSDB_JS_CF_RECLAIM_WAIT_MS` entry](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R246) carry the rules; [`docs/stats.md`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-b01b1c97dca975c3d5b4b674ae04dbc2c1c56305aeaa5c4ca17621d12aaa49b2R37) and [`src/stats.ts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-0df0236447e5ca6985bd30924fc81823bd106d103d5f91d5bd2f6ce6692c6ab0R61) carry the new stat.
- Tests: [`column_family_lifetime_test.cc`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-e9340e280f3595fdf726ebfbf143aa5b562b00efb699e614a0440412f5dc4b15R1) is [registered in `binding.gyp`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-0d6aa7470314ca774fd7e07f67b35ecd3045fe5cb3a0e5c99f48bae897905d32R252); [`drop-deferred-reclamation.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-f8f044fc7f40924322ac1a12445b9575e4bca27f95a7bb1c2fa8a0fac296fdd6R64) drives the child scenarios in [`fork-drop-deferred-commit.mts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-5b58eaf0632c5f1fd74034f81449367e814b2d92e379ad60bdf1b92fb0c2bf57R1) and the worker in [`drop-deferred-commit-worker.mts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-a916eb8fcf3c29eddba1eee144265d7a550755cd468e02680043790e3bdc532eR1); [`drop.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-522bb7de846f02893e17e7a83b8ece2b98b60f4da4deae466546af5b8146b5ceR161) expects the attributable error; [`stats.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-585f8ebeb1fbdd0be0a5b161851ea5be0ef2641b279c5c2fe2b42b8c86a56c53R144) covers the stat; and the last-handle teardown regression is covered by [`txn-close-commit-uaf.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-81c4427f007de295d2a8027ec6836c9d719b413de0ca46b1968e139edbe92653R60), [`fork-legacy-commit-close.mts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-c58af78c63c2640d8b12f3791ac4cc574b4a030884287fc0fc5e2d756d34151dR36), and [`commit-last-handle-worker.mts`](https://github.com/HarperFast/rocksdb-js/pull/850/changes?w=1#diff-e4e1cf736bf3e6e81bff84c5c452a9368dc0b90165f32899305eb626ef449853R1).

## Verification

- **Fails on base, passes with the fix.** The issue's worker reproducer (worker commits a 4 KB value while the main thread drops), 15 rounds per mode against the merge-base build: optimistic 1/15 poisoned (the rest fail at validation), pessimistic 15/15. Branch: 0/15 both modes; every commit error is `Column family "table" was dropped` and the unrelated family keeps writing.
- **New tests.** `test/native/column_family_lifetime_test.cc` (9 GoogleTests: retire/admit/release orderings, single reclaim claim, a 200-round concurrent admitters-vs-retirer race). `test/drop-deferred-reclamation.test.ts` (child scenarios in `fork-drop-deferred-commit.mts`): a child-process matrix (`ROCKSDB_JS_COMMIT_THREAD` 0/1/2 × optimistic/pessimistic × `drop()`/`dropSync()`) with the worker's commit parked after admission by `ROCKSDB_JS_COMMIT_EXECUTE_DELAY_MS`, asserting the drop returns at once, the admitted commit lands, `getLastError()` stays null, the unrelated family keeps writing and the recreated family is empty; the committing worker terminated mid-claim; `open()` waiting for the claim; and in-process: a staged-but-idle transaction on the dropping thread, an override-only batch whose home family was dropped, an override family dropped after staging (refused whole), an iterator open across the drop, a coordinated retry whose second attempt touches a different family set, a logged transaction refused before any log byte, failed-drop retry through the next drop / another family's drop / `open()` / close, the mode-2 "already dropped" retry, `log.warn` and the stat. `drop.test.ts` and `stats.test.ts` expectations updated for the attributable error and the new stat key.
- **Final local gates at `e454c78e`.** `pnpm build:binding` and `pnpm check` pass; the focused close/deferred-reclamation run passes 48/48, commit teardown passes 2/2, and `pnpm test:native` passes 224/224. The Deno-focused teardown file passes 3 tests with the unsupported sole-worker cleanup case skipped; Node passes all 4. The full Vitest suite before the narrow teardown follow-up passed 68 files with 1 conditionally skipped (995 tests passed, 9 skipped). CI is fully green with 24 successful checks and 2 intentional skips; the isolated macOS Node 22 verification-table failure on the first attempt passed on rerun (998 tests passed, 8 skipped). The final exact-head delta review completed Claude and Gemini with `LGTM`; the immediately preceding whole-artifact review also completed Cursor Grok and found no production correctness defect. Domain adjudication failed at the service layer on that full pass, so the receipt remains explicitly non-converged with `Human-Review-Need: 4` (round 34, eight full-review passes).
- **Post-approval gates at `a52d792b`.** `pnpm build:binding` and `pnpm fmt` pass. Full Vitest suite: 68 files, 1006 tests passed, 9 skipped (the single failure on the first run was a missing `dist/index.mjs`, i.e. an unbuilt bundle, and passes after `pnpm build:bundle`). Focused re-runs after the commit-lane change: `txn-close-commit-uaf` + `commit-teardown` + `concurrent-teardown` + `transactions` 191 passed / 6 skipped; `drop-deferred-reclamation` + `drop` + `column-families` 77 passed. Three further pre-push review rounds ran on this head (codex + gemini + cursor-grok + Harper domain adjudication, `independent=true`); the two majors they raised in the new code are fixed above, and the round-3 pass produced no new actionable finding.
- **Benchmark** (single thread, 200 puts/txn, 500 txns × 10 rounds × 3 interleaved reps per build, merge-base worktree build vs this branch on the same toolchain):

| families | staging ns/put min | median | p95 | `commitSync` txn/s median | transaction begin ns |
|---:|---:|---:|---:|---:|---:|
| 1 | 1559 → 1579 | 1579 → 1625 (+2.9%) | 1622 → 1688 | 4552 → 4544 (−0.2%) | 2067 → 2197 |
| 3 | 1674 → 1734 | 1755 → 1761 (+0.4%) | 1792 → 1860 | 3495 → 3498 (+0.1%) | 2175 → 2407 |
| 9 | 1673 → 1636 | 1701 → 1748 (+2.7%) | 1733 → 1801 | 3117 → 3103 (−0.4%) | 2564 → 2694 |

Three earlier runs of the same harness put the base build's own medians 1559–1579 / 1708–1755 / 1681–1701 ns/put across runs, so the staging deltas are inside the base's run-to-run spread; the commit throughput is flat. That transaction-begin measurement predates the later reduction from eight to two inline touched/claimed slots, so its +130 ns and ~230-byte footprint are conservative upper bounds rather than measurements of the final layout.

- **Isolated admission cost and the rejected per-database gate** (tour section "What this costs a write"), measured directly against the real `ColumnFamilyLifetime` because the end-to-end harness above cannot resolve an effect this small: per-family admission is 24.4 ns for 1 family, 144.8 ns for 8, 266.2 ns for 16, per commit. A single per-database gate saves 12.8 ns at 1 family but costs 479 ns at 8 concurrent committers and 1191 ns at 12, where the per-family counters stay flat at 31 and 36 ns.

Complexity: complicated

https://claude.ai/code/session_016pbfw2ynnvvGY4JotmoKUz

<sub>Review-Coverage: authored=claude; ran=codex,gemini,cursor-grok; adjudicated=domain; declined=cursor-composer; rounds=37; full=8 @ a52d792bbba6</sub>

<sub>Human-Review-Need: 4 (decisions: drop-ack-before-durable, defer-to-last-releaser, open-waits-under-registry-lock, admitted-writes-discarded) @ a52d792bbba6</sub>


src/binding/database/db_descriptor.cpp
```diff
@@ -1715,41 +1730,250 @@ uint32_t DBDescriptor::transactionGetNextId() {
...
+			if (!column->column) {
+				column->lifetime.unclaimReclaim();
+				releaseOperation();
+				return rocksdb::Status::OK();
+			}
+			for (auto& entry : this->retiring) {
+				if (entry.descriptor == column) {
+					entry.state = RetiringColumnFamily::State::Reclaiming;
+				}
+			}
```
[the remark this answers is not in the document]
kriszyp: Not needed: an entry leaves `retiring` only on a successful drop, and a successful drop never unclaims `reclaimClaimed`, so a second caller cannot pass `claimReclaim()` for a generation that was already reclaimed. The one path that erases entries without a drop is `finishClose()`, which also resets the generation's RocksDB handle, and the `!column->column` guard just above returns before `DropColumnFamily` there. — Claude Fable 5.1

cb1kenobi (commented):
No confirmed blocking findings were identified in the supplied diff. Existing discussion was checked to avoid duplicate findings, and the change appears ready at this commit.

—
Reviewed 4647dca7

cb1kenobi (commented):
Logical retirement plus commit-duration claims keep a drop from racing an in-flight commit into RocksDB's fatal invalid-column-family error. Existing discussion on the older OpenDB busy-wait and a redundant DropColumnFamily check does not apply to this head. No confirmed blocking defect remains on the changed lines.

—
Reviewed 4647dca7

cb1kenobi (commented):
A drop racing a commit already inside RocksDB used to poison the whole database with a fatal background error. This change retires the column family logically right away and defers the physical drop until no commit still holds a claim on it, so the name frees immediately and the last releaser deletes the data. I traced claim/release pairing, the reclaim bookkeeping, and the rewritten open loop and found no blocking defect on the changed lines. Tests were not run in this review and CI was still in progress at this commit.

—
Reviewed 755883b4

cb1kenobi (commented):
No confirmed blocking defect was found at this commit. Existing discussion was checked and no duplicate findings were raised.

—
Reviewed a6e79fbc

cb1kenobi (commented):
A drop racing a commit already inside RocksDB used to poison the whole database with a fatal background error. This change retires the column family logically right away and defers the physical drop until no commit still holds a claim on it. The latest commits only fix rebase leftovers (invariant numbering and the admission log format). No blocking defect remains on the changed lines.

—
Reviewed a4127458

cb1kenobi (commented):
No confirmed blocking findings remain on the changed lines. Existing discussion was checked and no duplicate findings were raised.

—
Reviewed 22869c16

src/binding/transaction/transaction_handle.cpp
```diff
@@ -696,6 +698,34 @@ void TransactionHandle::ensureSnapshot() {
...
+rocksdb::Status TransactionHandle::noteTouchedColumnFamily(const std::shared_ptr<ColumnFamilyDescriptor>& column) {
+	if (!column) {
+		return rocksdb::Status::Aborted("Database not open");
+	}
+	if (!column->droppable) {
+		return rocksdb::Status::OK();
+	}
+	try {
+		if (column->lifetime.isRetired()) {
+			return rocksdb::Status::ColumnFamilyDropped("column family \"" + column->name + "\" was dropped");
```
cb1kenobi: This status gets turned into an error and for consistency sake, all rocksdb-js errors begin with a capital letter.
```suggestion
			return rocksdb::Status::ColumnFamilyDropped("Column family \"" + column->name + "\" was dropped");
```

src/binding/transaction/transaction.cpp
```diff
@@ -277,11 +277,88 @@ static unsigned parkTimeoutMs() {
...
+				this->claimed.add(std::move(column));
+				return;
+			}
+			if (column && reclaimNow) {
+				descriptor.reclaimColumnFamily(column);
+			}
+			try {
+				status = column
+					? rocksdb::Status::ColumnFamilyDropped("column family \"" + column->name + "\" was dropped")
+					: rocksdb::Status::ColumnFamilyDropped("column family was dropped and reclaimed");
```
cb1kenobi:
```suggestion
					? rocksdb::Status::ColumnFamilyDropped("Column family \"" + column->name + "\" was dropped")
					: rocksdb::Status::ColumnFamilyDropped("Column family was dropped and reclaimed");
```

AGENTS.md
```diff
@@ -920,6 +928,68 @@ sufficient (env teardown does not honor tsfn acquire counts); see
...
... admission about to be refused; it unclaims, re-reads, and either retries or leaves the drop to that admission's release, so the generation is never left to nobody). Staging only **records** the families a transaction touched (`TransactionHandle::touchedColumnFamilies`: **weak** references plus a raw pointer for comparison, inline for 8, most-recent slot compared first, the never-droppable default family skipped, cleared by `resetTransaction()` because the retry callback may touch a different set) and refuses every write to a retired family; it holds no claim. Weak, not strong: a transaction can outlive its database (an aborted handle JS still references after `close()`), and a strong reference there destroyed the RocksDB column-family handle after `finishClose()` had destroyed the database (SIGSEGV in `test/txn-close-commit-uaf.test.ts`). The commit locks them only for its claim, with the overflow storage reserved before the first claim is published so recording one cannot fail after its count was taken; a lock that fails is a dropped-and-reclaimed generation and refuses the commit like a retired one. That is deliberate: a staged-but-idle, abandoned, or drain-timeout-leaked (#784)
```
cb1kenobi: This looks like a really long line, please run `pnpm fmt`.

AGENTS.md
```diff
@@ -920,6 +928,68 @@ sufficient (env teardown does not honor tsfn acquire counts); see
...
+    non-transactional writes keep #725's silent discard. `DBRegistry::OpenDB` finds a `retiring` entry for the name under `columnsMutex` (so a drop cannot slip between the decision and the create), then simply asks `reclaimColumnFamily` to run the physical drop, under `databasesMutex` like the create itself — deadlock-free because a claim is only ever held by a commit inside RocksDB on a lane, a libuv thread, or another thread's `commitSync`, never parked on the opener's event loop. `attempted` is the whole decision: true and failed throws, true and OK creates the fresh family, false means a commit still holds the generation or another thread is already dropping it, so the open waits in 20 ms slices bounded by `ROCKSDB_JS_CF_RECLAIM_WAIT_MS` (default 30 s). **Do not reintroduce a status enum on the `retiring` entry.** An earlier revision tracked `Pending`/`Reclaiming`/`Failed` there, which duplicated `lifetime.admitted`/`reclaimClaimed` in a second place that had to be kept in step under a different lock; asking the one function that already reads those atomics is both shorter and impossible to desynchronize.
```
cb1kenobi: This looks like a really long line, please run `pnpm fmt`.

AGENTS.md
```diff
@@ -920,6 +928,68 @@ sufficient (env teardown does not honor tsfn acquire counts); see
...
+
+    Reclamation is retryable and never silently lost: membership in `retiring` IS "retry me", so a
+    failed `DropColumnFamily` only has to release its reclaim claim and leave the entry in place
+    (the whole path is `noexcept` because it runs from commit completions and destructors, and the
+    claim is released before any diagnostic allocation). It reports through the global `log.warn`
+    event and `columnFamily.pendingReclaims`,
+    and is retried on the next drop on the database, the next `open()` of that name (retried inline with `columnsMutex` released; a second failure throws), and `finishClose()`, which then destroys the RocksDB handle of every generation still in `retiring` ahead of the database: on the legacy libuv path a claim can outlive the closables sweep's drain timeout (#784), and `reclaimColumnFamily` — which also participates in `operationsInFlight` and stands down once the descriptor is closing — skips a generation whose handle is gone rather than drop into a destroyed database. Retry is idempotent
```
cb1kenobi: This looks like a really long line, please run `pnpm fmt`.

test/drop.test.ts
```diff
@@ -187,8 +186,11 @@ describe('Drop', () => {
 				// the rejection identifies the column family it could not reach,
 				// instead of the unattributable environment-wide message
 				const [staleResult] = results;
+				expect(staleResult.status === 'rejected' && staleResult.reason.code).toBe(
+					'ERR_COLUMN_FAMILY_DROPPED'
+				);
 				expect(staleResult.status === 'rejected' && staleResult.reason.message).toMatch(
-					/Could not access column family \d+/
+					/column family "doomed" was dropped/
```
cb1kenobi:
```suggestion
					/Column family "doomed" was dropped/
```

test/drop-deferred-reclamation.test.ts
```diff
@@ -0,0 +1,426 @@
...
+		));
+
+	describe('a physical drop that fails', () => {
+		it('is reported by the drop that ran it, keeps the name retired, warns, and is retried by the next drop', () =>
+			dbRunner(
+				{ dbOptions: [{ name: 'table' }, { name: 'table' }, { name: 'other' }] },
+				async ({ db: db1 }, { db: db2 }, { db: other }) => {
+					const warnings: string[] = [];
+					const onWarning = (message: string) => {
+						if (message.includes('column family "table"')) warnings.push(message);
```
cb1kenobi, line 458:
```suggestion
						if (message.includes('Column family "table"')) warnings.push(message);
```

cb1kenobi (changes requested):
Couple small nits, but otherwise this is good stuff!

src/binding/transaction/transaction_handle.cpp
kriszyp: Fixed in 5472d278. The staging-time public error now begins with “Column family”, and the README contract matches it.

src/binding/transaction/transaction.cpp
kriszyp: Fixed in 5472d278. Both commit-admission variants now use the capitalized public error convention.

AGENTS.md
kriszyp: Fixed in 5472d278 by manually wrapping this invariant paragraph. `pnpm fmt` and the full `pnpm check` both pass on the pushed head.

AGENTS.md
kriszyp: Fixed in 5472d278 by manually wrapping the caller-contract paragraph. `pnpm fmt` and the full `pnpm check` pass.

AGENTS.md
kriszyp: Fixed in 5472d278 by manually wrapping the reclamation paragraph. `pnpm fmt` and the full `pnpm check` pass.

test/drop.test.ts
kriszyp: Fixed in 5472d278 along with the native producer strings; the rebuilt focused suite passes 52/52 and the full suite passes 990 with 9 conditional skips.

test/drop-deferred-reclamation.test.ts
kriszyp: Declining this exact replacement: the emitted warning is `Physical drop of column family "table" ...`, so searching for capitalized `Column family` would not match it. The current lowercase substring deliberately matches the phrase inside the capitalized warning, and the full test suite passes.

kriszyp: Declining this exact replacement: the emitted warning is `Physical drop of column family "table" ...`, so searching for capitalized `Column family` would not match it. The current lowercase substring deliberately matches the phrase inside the capitalized warning, and the full test suite passes.

src/binding/transaction/transaction.cpp
```diff
@@ -382,19 +472,38 @@ static void executeLogWork(TransactionCommitState* state) {
  */
 static void executeCommitWork(TransactionCommitState* state) {
 	auto txnHandle = state->handle;
+	auto descriptor = state->descriptor.lock();
 	// The log stage already failed the commit on any invalid-handle condition,
 	// but the handle can also be torn out between the stages (e.g. a timed-out
 	// DBHandle::close() resetting the descriptor mid-pipeline). Never let a
 	// commit that was skipped here resolve as success.
```
cb1kenobi, line 566: **Commit stage no longer detects a torn-out database handle** — Severity: High

`state->descriptor` is a `weak_ptr` captured when the commit state was constructed, so `lock()` still succeeds after `DBHandle::close()` has reset *that handle's* `descriptor`, as long as any other handle keeps the `DBDescriptor` alive. The guard that used to catch exactly that case (`!txnHandle->dbHandle || !txnHandle->dbHandle->descriptor`) is gone, while the comment directly above still says it is there.

**Failure mode:** two handles are open on the same database. Handle A's `close()` drain times out, resetting `A->descriptor` and destroying the transaction, while A's commit sits between the log stage and the commit stage (`ROCKSDB_JS_COMMIT_THREAD=2` queues them separately, so that window is real rather than instantaneous). `state->descriptor.lock()` still succeeds because handle B keeps the descriptor alive, the guard passes, and `executeCommitWork` falls through to `txnHandle->txn->Commit()` on a `txn` that `TransactionHandle::close()` already `delete`d and set to `nullptr` — a null dereference on a commit lane. Before this change the same interleaving aborted cleanly with `Database closed during transaction commit operation`.

**Simplest fix:** keep both checks rather than folding them into one:

```cpp
if (!txnHandle || !txnHandle->dbHandle || !txnHandle->dbHandle->descriptor || !descriptor) {
```

The locked `descriptor` is still the right thing for `claim.release(descriptor.get())` and `resetTransaction(descriptor)` to use — it just should not also be what decides the handle is still usable.

—
Reviewed 5fbe1ff9

cb1kenobi (commented):
The deferred-drop design itself checks out — claims are paired on every commit exit and the reclaim gate settles on exactly one dropper. One regression: the new weak_ptr on the commit state also replaced the guard that detected a database handle torn out between the log and commit stages, so a commit can now reach txn->Commit() on a transaction that close() already destroyed. Restoring the handle and descriptor null checks alongside the weak lock fixes it. Chris's existing nits on error-message capitalization and AGENTS.md formatting are still open at this head.

—
Reviewed 5fbe1ff9

src/binding/transaction/transaction.cpp
kriszyp: Declining after tracing and reproducing the exact interleaving. A timed-out `TransactionHandle::close()` does **not** delete or null `txn`: `waitForAsyncWorkCompletion()` returns false and `close()` returns at transaction_handle.cpp:389-400, before deletion at :439-442. Conversely, a successful drain cannot happen between the two lanes because the one registered async-work count is released only by `executeCommitWork()` at transaction.cpp:577-579. The old handle-descriptor guard therefore prevented an admitted commit from completing after its DB handle closed; retaining the weak descriptor is the deliberate fix. I ran the two-handle `handle-closed` fixture in mode 2 with a 6000 ms execute delay (past the 5000 ms drain bound): it exits 0, the admitted commit lands, reclamation completes, and the database remains healthy. No null dereference occurs because the timeout path deliberately preserves `txn`.

src/binding/transaction/transaction_handle.cpp
```diff
@@ -696,6 +698,47 @@ void TransactionHandle::ensureSnapshot() {
...
+	const std::shared_ptr<ColumnFamilyDescriptor>& column,
+	bool& added
+) {
+	added = false;
+	if (!column) {
+		return rocksdb::Status::Aborted("Database not open");
+	}
+	if (!column->droppable) {
+		return rocksdb::Status::OK();
+	}
```
cb1kenobi, line 756: **Staging refusal leaves the transaction committable — Severity: High**

When this error is returned, the retired family is not added to `touchedColumnFamilies` and `writesAbandoned` remains false. A caller can catch the staging exception and commit earlier writes to other families, violating the documented whole-transaction refusal and leaving existing verification-table intents held. Before returning `ColumnFamilyDropped`, set `writesAbandoned = true` and release any locked VT intents, matching the commit-admission refusal path.

—
Reviewed 5472d278

cb1kenobi (commented):
A dropped-family staging error does not abandon the transaction, so callers that catch it can still commit earlier writes. Mark the transaction abandoned and release its verification-table intents before returning the error.

—
Reviewed 5472d278

cb1kenobi (commented):
This PR makes a column-family drop retire the name immediately and defer the physical delete until no in-flight commit still holds it, which fixes the database-wide fatal error a drop used to cause when it raced a commit. I traced the claim/release pairing on every commit exit, the exactly-once reclaim gate, the inline-vector bounds, and the rewritten open-and-wait loop, and found no new blocking defect on the changed lines. The previously raised comment about the staging-time refusal not abandoning the transaction is still open at this commit and was not repeated here. Tests were not run in this pass.

—
Reviewed 5472d278

cb1kenobi (commented):
The head commit makes a staging-time write to a dropped column family abandon the whole transaction and release its verification-table intents, instead of failing only that one call and letting a caller commit its earlier writes. I traced the abandon-before-throw ordering, the touched-set record/unwind around the RocksDB write, the inline-vector bounds, and the claim/release pairing on every commit exit, and found no blocking defect. Findings already raised at these code paths were checked and not repeated.

—
Reviewed 8fb6209e

src/binding/transaction/transaction_handle.cpp
kriszyp: Fixed in 8fb6209e. The retired-family staging path now sets `writesAbandoned` before any fallible work, releases any held verification-table intents, and then returns `ERR_COLUMN_FAMILY_DROPPED`. A caller that catches that error can still inspect transaction reads, but its later commit fails with `ERR_WRITES_ABANDONED` and cannot partially apply earlier writes. The focused test `abandons the transaction and releases its intents when a staging refusal is caught` proves both halves: `verifyVersion` succeeds after the refusal (the intent was released), and the earlier staged write remains absent after the rejected commit. I reran that test on 8fb6209e; it passes.

cb1kenobi (commented):
No confirmed blocking findings remain on the changed lines. Existing discussion already covers the relevant code paths, and the head commit resolves the staging-refusal issue.

—
Reviewed 8fb6209e

cb1kenobi (commented):
This PR makes a column-family drop retire the name immediately and defer the physical delete until no in-flight commit still holds a claim, which stops a drop racing a commit from poisoning the whole database. I traced the claim/release pairing on every commit exit, the exactly-once reclaim gate, the inline-vector bounds, the staging abandon-and-unwind path, and the rewritten open-and-wait loop, and found no blocking defect on the changed lines. Findings already raised in discussion were checked and not repeated. Note that the Deno (ubuntu) CI leg failed at this commit and was not investigated here.

—
Reviewed 06b7f58f

src/binding/database/db_registry.cpp
```diff
@@ -421,199 +440,242 @@ std::unique_ptr<DBHandleParams> DBRegistry::OpenDB(const std::string& path, cons
...
-			) {
-				pathClosingCondition = existingEntry.condition;
-				break;
+	// Armed on the first wait for a reclaiming generation, so time spent
+	// waiting for a closing database on this path does not count against it.
+	std::optional<std::chrono::steady_clock::time_point> reclaimDeadline;
+
+	// Re-entered after every reclaim wait: the slice releases `databasesMutex`,
+	// so the entry must be re-found.
+	for (;;) {
```
cb1kenobi, line 449: At some point, we should standardize our infinite loops. We use `while (true)` 10 times and `for (;;)` 5 times. There's no difference, but consistency is king.
```suggestion
	while (true) {
```

cb1kenobi (commented):
The new commits limit Node-specific worker fixtures to Node and clarify runtime coverage. No confirmed blocking defect was found on changed lines, and existing discussion was not repeated.

—
Reviewed e454c78e

src/binding/database/db_registry.cpp
kriszyp: Agreed that a repository-wide convention would be useful, but I am leaving this loop unchanged here: the native layer currently uses both forms (including three other `for (;;)` loops), and there is no documented or formatter-enforced preference. Converting this single loop would be unrelated style churn rather than completing a standardization. I have recorded it as a non-blocking follow-up suggestion.

cb1kenobi (commented):
This change makes a column-family drop free the name immediately and postpone the actual deletion until no in-flight commit is still using that family, which fixes a drop racing a commit poisoning the entire database. I followed the claim and release pairing through every way a commit can exit, the exactly-once deletion gate, the inline-vector bounds and unwind, and the rewritten open-and-wait loop, and found no blocking defect. Points already raised in discussion were checked and not repeated.

—
Reviewed e454c78e

src/binding/database/database.cpp
```diff
@@ -740,20 +740,35 @@ napi_value Database::Destroy(napi_env env, napi_callback_info info) {
...
+		dbHandle.descriptor->retryPendingReclaims();
+		bool retiredNow = false;
+		rocksdb::Status status = dbHandle.descriptor->retireColumnFamily(dbHandle.columnDescriptor, retiredNow);
+		if (retiredNow && dbHandle.enableVerificationTable) {
+			VerificationTable* vt = DBSettings::getInstance().getVerificationTableRaw();
+			if (vt) vt->settleAllSlots();
+		}
+		return status;
+	} catch (const std::exception& e) {
+		try {
```
cb1kenobi, line 759: Add a comment to explain why there's a try/catch in a catch.
```suggestion
		// wrap in a try/catch in case `std::string` throws with a `std::bad_alloc`
		try {
```

src/binding/transaction/transaction_handle.h
```diff
@@ -35,6 +36,124 @@ enum class TransactionState {
 	Aborted     // Transaction has been aborted/rolled back
 };
 
+template<typename T, size_t N>
```
cb1kenobi, line 52: Document why this struct is so awesome:
```suggestion
// A small vector that stores up to InlineCapacity elements inline to avoid
// heap allocation in the common case, and uses overflow storage beyond that.
template<typename T, size_t N>
```

cb1kenobi (approved):

src/binding/database/database.cpp
kriszyp: Good catch — added in `8e58dc0f`, with the specific reason rather than the general one: `Status::IOError(msg)` heap-copies the message, so building the status can itself throw `std::bad_alloc`, and `dropColumnFamily` is `noexcept` (invariant 23 — reclamation runs from commit completions and destructors). The inner catch lets that second failure fall through to the message-free `IOError()` instead of terminating.

— Claude Opus 5

src/binding/transaction/transaction_handle.h
kriszyp: Documented in `8e58dc0f`. Went with the why rather than the what, since the shape is recognizable as a small-vector but the reason for it here is not: both users — the set of column-family generations a transaction has staged a write to, and a commit's claim over that set — are rebuilt per transaction on paths that run for every staged write, and a transaction almost always names one or two families, so the inline slots keep the common case allocation-free. Also noted why `overflow` is public: `ColumnFamilyCommitClaim::admit` reserves it up front so recording a claim can never fail after its count was taken (invariant 23).

— Claude Opus 5

cb1kenobi (approved):

cb1kenobi (commented):
The latest commit fixes a use-after-free where releasing a commit's in-flight operation claim depended on C++ stack unwind order relative to a descriptor pin; I independently traced the destruction order and confirmed the fix closes the gap without introducing a new one. No blocking issues found on the changed lines.

—
Reviewed a52d792b

cb1kenobi (commented):
No confirmed blocking findings were identified on changed lines at a52d792b. Existing discussion was checked and no duplicate findings were raised. The change is ready.

—
Reviewed a52d792b

PR #853, opened 2026-09-14
fix(transaction): widen transaction ids to 64 bits so log writes survive past 2^31

### LINKED ISSUE #851: Transaction id produced as uint32 but read as int32 in TransactionLog.addEntry: every addEntry throws once a database's counter passes 2^31
**Version:** `@harperfast/rocksdb-js` 2.9.0 (also present on `main` as of 2026-09-14: `transaction_log.cpp:153`).

## Apparent bug

The transaction id is produced as an unsigned 32-bit value but consumed as a signed one on the transaction-log write path, so ids at or above 2,147,483,648 (2^31) are rejected:

- `DBDescriptor::nextTransactionId` is `std::atomic<uint32_t>` (`db_descriptor.h:336`), incremented per transaction in `transactionGetNextId()` (`db_descriptor.cpp:1640`); `TransactionHandle::id` is `uint32_t` (`transaction_handle.h:82`).
- `Transaction.id` is exposed to JS with `napi_create_uint32` (`transaction.cpp:1237`), so JS sees the full unsigned range.
- `TransactionLog.addEntry(entry, txnId)` reads the id back with `napi_get_value_int32` (`transaction_log.cpp:153`) and rejects negatives (`:154-156`). N-API truncates a number above `INT32_MAX` to its low 32 bits, so every id from 2^31 upward arrives negative and throws:

```
TypeError: Invalid argument, transaction id must be a non-negative integer
```

Once a database's counter passes 2^31 in a process, every subsequent `addEntry` on that database throws until the process restarts (the counter starts at 1 per process). The counter advances on every transaction, not only on ones that write log entries, so a long-running process with a high transaction rate reaches this in days.

## Reproduction (2.9.0, fresh temp database, macOS arm64)

```js
const { RocksDatabase } = await import('@harperfast/rocksdb-js');
const db = RocksDatabase.open(dir);
const log = db.useLog('probe');
const buf = Buffer.from('x');
for (const id of [1, 2147483647, 2147483648, 4294967295]) {
  try { log.addEntry(buf, id); console.log(id, '-> no throw'); }
  catch (e) { console.log(id, '->', e.constructor.name + ':', e.message); }
}
```

```
1          -> Error: Transaction id 1 not found
2147483647 -> Error: Transaction id 2147483647 not found
2147483648 -> TypeError: Invalid argument, transaction id must be a non-negative integer
4294967295 -> TypeError: Invalid argument, transaction id must be a non-negative integer
```

Ids that fit in int32 reach the store lookup (and fail there only because no such transaction exists in the test); ids at or above 2^31 never reach it.

## Observed effect in a long-running process

On a Harper process that had been up about 55 hours, this error began appearing at roughly 100 per second from the replication apply path (`RocksTransactionLogStore.put` → `log.addEntry(entryBinary, options.transaction.id)`, `RocksTransactionLogStore.ts:147`, inside `writeCommit`). RocksDB continued to commit records normally (new SST files, records present and readable), but no transaction-log entries were written from that point on, so the process could no longer replicate or replay those writes. Only a process restart clears it.

## Related observation

Because the counter is 32-bit, it also wraps at 2^32 and would start reusing ids that may still belong to live transactions (`transactionGet(id)` lookups). Not observed; noted because it follows from the same width.

kriszyp:
Transaction ids were allocated from a 32-bit counter but read back on the transaction-log write path as a *signed* 32-bit int, so once a database passed ~2.1 billion transactions every `addEntry` threw and the process stopped writing transaction-log entries entirely — permanently, until restart. This widens ids to 64 bits internally, surfaces them to JS as doubles, and routes every id entry point through one safe-integer parser.

Fixes #851.

## For the human reviewer

1. **Widen the id space vs. recycle ids of closed transactions.** Widening obliges every N-API boundary to be audited — round 1 of review proved that, see Changes — while recycling keeps the id narrow but reintroduces the reuse collision the [descriptor comment](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R333) was written to avoid. Chose widening. Reversing later means touching the same set of sites again.
2. **Surface ids as JS numbers capped at `Number.MAX_SAFE_INTEGER`, vs. BigInt or an opaque handle.** Numbers keep `txn.id` a `number` for every existing consumer, at the cost of a ceiling. The ceiling is [enforced at ingress](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-183eb3ef8b213558e2fe912d8864033f31c9807e51d29feedb5cbf56ac034257R423) but *not* at the producer, so above 2^53 the counter could mint an id the parser would reject. Unreachable in practice — ~285 years at 1M txn/s — and deliberately left unguarded rather than adding dead code. Changing `txn.id`'s JS type later is a breaking change for consumers.
3. **Validate at ingress rather than trusting the producer.** One parser at every entry point means a value arriving from JS is checked once, wherever it came from; trusting ids the binding itself emitted is what let the original defect reach production. Cost is one out-of-line call on the transactional path.
4. **Boundary assertions vs. a test seam that seeds the counter.** Ids above 2^31 are only reachable after 2^31 allocations, so the tests assert that such ids get *past* validation and fail the registry lookup. A seam to seed `nextTransactionId` is the only thing that would have caught the `database.cpp` gap below by execution rather than by review — see Verification for what that leaves uncovered. Not in this PR, which is wanted urgently; worth a follow-up.

## Changes

The id is a live-registry key only — [`TransactionLogHandle::addEntry`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-051460078325270d640664374a4186740917f42ec8ba4180f45440ebab0c256fR35) uses it to look the transaction up and nothing writes it into a log frame — so this widening carries **no on-disk format change and no replication or recovery surface**, and no behavior change below 2^31.

[`db_descriptor.h`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-a31cff767fb80ff129fd8f72784731a6c6db684ab520782ae1edddd7779d2587R341) and [`db_descriptor.cpp`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-bd416f52763b30cd2b46afe3d5ed49041190ba4785107273ac5cb132f6facc2eR1677) widen the `nextTransactionId` counter, the registry map key and the `transactionGet` signature to `uint64_t`. [`transaction_handle.h`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-c49d956d0a464b270f42df08ca24d605ef2187bd043c9a6d69ba850e1d6401c7R82) and [`transaction_handle.cpp`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-5d0063e928e32a18b23d6ce43f80fbd6fee68e69c9bccd4c792307bae0e9d4fdR146) widen the id each handle stores, with its diagnostic format specifiers updated to match. [`transaction_log_handle.h`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-2f12a9cacea2c52e6111440e5531aab290c18379a327a9ed5d771c6482f20daaR38) carries the widened type through the log handle.

[`readTransactionId`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-183eb3ef8b213558e2fe912d8864033f31c9807e51d29feedb5cbf56ac034257R415) in `helpers.cpp` is the single parser: it reads a double, then rejects non-integral, negative and above-`MAX_SAFE_INTEGER` values before narrowing. Its [declaration](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-d6c8884334f810d03dcf69d74e7c0a53dcaffb3e609b79a0255798f964d56492R229) in `helpers.h` carries the constraint that keeps the next call site correct — do not reach for the fixed-width N-API helpers on this path.

**Where to look hardest:** that every id entry point actually uses it. [`database.cpp`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-06197a1d979aaed331a1e6bab83dfcdc4a16f049f865c5a9814006c4a2d25c83R1045) converts five (`Get`, `GetSync`, `GetCount`, `PutSync`, `RemoveSync`), [`db_iterator.cpp`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-a2115b03db8a169d61f4f35b833a2576a009531050160d3e6fbfb7a997faca08R159) the sixth, and [`transaction_log.cpp`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-6a5fb51459adc2a017bf9767a4dc899c0681885c1fb96431af114466605d3448R155) the seventh — the one the original defect lived on. The first six were **not** in the first commit: they read the id with `napi_get_value_uint32`, which would have moved this exact break from 2^31 to 2^32 rather than removing it, and a wrapped id colliding with a live transaction stages writes into *that* transaction instead of throwing. Cross-model review caught it. The audit behind the fix is that all seven `transactionGet` callers are now fed by the parser, and no `napi_create_uint32`/`int32` emits or reads an id anywhere in `src/binding`.

On the emitting side, [`Transaction::Id`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR1237) and [`Transaction::UseLog`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR1448) emit doubles instead of `napi_create_uint32`, and [`db_registry.cpp`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-1ac655e46fe2a38bac691bf6882595e27c00f8f61015ff57b7d305c13de15b98R756) fixes the same truncation in the diagnostics dump, which would otherwise have reported a wrapped id to anyone debugging this. Separately, [`Transaction::Commit`'s completion callback](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-f934da63dcfdc6ecd2904bdded2fdd7fe3f5b2c7f868edb65449be7985888f5eR860) passed an `unsigned long long` to `%d`; `rocksdb_js::debugLog` carries no printf format attribute so no build diagnoses it, and above 2^32 a DEBUG build logged the low 32 bits and pointed the reader at the wrong transaction.

[`README.md`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R1163) documented `txn.id` as "a 32-bit unsigned integer", which this change makes false — a consumer that took it at face value and normalized the id (`id >>> 0`, a `Uint32Array` slot, a 32-bit wire field) would collide two live transactions past 2^32, exactly the window the widening exists to survive.

## Verification

End-to-end route: **unit tests at the N-API argument boundary**, plus a live read-only measurement on the affected production cluster that confirmed the diagnosis. Ids above 2^31 are only reachable after 2^31 allocations, so no test creates one; the tests instead assert that such ids get past validation and reach the registry lookup.

[`test/transaction-log.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-22a3c4e9b693561b0d3d6534ae81880ea0dec05f14fc12b7bbc837a5748e59e5R1198) proves `addEntry` accepts 2^31, 2^32+5 and `MAX_SAFE_INTEGER` and still rejects `-1`, `1.5`, `NaN` and `MAX_SAFE_INTEGER + 2`. [`test/transactions.test.ts`](https://github.com/HarperFast/rocksdb-js/pull/853/changes?w=1#diff-5bd3a0a6c44eed3bd5aca184b809304846d60a20d1e69ec89025d35d51427465R800) proves the same for all six `Database` entry points (`get`, `getSync`, `getKeysCount`, `putSync`, `removeSync`, `getRange`) across all four transaction modes — without the last three, a partial revert of the `database.cpp` change left the suite green.

Both go red on `main` (stashed `src/`, kept the tests, forced `node-gyp rebuild`), the first with the production error verbatim:

```
× should accept transaction ids at or above 2^31
  Expected: "Transaction id 2147483648 not found"
  Received: "Invalid argument, transaction id must be a non-negative integer"
× should error if transaction id is invalid
  expected to throw /must be a non-negative integer/ but got 'Transaction id 1 not found'
```

That second red line is the aliasing hazard rather than the rejection: on `main`, `MAX_SAFE_INTEGER + 2` wraps to `1` and resolves against whatever live transaction holds id 1. The `Database` paths measured the same before the fix — `2**31`, `2**32+5` and `MAX_SAFE_INTEGER` returned `txnId: 0`, `5` and `4294967295`; each now reports its full id.

`pnpm test` — 955 tests across 67 files, 938 passed / 17 skipped. `npm run test:native` — 194 gtest cases, 191 passed / 3 skipped (madvise cases this platform skips). `tsc --noEmit`, `oxlint` and `oxfmt --check` clean.

**Not covered:** nothing drives a transaction actually *allocated* above 2^31, so the outbound `napi_create_double` conversion and live-registry behavior at that magnitude rest on code trace. That is decision 4 above.

The pre-push harness's `one-engine-only` self-check fires on this diff ("no touched test mentions LMDB"); it is a false positive — rocksdb-js has no LMDB arm.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=gemini,cursor-grok,codex; adjudicated=domain; declined=cursor-composer; rounds=3; full=3 @ 60d33ecaa8f6</sub>

<sub>Human-Review-Need: 4 @ 60d33ecaa8f6</sub>


cb1kenobi (commented):
The widening from 32-bit to 64-bit transaction ids is correct and complete across every id entry point in the diff, and the shared readTransactionId validator handles NaN, infinity, negatives, non-integers and the 2^53 ceiling correctly. Error message shapes and the UseLog/Constructor round-trip stay consistent, so existing assertions still hold. One thing to confirm before merge: the Native C++ tests on windows-latest check is failing on this commit while every other matrix leg passes, which contradicts the 'full suite green' claim in the description. Its very short runtime suggests an early build or setup failure rather than a regression from this diff.

—
Reviewed 67d22fb1

cb1kenobi (commented):
No blocking defects were confirmed in the changed lines. The 64-bit ID flow and safe-integer validation are consistent across the reviewed entry points, and all current checks pass.

—
Reviewed 60d33eca

cb1kenobi (approved):

